PR #40, opened 2018-07-13
Crowd map remove all from parameters and sensors

3v0k4:
Closes https://llp.kanbanery.com/projects/8040/board/tasks/2322468

Done
- Remove all from parameters' and sensors' dropdowns in crowd map
- Select sensor with the most sessions when changing parameter in crow map

Improved
- remove duplication by pushing code to `sensors.js`
- move some private functions (eg `max`) out of the exported service 
- added comments for the future, when the code will be cleaner we will remove them

Notice that the sensor selected below is the second (most sessions) not the first.
jul-13-2018 13-19-22

### DIFF
--- a/app/assets/javascripts/code/services/sensors.js
+++ b/app/assets/javascripts/code/services/sensors.js
@@ -143,6 +143,24 @@ angular.module("aircasting").factory('sensors', ['params', '$http', 'spinner', f
     buildSensorId: function(sensor) {
       return sensor.measurement_type + "-" + sensor.sensor_name + " (" + sensor.unit_symbol + ")";
     },
+    onSensorsSelectedIdChange: function(newValue, oldValue, callback = null) {
+      console.log("onSensorsSelectedIdChange - ", newValue, " - ", oldValue);
+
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/code/services/sensors.js
```diff
@@ -1,4 +1,8 @@
 angular.module("aircasting").factory('sensors', ['params', '$http', 'spinner', function(params, $http, spinner) {
+  function sensorChangedToAll(newValue) {
```
3v0k4: private fnc

app/assets/javascripts/code/services/sensors.js
```diff
@@ -121,24 +122,47 @@ angular.module("aircasting").factory('sensors', ['params', '$http', 'spinner', f
         return this.sensors;
       }
     },
-    onSelectedParameterChange: function(selectedParameter) {
+    onSelectedParameterChange: function(selectedParameter, oldValue, onChangeSelectFirstSensor = false) {
```
3v0k4: another fnc

PR #46, opened 2018-09-03
Mobile: return to prev map position after deselecting session

3v0k4:


app/assets/javascripts/code/services/_mobile_sessions.js
```diff
@@ -71,16 +74,18 @@ export const mobileSessions = (
...
 
     deselectSession: function(id) {
-      var session = this.find(id);
-      if(!session) return;
+      const session = this.find(id);
+      if (!session) return;
       session.loaded = false;
       session.$selected = false;
       session.alreadySelected = false;
-      drawSession.undoDraw(session, boundsCalculator(this.sessions));
```
3v0k4: This is ok only if one session can be checked at once.

PR #50, opened 2018-11-13
Upgrade(ruby): from 2.0 to 2.2

AkkarinForest:
We are going through the process of upgrading the stack in this project. We decided to go with small jumps and start with Ruby.

According to [this blog post](https://weblog.rubyonrails.org/2015/6/16/Rails-3-2-22-4-1-11-and-4-2-2-have-been-released-and-more/) ruby 2.2 is the last version compatible with rails 3.2

In the process of upgrading:
- I deleted the guard gems as they were causing dependency issues. We can install them again when we are done with upgrading.
- I had to add `test-unit` gem due to the error, while running tests. We can try to delete it if it turns out to be obsolete after another update. Error:
```
/Users/alexander/.rvm/gems/ruby-2.2.10/gems/activesupport-3.2.22.5/lib/active_support/dependencies.rb:251:in `require': Please add test-unit gem to your Gemfile: `gem 'test-unit', '~> 3.0'` (cannot load such file -- test/unit/testcase) (LoadError)
```
- I had to restrict RSpec gem version to ~> 2.99. Otherwise, version 3.8 was installed and ~ 60 tests were failing. We can consider upgrading to the newest version and rewriting the tests in the next step.
- I had to restrict mysql2 version, otherwise a higher version is installed and cause an error while running tests. We should check if it's possible to remove this restrictions after another update. Error:
```
/Users/alexander/.rvm/gems/ruby-2.2.10/gems/bundler-1.17.1/lib/bundler/rubygems_integration.rb:408:in `block (2 levels) in replace_gem': Please install the mysql2 adapter: `gem install activerecord-mysql2-adapter` (can't activate mysql2 (~> 0.3.10), already activated mysql2-0.5.2. Make sure all dependencies are added to Gemfile.) (LoadError)
```


3v0k4:
I'd speculate some gems are not even used (`guard` for example) and we should remove mercilessly those things.

Great job and thanks for providing the context in the description! Let's wrap up and push later today

config/deploy/production.rb
```diff
@@ -44,8 +44,8 @@
     rbenv_root: '/usr/local/rbenv'
 }
 set :rbenv_roles, :all
-set :rbenv_ruby, '2.0.0-p647'
-set :rbenv_ruby_dir, '/usr/local/rbenv/versions/2.0.0-p647'
+set :rbenv_ruby, '2.2.10'
```
3v0k4: There's duplicated `:rbenv_ruby` in `config/deploy/*` files. Could you please remove one of the two?

PR #52, opened 2018-11-14
feat(outliers): service to filter measurement outliers

3v0k4:
This PR introduces a service `Outliers::FilterMeasurements` that takes some measurements and filters out outliers.

This is the first step to be able to cache the correct bounding box of a session. Correct means that the bounding box doesn't take into account outliers.

### DIFF
--- a/app/services/outliers/calculate_centroid.rb
+++ b/app/services/outliers/calculate_centroid.rb
@@ -1,0 +1,22 @@
+class Outliers::CalculateCentroid
+  def call(measurements)
+    lng_lat = sum_coordinates(measurements)
+    average(lng_lat, BigDecimal.new(measurements.size))
+  end
+
+  private
+
+  def sum_coordinates(measurements)
+    measurements.reduce([zero, zero]) do |(lng, lat), x|
+      [lng + x.longitude, lat + x.latitude]
+    end
+  end
+
+  def average(lng_lat, amount)
+    lng_lat.map { |x| x / amount }
+  end
+
+  def zero
+    "0".to_d
+  end
+end
--- a/app/services/outliers/calculate_centroid.rb
+++ b/app/services/outliers/calculate_centroid.rb
@@ -12,8 +12,8 @@ class Outliers::CalculateCentroid
     end
   end
 
-  def average(lng_lat, amount)
-    lng_lat.map { |x| x / amount }
+  def average(lng_lat, number)
+    lng_lat.map { |x| x / number }
   end
 
   def zero
--- a/app/services/outliers/calculate_distance.rb
+++ b/app/services/outliers/calculate_distance.rb
@@ -1,0 +1,9 @@
+class Outliers::CalculateDistance
+  def call(lng_lat1, lng_lat2)
+    xx1, yy1 = lng_lat1
+    xx2, yy2 = lng_lat2
+
+    BigDecimal.new((xx1 - xx2) ** 2 + (yy1 - yy2) ** 2).sqrt(12)
+  end
+end
+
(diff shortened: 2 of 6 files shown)


app/services/outliers/filter_measurements.rb
```diff
@@ -0,0 +1,24 @@
+class Outliers::FilterMeasurements
+  MAX_DISTANCE = 100
```
3v0k4: I need to tweak this. Will in a future PR

app/services/outliers/calculate_centroid.rb
```diff
@@ -0,0 +1,22 @@
...
+  private
+
+  def sum_coordinates(measurements)
+    measurements.reduce([zero, zero]) do |(lng, lat), x|
+      [lng + x.longitude, lat + x.latitude]
+    end
+  end
+
+  def average(lng_lat, amount)
+    lng_lat.map { |x| x / amount }
```
AkkarinForest: Is it 100% sure that amount will never be 0? If it's not an issue in this case then ignore this comment :)

And it's just a detail, but it might be better to call it number. Amount is for uncountable stuff. That's how I remember :D https://pl.pinterest.com/pin/650910952367202906/

app/services/outliers/calculate_distance.rb
```diff
@@ -0,0 +1,9 @@
+class Outliers::CalculateDistance
+  def call(lng_lat1, lng_lat2)
```
AkkarinForest: What do you think about using `coordinates` as a name instead of `lng_lat`? And then we could use `longitude, latitude = coordinates` to unpack them.

spec/services/outliers/calculate_distance_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  it "it returns the distance 1" do
+    lng_lat1 = [one, one]
+    lng_lat2 = [one, two]
+
+    actual = @subject.call(lng_lat1, lng_lat2)
+
+    expect(actual).to eq(one)
+  end
+
+  it "it returns the distance 2" do
```
AkkarinForest: This test case returns the distance 1.41... not 2. Unless you just meant to enumerate the test cases this way?

spec/services/outliers/filter_measurements_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      longitude: one_thousand,
+      latitude: one_thousand,
+    )
+
+    measurements = [m1, m2, m3]
+
+
+    actual = subject.call(measurements)
+
+
```
AkkarinForest: no need for new line

app/services/outliers/calculate_distance.rb
3v0k4: I prefer `lng_lat` cause it's clear what's the order in the array. Makes it harder to do 
```
lat, lng = lng_lat
```
which is wrong
vs
```
lng, lat = lng_lat
```

spec/services/outliers/calculate_distance_spec.rb
3v0k4: Second one. I was enumerating ;)

spec/services/outliers/filter_measurements_spec.rb
3v0k4: I do that on purpose. I style tests in the following way
```
# given

# when

# then
```

For example
```
  it "when the points are the same it returns 0" do
     lng_lat = [one, one] # given

     actual = @subject.call(lng_lat, lng_lat) # when

     expect(actual).to eq(zero) # then
  end
```

Unfortunately, sometimes tests get long. In that case, I use two lines instead of one to divide. If the test is long it's a smell though.

app/services/outliers/calculate_centroid.rb
3v0k4: Good catch! I've added a guard in the calling method (https://github.com/HabitatMap/AirCasting/pull/52/commits/a5f136f028d6f93f848fa11c3e2e34c2a1b730a3)

PR #53, opened 2018-11-15
Remove outliers when calculating bounding box for stream

3v0k4:
This PR wraps up the feature.

In particular, when calculating the bounding box for all the measurements on a stream, the code tries to eliminate outliers.

The entry point is `Outliers::CalculateBoundingBox`. In order it calls:
- `Outliers::CalculateCentroid` which calculates the geographic center of all the measurements
- `Outliers::CalculateBoundingBoxes` to calculate both the bounding box with all measurements and the one without outliers (outliers are identified using `Outliers::CalculateDistance` and checking if a measurement is closer than `MAX_DISTANCE` to the centroid)
- `Outliers::SelectBoundingBox` which decides which bounding box to use (i.e. the one with all measurements or the one with filtered outliers); the decision is taken by considering how many outliers have been identified; if the number of outliers is above 5% then we decide to take the bounding box out of all measurements. In fact, outliers to be such must be below 5%.


app/repositories/streams_repository.rb
```diff
@@ -3,11 +3,13 @@ def find(id)
     Stream.find(id)
   end
 
-  def calc_bounding_box!(stream)
-    stream.min_latitude = stream.measurements.minimum(:latitude)
-    stream.max_latitude = stream.measurements.maximum(:latitude)
-    stream.min_longitude = stream.measurements.minimum(:longitude)
-    stream.max_longitude = stream.measurements.maximum(:longitude)
+  def calc_bounding_box!(stream, calculate_bounding_box = Outliers::CalculateBoundingBox.new)
```
3v0k4: Most of this stuff should be extracted to a service. But have no chance to make it happen now

spec/services/outliers/calculate_bounding_box_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  it "delegates to other services" do
+    centroid = [one, one]
+    measurements = [Measurement.new]
+    bounding_boxes = { with_outliers: { a: 1 }, without_outliers: { b: 2 } }
+    bounding_box = bounding_boxes.fetch(:without_outliers)
+    calculate_centroid = double
+    calculate_centroid.should_receive(:call).with(measurements) { centroid }
+    calculate_bounding_boxes = double
+    calculate_bounding_boxes.should_receive(:call).with(centroid, measurements) { bounding_boxes }
```
AkkarinForest: In the spirit of upgrading would you consider using:
expect(double).to receive(:foo).with(something) { return }
source: https://relishapp.com/rspec/rspec-mocks/docs/old-syntax/should-receive

app/services/outliers/calculate_bounding_boxes.rb
```diff
@@ -0,0 +1,64 @@
+class Outliers::CalculateBoundingBoxes
+  INITIAL_BOUNDING_BOX = {
+    min_latitude: "+90".to_d,
+    max_latitude: "-90".to_d,
+    min_longitude: "+180".to_d,
+    max_longitude: "-180".to_d,
+  }
+  INITIAL_BOUNDING_BOXES = {
```
AkkarinForest: The name INITIAL_BOUNDING_BOXES doesn't seem the best place to contain number_of_outliers. But I don't have any idea how to call differently, so you can ignore me :)

3v0k4: I feel you. Same here :)

spec/services/outliers/calculate_bounding_box_spec.rb
3v0k4: Totally. Will do before deploy. Good catch

PR #54, opened 2018-11-15
Add utc time to measurements coming from fixed sessions.

AkkarinForest:
It's not clear what is the bast place to put the logic that will add the utc time to the measurements. I wen't with MeasurementsCreator class even though it doesn't really creates the measurements.

I will leave my notes here, just in case they help someone :). I looked through the structure of the code and there are 3 controllers that created measurements, calling classes in this order:

* controller: Api::Realtime::MeasurementController.create
  * lib: RealtimeMeasurementBuilder.new.build!
  * lib: RealtimeMeasurementBuilder.new.build_measuremnt!
  * model: Stream.build_or_update!
  * service: MeasurementsCreator.call
     - at this moment it becomes async so the time has to be added before
  * service: MeasurementsCreator.new.call
  * model: Stream.build_measurements!
  * model: Measurement.new

* Api::Realtime::SessionController.create
   * lib: SessionBuilder.build!
   * lib: SessionBuilder.build_session!
   * model: Stream.build!
   * service: MeasurementsCreator.call
      - at this moment it becomes async so the time has to be added before
   * service: MeasurementsCreator.new.call
   * model: Stream.build_measurements!
   * model: Measurement.new

* Api::MeasurementSessionController.create 
  I think this is only used by mobile sessions
  * lib: SessionBuilder.build!
  * lib: SessionBuilder.build_session!
  * model: Stream.build!
  * service: MeasurementsCreator.call
     - at this moment it becomes async so the time has to be added before
  * service: MeasurementsCreator.new.call
  * model: Stream.build_measurements!
  * model: Measurement.new

### DIFF
--- a/app/models/stream.rb
+++ b/app/models/stream.rb
@@ -180,8 +180,8 @@ class Stream < ActiveRecord::Base
     max_longitude.present? &&
     min_longitude.present?
   end
-  
+
   def fixed?
-    self.session.type == "FixedSession"
+    session.fixed?
   end
 end
--- a/app/models/stream.rb
+++ b/app/models/stream.rb
@@ -180,4 +180,8 @@ class Stream < ActiveRecord::Base
     max_longitude.present? &&
     min_longitude.present?
   end
+  
+  def fixed?
+    self.session.type == "FixedSession"
+  end
 end
--- a/app/services/measurements_creator.rb
+++ b/app/services/measurements_creator.rb
@@ -2,12 +2,7 @@ class MeasurementsCreator
   SLICE_SIZE = 500
 
   def self.call(stream, measurements_attributes)
-    if stream.session.type == "FixedSession"
-      measurements_attributes = measurements_attributes.map do |measurement_attributes|
-        measurement_attributes[:utc_time] = Time.now.utc
-        measurement_attributes
-      end
-    end
+    measurements_attributes = add_arrival_time(measurements_attributes) if stream.fixed?
 
     if measurements_attributes.count == 1
       new.call(stream, measurements_attributes)
@@ -26,7 +21,7 @@ class MeasurementsCreator
     stream.build_measurements!(measurements_attributes)
     stream.after_measurements_created
 
-    return if stream.session.type == 'FixedSession'
+    return if stream.fixed?
     streams_repository.calc_bounding_box!(stream)
     streams_repository.calc_average_value!(stream)
   end
@@ -34,4 +29,11 @@ class MeasurementsCreator
   private
 
   attr_reader :streams_repository
+
+  def self.add_arrival_time(measurements_attributes)
+    measurements_attributes.map do |measurement_attributes|
+      measurement_attributes[:arrival_utc_time] = Time.current.utc
+      measurement_attributes
+    end
+  end
 end
--- a/app/services/measurements_creator.rb
+++ b/app/services/measurements_creator.rb
@@ -2,6 +2,13 @@ class MeasurementsCreator
   SLICE_SIZE = 500
 
   def self.call(stream, measurements_attributes)
+    if stream.session.type == "FixedSession"
+      measurements_attributes = measurements_attributes.map do |measurement_attributes|
+        measurement_attributes[:utc_time] = Time.now.utc
+        measurement_attributes
+      end
+    end
+
     if measurements_attributes.count == 1
       new.call(stream, measurements_attributes)
     else
--- a/app/services/measurements_creator.rb
+++ b/app/services/measurements_creator.rb
@@ -32,7 +32,7 @@ class MeasurementsCreator
 
   def self.add_arrival_time(measurements_attributes)
     measurements_attributes.map do |measurement_attributes|
-      measurement_attributes[:arrival_utc_time] = Time.current.utc
+      measurement_attributes[:arrival_utc_time] = Time.current
       measurement_attributes
     end
   end
--- a/db/migrate/20181114165757_add_utc_time_to_measurement.rb
+++ b/db/migrate/20181114165757_add_utc_time_to_measurement.rb
@@ -1,0 +1,5 @@
+class AddUtcTimeToMeasurement < ActiveRecord::Migration
+  def change
+    add_column :measurements, :utc_time, :datetime
+  end
+end
--- a/db/migrate/20181114165757_add_utc_time_to_measurement.rb
+++ b/db/migrate/20181114165757_add_utc_time_to_measurement.rb
@@ -1,5 +1,5 @@
 class AddUtcTimeToMeasurement < ActiveRecord::Migration
   def change
-    add_column :measurements, :utc_time, :datetime
+    add_column :measurements, :arrival_utc_time, :datetime
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -32,7 +32,7 @@ ActiveRecord::Schema.define(:version => 20181114165757) do
     t.integer  "stream_id"
     t.integer  "milliseconds",                                   :default => 0
     t.float    "measured_value"
-    t.datetime "utc_time"
+    t.datetime "arrival_utc_time"
   end
 
   add_index "measurements", ["latitude"], :name => "index_measurements_on_latitude"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20181113100336) do
+ActiveRecord::Schema.define(:version => 20181114165757) do
 
   create_table "deleted_sessions", :force => true do |t|
     t.datetime "created_at", :null => false
@@ -32,6 +32,7 @@ ActiveRecord::Schema.define(:version => 20181113100336) do
     t.integer  "stream_id"
     t.integer  "milliseconds",                                   :default => 0
     t.float    "measured_value"
+    t.datetime "utc_time"
   end
 
   add_index "measurements", ["latitude"], :name => "index_measurements_on_latitude"
--- a/spec/services/measurements_creator_spec.rb
+++ b/spec/services/measurements_creator_spec.rb
@@ -1,23 +1,21 @@
 require "spec_helper"
 
 describe MeasurementsCreator do
-  let(:measurements_attributes) { [{:longitude=>25.4356212, :latitude=>56.4523456, :time=>"2016-05-11T17:09:02", :milliseconds=>925, :measured_value=>59.15683475380729, :value=>59.15683475380729}] }
-
-  it "creates a mesurement" do
+  it "creates a measurement" do
     stream = create_stream!
     MeasurementsCreator.call(stream, measurements_attributes)
 
     expect(Measurement.count).to eq(1)
   end
 
-  it "creates a mesurement with utc time for fixed sessions" do
+  it "creates a measurement with utc time for fixed sessions" do
     stream = create_stream!("FixedSession")
     MeasurementsCreator.call(stream, measurements_attributes)
 
     expect(Measurement.first.utc_time).to be_within(1.second).of Time.now
   end
 
-  it "creates a mesurement without utc time for mobile sessions" do
+  it "creates a measurement without utc time for mobile sessions" do
     stream = create_stream!("MobileSession")
     MeasurementsCreator.call(stream, measurements_attributes)
 
@@ -25,6 +23,16 @@ describe MeasurementsCreator do
   end
 
   private
+
+  def measurements_attributes
+    [{ longitude: 25.4356212,
+      latitude: 56.4523456,
+      time: "2016-05-11T17:09:02",
+      milliseconds: 925,
+      measured_value: 59.15683475380729,
+      value: 59.15683475380729,
+    }]
+  end
 
   def create_stream!(session_type  = "MobileSession")
     Stream.create!(
--- a/spec/services/measurements_creator_spec.rb
+++ b/spec/services/measurements_creator_spec.rb
@@ -8,7 +8,7 @@ describe MeasurementsCreator do
         session = create_session!
         stream = create_stream!(session: session)
 
-        MeasurementsCreator.call(stream, [measurement_attributes])
+        MeasurementsCreator.call(stream, single_measurement_attributes)
 
         expect(Measurement.count).to eq(1)
       end
@@ -17,7 +17,7 @@ describe MeasurementsCreator do
         session = create_session!(type: "FixedSession")
         stream = create_stream!(session: session)
 
-        MeasurementsCreator.call(stream, [measurement_attributes])
+        MeasurementsCreator.call(stream, single_measurement_attributes)
 
         expect(Measurement.first.arrival_utc_time).to be_within(1.second).of Time.current
       end
@@ -26,7 +26,7 @@ describe MeasurementsCreator do
         session = create_session!(type: "MoblieSession")
         stream = create_stream!(session: session)
 
-        MeasurementsCreator.call(stream, [measurement_attributes])
+        MeasurementsCreator.call(stream, single_measurement_attributes)
 
         expect(Measurement.first.arrival_utc_time).to be_nil
       end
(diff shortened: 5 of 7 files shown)


spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,60 @@
+require "spec_helper"
+
+describe MeasurementsCreator do
+  let(:measurements_attributes) { [{:longitude=>25.4356212, :latitude=>56.4523456, :time=>"2016-05-11T17:09:02", :milliseconds=>925, :measured_value=>59.15683475380729, :value=>59.15683475380729}] }
```
3v0k4: Please consider one key/value per line so that code does get to horizontal.
Also, consider using the newer syntax (ie `{ key: value }`)

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  it "creates a mesurement" do
```
3v0k4: typo :)

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    stream = create_stream!
```
3v0k4: Would the test work if the stream was not persisted? (ie `Stream.new` instead of `Stream.create`). That way we avoid one trip to db and make tests faster

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    MeasurementsCreator.call(stream, measurements_attributes)
```
3v0k4: Please leave a line in between the given and the when
```
stream = create_stream!

MeasurementsCreator.call(stream, measurements_attributes)

 expect(Measurement.count).to eq(1)
```

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    expect(Measurement.count).to eq(1)
+  end
+
+  it "creates a mesurement with utc time for fixed sessions" do
```
3v0k4: Personally I follow the convention of putting context first in the test name. In this case it would be "when session is fixed it creates a measurement with utc time". But I'm fine with this. Also, watch out for the spelling in measurement.

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    stream = create_stream!("FixedSession")
+    MeasurementsCreator.call(stream, measurements_attributes)
+
+    expect(Measurement.first.utc_time).to be_within(1.second).of Time.now
```
3v0k4: This is a bit hacked. Ping me if you want to have a chat about this.

spec/services/measurements_creator_spec.rb
3v0k4: I'd suggest you move this in a private method at the end. The attributes don't really matter so it's better "hidden" at the end

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  it "creates a mesurement without utc time for mobile sessions" do
+    stream = create_stream!("MobileSession")
+    MeasurementsCreator.call(stream, measurements_attributes)
+
+    expect(Measurement.first.utc_time).to be_nil
+  end
+
+  private
+
+  def create_stream!(session_type  = "MobileSession")
```
3v0k4: Have you considered doing something like this (https://github.com/HabitatMap/AirCasting/blob/staging/spec/services/csv/export_sessions_to_csv_spec.rb#L71)?

That way you have the same api of active_record (you pass a hash) and all the defaults. That way you are explicit with the important stuff `session_type` and hide all of the noise

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    Stream.create!(
+      sensor_package_name: "AirBeam2:00189610719F",
+      sensor_name: "AirBeam2-F",
+      measurement_type: "Temperature",
+      unit_name: "Fahrenheit",
+      session: create_session!(session_type),
```
3v0k4: Have you considered keeping create_stream and create_session independent and compose them in another method or directly in the test?

db/schema.rb
```diff
@@ -32,6 +32,7 @@
     t.integer  "stream_id"
     t.integer  "milliseconds",                                   :default => 0
     t.float    "measured_value"
+    t.datetime "utc_time"
```
3v0k4: I'd try to think of a more descriptive name. `utc_time` doesn't say what time that is

app/services/measurements_creator.rb
```diff
@@ -2,6 +2,13 @@ class MeasurementsCreator
   SLICE_SIZE = 500
 
   def self.call(stream, measurements_attributes)
```
3v0k4: In the spirit of unlegacing the app we should probably cover the entire method in tests

app/services/measurements_creator.rb
```diff
@@ -2,6 +2,13 @@ class MeasurementsCreator
...
+    if stream.session.type == "FixedSession"
```
3v0k4: I haven't studied the code paths much. Wouldn't it be enough to have this in `#call`? You wouldn't need the `if` there cause there's an early return there

app/services/measurements_creator.rb
```diff
@@ -2,6 +2,13 @@ class MeasurementsCreator
...
+      measurements_attributes = measurements_attributes.map do |measurement_attributes|
+        measurement_attributes[:utc_time] = Time.now.utc
```
3v0k4: Not sure in this case but may be better `Time.current` (https://robots.thoughtbot.com/its-about-time-zones)

app/services/measurements_creator.rb
AkkarinForest: You mean `MeasurementCreator.new.call`? That's not possible - cause there is also `AsyncMeasurementsCreator` (it's called on line 9 in this file). And this async creator then calls directly `MeasurementCreator.new.call` so it's too late.

3v0k4: Good point! We should check if the async is called only for mobile measurements. I have a gut feeling that the if could be related to the type of session.

spec/services/measurements_creator_spec.rb
AkkarinForest: I like the idea of having the same api. I could do something like that:
```
session = create_session!(type: "FixedSession")
stream = create_stream!(session: session)
```
but I am not sure about passing all the defaults, they make noise that's irrelevant to this test case.

spec/services/measurements_creator_spec.rb
AkkarinForest: Not in the current form. Couse `Steram.build_measurements` uses the database. But maybe this should be a unit test and I should stub all methods in stream class??? But I'm not sure I know how to do this.

db/schema.rb
AkkarinForest: I don't know... 
You mean it should be more specific and mention that it's for fixed sessions only? What if in the future someone wants to use it for mobile? They would have to rename. I googled and renaming columns is one of the migrations that causes down times.

app/models/stream.rb
```diff
@@ -180,4 +180,8 @@ def has_bounds?
     max_longitude.present? &&
     min_longitude.present?
   end
+  
+  def fixed?
+    self.session.type == "FixedSession"
```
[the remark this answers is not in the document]
3v0k4: Also, watch out cause this method potentially executes an additional call to db to get the session. In DDD they solve this sort of problems by discovering what's the "root aggregate" in a domain and then executing all "accesses" to the domain through that. In other words, we would prolly always call things on a session and internally the session would be in charge of loading everything from db once for all, updating the state and persisting back

app/services/measurements_creator.rb
```diff
@@ -19,12 +21,19 @@ def call(stream, measurements_attributes)
     stream.build_measurements!(measurements_attributes)
     stream.after_measurements_created
 
-    return if stream.session.type == 'FixedSession'
```
3v0k4: This is a really nice step forward!

app/services/measurements_creator.rb
```diff
@@ -19,12 +21,19 @@ def call(stream, measurements_attributes)
...
     streams_repository.calc_average_value!(stream)
   end
 
   private
 
   attr_reader :streams_repository
+
+  def self.add_arrival_time(measurements_attributes)
+    measurements_attributes.map do |measurement_attributes|
+      measurement_attributes[:arrival_utc_time] = Time.current.utc
```
3v0k4: Does `Time.current` give you non utc? If that's the case I think the Rails config is wrong. Maybe it would be better to fix the config and drop the `.utc`?

db/migrate/20181114165757_add_utc_time_to_measurement.rb
```diff
@@ -0,0 +1,5 @@
+class AddUtcTimeToMeasurement < ActiveRecord::Migration
+  def change
+    add_column :measurements, :arrival_utc_time, :datetime
```
3v0k4: It's interesting that thinking about the name brought us to something similar to what is formalized https://en.wikipedia.org/wiki/Time_of_arrival

3v0k4: See for yourself if that formalization is a better naming. In any case, I feel what is in place now is already a good one

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,148 @@
+require "spec_helper"
+
+describe MeasurementsCreator do
+  describe "#self.call" do
```
3v0k4: In Ruby lingo you use a `.` for class methods and `#` for instance methods. So you could go with `.call` here

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+    before() do
```
3v0k4: `()` not needed ;)

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+      @arrival_utc_time = Time.utc(2018, 11, 19)
```
3v0k4: The `@arrival_utc_time` var is used in only one test. I'd suggest you stub directly there to improve readability and have less noise in the other tests

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+      allow(Time).to receive(:current).and_return(@arrival_utc_time)
```
3v0k4: I like this approach!

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+      it "when sessions is fixed creates a measurement with utc time" do
+        session = create_session!(type: "FixedSession")
+        stream = create_stream!(session: session)
+
+        MeasurementsCreator.call(stream, single_measurement_attributes)
+
+        expect(Measurement.first.arrival_utc_time).to eq(@arrival_utc_time)
+      end
+
+      it "when session is moblie creates a measurement without utc time" do
```
3v0k4: typo in moblie ;)

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+    end
+
+    context "when there are more than one measurements" do
+      it "schedules the creation of measurements" do
+        session = create_session!
+        stream = create_stream!(session: session)
+
+        MeasurementsCreator.call(stream, multiple_measurements_attributes)
+
+        assert_equal 1, AsyncMeasurementsCreator.jobs.size
```
3v0k4: For consistency you could use
```
expect().to be()
```

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+      end
+    end
+  end
+
+  describe "#call" do
+    let(:streams_repository) { double(StreamsRepository.new) }
+    let(:measurements_creator) { MeasurementsCreator.new(streams_repository) }
+    let(:stream) { double(Stream.new) }
+
+    before() do
```
3v0k4: `()`

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+
+      it "performs a callback after measuremnets are created" do
+        expect(stream).to receive(:after_measurements_created).with(no_args)
+
+        measurements_creator.call(stream, single_measurement_attributes)
+      end
+    end
+
+    context "for sessions that are not fixed" do
+      before() do
```
3v0k4: `()`

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+        allow(stream).to receive(:fixed?).and_return(false)
+      end
+
+      it "calculates bounding box of stream" do
+        expect(streams_repository).to receive(:calc_bounding_box!).with(stream)
+
+        measurements_creator.call(stream, single_measurement_attributes)
+      end
+
+      it "calculattes average value of stream" do
```
3v0k4: typo in calculattes

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+    }
+  end
+
+  def create_stream!(attributes)
+    Stream.create!(
+      sensor_package_name: "AirBeam2:00189610719F",
+      sensor_name: "AirBeam2-F",
+      measurement_type: "Temperature",
+      unit_name: "Fahrenheit",
+      session: attributes[:session],
```
3v0k4: If you are always expecting the key / value to be there `#fetch` is a better way. It raises if the key is missing. A fast failure can save hours of debugging.

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+      title: "Example Session",
+      user: User.new,
+      uuid: "845342a6-f9f4-4835-86b3-b100163ec39a",
+      calibration: 100,
+      offset_60_db: 0,
+      start_time: DateTime.current,
+      start_time_local: DateTime.current,
+      end_time: DateTime.current,
+      end_time_local: DateTime.current,
+      type: attributes[:type]
```
3v0k4: As above with `#fetch`

spec/services/measurements_creator_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+
+        assert_equal 1, AsyncMeasurementsCreator.jobs.size
+      end
+    end
+  end
+
+  describe "#call" do
+    let(:streams_repository) { double(StreamsRepository.new) }
+    let(:measurements_creator) { MeasurementsCreator.new(streams_repository) }
+    let(:stream) { double(Stream.new) }
```
3v0k4: Didn't know you can pass an object inside `double`. What's the advantage over just `double`?

app/models/stream.rb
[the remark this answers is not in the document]
AkkarinForest: Sure, I like the ideas from your first two comments. But in the last one, I am not sure what exactly do you suggest I do?

spec/services/measurements_creator_spec.rb
AkkarinForest: It's also used on line 7

spec/services/measurements_creator_spec.rb
AkkarinForest: Sorry, it's a mistake. I didn't know the exact syntax and I found a similar example somewhere, cause I couldn't find the documentation for RSpec 2.99. 
Anyway, now I read the documentation 3.8 and it turns out the first argument passed to double is just its name, that's displayed if the test fails. I checked and it works the same in our rspec, so I will change this to `double("stream")`: 
screen shot 2018-11-20 at 11 46 04
screen shot 2018-11-20 at 11 46 33

app/models/stream.rb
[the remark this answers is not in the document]
3v0k4: For now I'd say nothing about the last comment. Maybe you could check (or we can together) if the session is already in memory or if that hits db again

spec/services/measurements_creator_spec.rb
3v0k4: I see but only one test is using that, correct?

PR #57, opened 2018-11-23
Add crowd layer to mobile map: ui

3v0k4:
This is the first step towards adding the crowd map in the mobile tab. In practice, in the mobile map it will be possible to enable (checkbox) the crow map layer and regulate the resolution as in the crowd map.

The section in the html is wrapped in `<section ng-show="false">` so that I can merge and go on with the work.

### DIFF
--- a/app/assets/javascripts/code/controllers/_mobile_sessions_map_ctrl.js
+++ b/app/assets/javascripts/code/controllers/_mobile_sessions_map_ctrl.js
@@ -93,5 +93,14 @@ export const MobileSessionsMapCtrl = (
       map.goToAddress(newValue.location);
     }, true);
 
+  $scope.$watch("storage.data.crowdMap", function(newValue, oldValue) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/code/controllers/_mobile_sessions_map_ctrl.js
```diff
@@ -41,17 +41,22 @@ export const MobileSessionsMapCtrl = (
         $scope.$digest();
       });
     }
-    _.each(['sensor', 'location', 'usernames'], function(name) {
+
+    ['sensor', 'location', 'usernames', 'layers'].forEach(function(name) {
```
3v0k4: Tested and refactored to remove `_`. In the long term I'd like to remove that dependency. I'd prefer to use Ramda or something similar

3v0k4:
@AkkarinForest I merged but please feel free to go through the code

PR #59, opened 2018-11-28
Mobile: crowd rectangles are limited to the sessions in the list

3v0k4:
Instead of getting the avarages for the crowd map layer for all the sessions in the viewport, now the request is limited to the sessions in the list. Since the session list is paginated this means only the first page is requested.


app/assets/javascripts/code/controllers/_mobile_sessions_map_ctrl.js
```diff
@@ -14,8 +14,7 @@ export const MobileSessionsMapCtrl = (
   functionBlocker,
```
3v0k4: Moved all the logic to sessions_list_ctrl (it will make more sense later when fixed will have the crowd layer and crowd map will be removed)

app/assets/javascripts/code/services/_sessions_utils.js
```diff
@@ -0,0 +1,96 @@
+import _ from 'underscore';
```
3v0k4: This is a new file cause I've decoupled from angular with the usual trick (ie `_`)

spec/controllers/averages_controller_spec.rb
```diff
@@ -1,47 +1,37 @@
...
-
-    before(:each) do
-      info = double
-      expect(AverageInfo).to receive(:new).with(expected).and_return(info)
-      expect(info).to receive(:as_json).and_return(result)
-    end
-
-    it { is_expected.to be_successful }
+  it "#index adds defaults to missing params and returns result from AverageInfo#as_json" do
+    q = {
```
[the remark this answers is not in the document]
3v0k4: Not sure too but didn't have any better ideas. When you are reviewing feel free to propose alternatives. (Easier to implement for me :))

PR #61, opened 2018-12-05
Add photos urls to the notes csv files.

AkkarinForest:
BEFORE DEPLOYING provision the container with a new secret variable containing host url. [Related PR](https://github.com/LunarLogic/playing-zone/pull/229)

I had to switch from using raw sql to using active record in app/services/csv/repository.rb, because paperclip gem only works with active record objects. 
As a consequence, the append_notes_content_spec is touching the db. Any suggestions on how to improve that are very welcome. I remember there is an option to use build instead of create, but I think it doesn't work in our rails version :(

Also, I created a separate method to find session title. I know that it's running one more extra query to the db, but looking for a session title in the query that selects notes would mean selecting the title n times, where n is the number of notes. So I guess that's not good either.


app/services/csv/create_files.rb
```diff
@@ -84,11 +77,11 @@ def initialize(append_notes_content = Csv::AppendNotesContent.new)
     @append_notes_content = append_notes_content
   end
 
-  def call(data)
-    session_title = data.notes.first["session_title"] || ""
-    filename = "notes_from_#{session_title.parameterize('_')}_#{data.session_id}__"
+  def call(notes, session_title, session_id)
+    session_title = session_title || ""
```
AkkarinForest: I'm not sure if it's possible that the session title is nil, but we were accounting for that option before so I left it.

3v0k4: If we don't have a db not null constraint unfortunately we have to

spec/services/csv/append_notes_content_spec.rb
```diff
@@ -5,50 +5,39 @@
...
-      "notes" => notes,
-      "session_id" => 123
+  def create_note!
+    Note.create!(
+      session: Session.new,
+      text: "Example Note",
+      date: date,
+      latitude: latitude,
+      longitude: longitude,
+      photo: File.new("#{Rails.root}/spec/support/example_photo.jpg"),
```
3v0k4: Is this file removed after running the tests?

spec/services/csv/export_sessions_to_csv_spec.rb
```diff
@@ -51,27 +51,28 @@
...
       longitude: BigDecimal.new("-73.97631499"),
+      photo: File.new("#{Rails.root}/spec/support/example_photo.jpg"),
       session: session,
     )
 
     zip_path = @subject.call([session.id])
 
     Zip::File.open(zip_path) do |zip_file|
-      actual_contents = zip_file.entries.map { |entry| entry.get_input_stream.read }
+      actual_contents = zip_file.entries.map { |entry| entry.get_input_stream.read }.last
```
3v0k4: Maybe we could extract a method like `last_file_content(zip_file)` or `file_contents(zip_file).last`

spec/services/csv/append_notes_content_spec.rb
```diff
@@ -5,50 +5,39 @@
...
-
-    expected_content = [
-      expected_headers,
-      ["Example Note","2018-08-20T11:16:44",latitude,longitude],
-      ["Example Note 2","2018-08-20T11:16:44",latitude,longitude],
-    ]
-    expect(actual_content).to eq(expected_content)
+    expected_note = ["Example Note","2018-08-20T11:16:44",latitude,longitude,photo_url]
+    expect(actual_note[0..-2]).to eq(expected_note[0..-2])
+    expect(actual_note[-1]).to match(expected_note[-1])
```
3v0k4: Could that be only one expect?

app/services/csv/append_notes_content.rb
```diff
@@ -1,26 +1,33 @@
...
 
   def format_time(date)
     date.strftime("%FT%T")
   end
+
+  def format_url(note)
+    return "No photo was attached" unless note.photo_file_name
+
+    current_host_url = Rails.application.config.secrets.fetch('HOST_URL')
+    current_host_url.to_s + note.photo.url
```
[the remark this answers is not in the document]
AkkarinForest: Yes. That is why I didn't add a `/` at the end of HOST_URL, but now I'm thinking that maybe it's better to have it in both places. The url will work fine with two slashes. And just in case someone will use `HOST_URL` in another place, it's safer if it contains the final `/`.

spec/services/csv/append_notes_content_spec.rb
AkkarinForest: I didn't find a way to make it work. The last entry in this array is a URL and it contains a hash and I thought that using a regexp to match this URL is better than hardcoding the hash in the expected value. 
But in order to compare regexps I need to use `match` on the string. Using `match` on the entire array doesn't work the same. 
If you have an idea how to merge them together, please share.

app/services/csv/append_notes_content.rb
```diff
@@ -1,26 +1,33 @@
...
   end
 
   def format_time(date)
     date.strftime("%FT%T")
   end
+
+  def format_url(note)
+    return "No photo was attached" unless note.photo_file_name
+
+    current_host_url = Rails.application.config.secrets.fetch('HOST_URL')
```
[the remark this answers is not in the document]
AkkarinForest: That's a very good suggestion.

spec/services/csv/append_notes_content_spec.rb
AkkarinForest: No! 😱 good catch!

PR #62, opened 2018-12-06
Mobile crowd rectangles info

AkkarinForest:
When a user selects a rectangle on mobile crowd map, they see a window with statistics for this rectangle.

Also, I added one more test and I kept it in a separate commit, so if you don't like it we can throw it away :) If it's ok, then I can squash the commits together.


app/assets/javascripts/tests/_update_crowd_map_layer.test.js
```diff
@@ -86,6 +86,44 @@ test('when buildQueryParamsForAverages return false it does not request the aver
...
+  };
+  const $http = mockHttp({ shouldFail: false });
+  const $location = { path: () => "/map_sessions" };
+  const infoWindow = mock('show');
+  const queryParamsForAverages = {};
+  const service = _updateCrowdMapLayer({ map, $http, $location, infoWindow });
+
+  service.call();
+
+  t.true(infoWindow.wasCalledWith2(queryParamsForAverages));
```
3v0k4: There's a problem with this test. That I don't see the relationship between `buildQueryParamsForAverages.call` and `infoWindow.wasCalledWith2({})`. But you are almost there. You could also test `wasCalledWith3`

PR #64, opened 2018-12-11
Show mobile sessions starting points

AkkarinForest:
This is the first step in marking mobile session starting points. I added the logic that shows the points, default blue for every session for now.

I guess it should have more tests, but I'm having trouble writing them, especially for drawSession, since there isn't anything similar that I can get inspired by :D

Also, I need to remember to tests how it behaves when crowd map for mobile is in place. Cause it wasn't easy to trace all the places where stuff are drawn and unDrawn, so it's possible I missed something.

### DIFF
--- a/app/assets/javascripts/code/services/_draw_session.js
+++ b/app/assets/javascripts/code/services/_draw_session.js
@@ -10,7 +10,7 @@ export const drawSession = (
   };
 
   DrawSession.prototype = {
-    drawMobileSession: function(session, bounds) {
+    drawMobileSession: function(session) {
       if(!session || !session.loaded || !sensors.anySelected()){
         return;
       }
@@ -41,7 +41,7 @@ export const drawSession = (
       session.drawed = true;
     },
 
-    drawFixedSession: function(session, bounds) {
+    drawFixedSession: function(session) {
       this.undoDraw(session);
       session.markers = [];
       session.noteDrawings = [];
@@ -50,7 +50,7 @@ export const drawSession = (
       var level;
 
       if (sensors.anySelected() && !sensors.tmpSelected() && session.last_hour_average) {
-        level = this.calculateHeatLevel(session.last_hour_average);
+        level = this._calculateHeatLevel(session.last_hour_average);
       } else {
         level = 0;
       }
@@ -61,7 +61,21 @@ export const drawSession = (
       return session.markers;
     },
 
-    calculateHeatLevel: function(value) {
+    drawMobileSessionStartPoint: function(session) {
+      if(!session || !sensors.anySelected()){
+        return;
+      }
+      this.undoDraw(session);
+
+      session.markers = [];
+      var markerOptions = {title: session.title, zIndex: 0};
+      var lngLatObject = { longitude: session.starting_longitude, latitude: session.starting_latitude }
+      session.markers.push(map.drawMarker(lngLatObject, markerOptions, null, "0"));
+
+      session.drawed = true;
+    },
+
+    _calculateHeatLevel: function(value) {
       return heat.getLevel(value);
     },
 
--- a/app/assets/javascripts/code/services/_draw_session.js
+++ b/app/assets/javascripts/code/services/_draw_session.js
@@ -50,7 +50,7 @@ export const drawSession = (
       var level;
 
       if (sensors.anySelected() && !sensors.tmpSelected() && session.last_hour_average) {
-        level = this._calculateHeatLevel(session.last_hour_average);
+        level = this.calculateHeatLevel(session.last_hour_average);
       } else {
         level = 0;
       }
@@ -62,21 +62,11 @@ export const drawSession = (
     },
 
     drawMobileSessionStartPoint: function(session) {
-      if(!session || !sensors.anySelected()){
-        return;
-      }
       this.undoDraw(session);
 
-      session.markers = [];
-      var markerOptions = {title: session.title, zIndex: 0};
-      var lngLatObject = { longitude: session.starting_longitude, latitude: session.starting_latitude }
+      var markerOptions = map.defaultMarkerOptions;
+      var lngLatObject = { longitude: session.startingLongitude, latitude: session.startingLatitude }
       session.markers.push(map.drawMarker(lngLatObject, markerOptions, null, "0"));
-
-      session.drawed = true;
-    },
-
-    _calculateHeatLevel: function(value) {
-      return heat.getLevel(value);
     },
 
     undoDraw: function(session, mapPosition) {
@@ -125,4 +115,8 @@ export const drawSession = (
     }
   };
   return new DrawSession();
-}
+};
+
+function calculateHeatLevel(value) {
+  return heat.getLevel(value);
+};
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/code/services/_mobile_sessions.js
```diff
@@ -95,12 +98,19 @@ export const mobileSessions = (
 
     reSelectSession: function(id) {
```
AkkarinForest: Not sure what this is doing and if I should clear the starting points here like I do in selectSession

app/models/session.rb
```diff
@@ -209,6 +209,9 @@ def as_json(opts=nil)
       end
     end
 
+    first_measurement = fetch_first_measurement_coordinates
+    res.merge!(:starting_longitude => first_measurement.longitude, :starting_latitude => first_measurement.latitude)
+
```
AkkarinForest: this file is work in progress, I wanted to focus on js first

app/assets/javascripts/code/services/_draw_session.js
```diff
@@ -61,7 +61,21 @@ export const drawSession = (
...
-    calculateHeatLevel: function(value) {
+    drawMobileSessionStartPoint: function(session) {
+      if(!session || !sensors.anySelected()){
+        return;
+      }
+      this.undoDraw(session);
+
+      session.markers = [];
+      var markerOptions = {title: session.title, zIndex: 0};
+      var lngLatObject = { longitude: session.starting_longitude, latitude: session.starting_latitude }
```
3v0k4: We should be using camelCase for Js variables

app/assets/javascripts/code/services/_draw_session.js
```diff
@@ -61,7 +61,21 @@ export const drawSession = (
       return session.markers;
     },
 
-    calculateHeatLevel: function(value) {
+    drawMobileSessionStartPoint: function(session) {
```
3v0k4: There's a lot of smelly things in here:
- `if(!session || !sensors.anySelected())` maybe the function shouldn't even be called in the first place in those cases?
- `session.markers = []` maybe this is not this module's responsibility (ie maybe a session module should be doing this?)
- `var markerOptions = {title: session.title, zIndex: 0};` this seems like a map responsibility
- `session.markers.push(map.drawMarker(lngLatObject, markerOptions, null, "0"));` mixed session / map responsibility?
- `session.drawed = true;` again session

The code is really procedural, maybe we could try to separate the responsibilities better. I know this is what we have in place and it's difficult to understand what's going on / how to decouple. But this is where we can make the difference

app/assets/javascripts/code/services/_fixed_sessions.js
```diff
@@ -129,7 +129,7 @@ export const fixedSessions = (
 
     drawSessionsInLocation: function() {
       map.markers = [];
```
3v0k4: This is prolly a responsibility of the map module

app/assets/javascripts/code/services/_fixed_sessions.js
```diff
@@ -129,7 +129,7 @@ export const fixedSessions = (
...
-      _(this.get()).each(session => drawSession.drawFixedSession(session, boundsCalculator(this.sessions)));
+      _(this.get()).each(session => drawSession.drawFixedSession(session));
```
3v0k4: If possible it would be great if we could reduce the usage of `_` so we can move to something like ramda. Example
```
(this.get() || []).forEach(drawSession.drawFixedSession)
```
But prolly `this.get()` should always return an array (ie `(this.get() || [])` not needed)?

app/assets/javascripts/code/services/_mobile_sessions.js
```diff
@@ -81,26 +83,34 @@ export const mobileSessions = (
...
+      drawSession.clear(this.sessions);
       const callback = (session, allSelected) => (data) => {
-        const draw = () => drawSession.drawMobileSession(session, boundsCalculator(allSelected));
+        const draw = () => drawSession.drawMobileSession(session);
         sessionsUtils.onSingleSessionFetch(session, data, draw);
       }
       this._selectSession(id, callback);
     },
 
+    drawSessionsInLocation: function() {
```
3v0k4: This is really similar to the fixed one.

Also, keep in mind the `sessionUtils` refactor was mine and I hate myself for doing it. I should have eliminated the `this` instead of passing it to `sessionsUtils`. The idea was to keep there the common code.

app/assets/javascripts/tests/_mobile_sessions.test.js
```diff
@@ -295,6 +295,20 @@ test('hasSelectedSessions with selected session returns true', t => {
   t.end();
 });
 
+test('drawSessionsInLocation calls drawSession.drawMobileSessionStartPoint', t => {
+  const session = "session"
+  const sessions = { session1: session }
+  const sessionsUtils = { get: () => (sessions) };
```
3v0k4: You can drop the `( )` in `get: () => (sessions)`. You need them only when it's `get: () => ({sessions})`

app/models/session.rb
3v0k4: 👍

app/assets/javascripts/tests/_mobile_sessions.test.js
```diff
@@ -295,6 +295,20 @@ test('hasSelectedSessions with selected session returns true', t => {
   t.end();
 });
 
+test('drawSessionsInLocation calls drawSession.drawMobileSessionStartPoint', t => {
+  const session = "session"
+  const sessions = { session1: session }
```
3v0k4: I'm not totally sure why the key is called `session1` but I trust you ;)

app/assets/javascripts/code/services/_draw_session.js
```diff
@@ -50,7 +50,7 @@ export const drawSession = (
       var level;
 
       if (sensors.anySelected() && !sensors.tmpSelected() && session.last_hour_average) {
-        level = this.calculateHeatLevel(session.last_hour_average);
+        level = this._calculateHeatLevel(session.last_hour_average);
```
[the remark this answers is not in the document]
AkkarinForest: Sure, I wasn't sure what convention are we using and I saw some `_` in another file. But that's better, let's try to divide functions this way in all the files that we are editing.

app/assets/javascripts/code/services/_fixed_sessions.js
AkkarinForest: `[]` is not needed

app/assets/javascripts/code/services/_draw_session.js
AkkarinForest: They come from rails part, that's why they are snake, but I guess I can change them there and have camel here.

3v0k4: For now it's fine. We should use serializers on the rails side doing that automatically

app/assets/javascripts/code/services/_fixed_sessions.js
3v0k4: I think that was done case `null`, `undefined` or a single value become array with `_()`. Example `_(1) == [1]` or `_(null) == []`

app/assets/javascripts/code/services/_mobile_sessions.js
AkkarinForest: I changed and it's different from fixed now. But we should probably visit and refactor fixed someday as well, then we can also extract similar parts to utils.

PR #65, opened 2018-12-13
Parameter sensor prioritization

3v0k4:
* Default to Particulate Matter / AB2 PM2.5 in mobile tab
*  Apply following order to sensors in dropdown

Default Parameter: Particulate Matter 
Default Sensor: AirBeam2-PM2.5 (ug/m3)
Prioritized Sensor: AirBeam2-PM1 (ug/m3)
Prioritized Sensor: AirBeam2-PM10 (ug/m3)
Prioritized Sensor: AirBeam-PM (ug/m3)

Prioritized Parameter: Relative Humidity
Default Sensor: AirBeam2-RH (%)
Prioritized Sensor: AirBeam-RH (%)

Prioritized Parameter: Temperature
Default Sensor: AirBeam2-F (F)
Prioritized Sensor: AirBeam-F (F)

Prioritized Parameter: Sound Levels
Default Sensor: Phone Microphone 

Prioritized Parameter: All
Default Sensor: All


app/assets/javascripts/code/services/_sensors.js
```diff
@@ -163,3 +148,76 @@ const max = (valueOf, xs) => {
...
+      sensor_name:      "Phone Microphone",
+      unit_symbol:      "dB"
+    })
+  };
+
+  return DEFAULT_IDS[parameter.id] || max(sensor => sensor.session_count, sensors).id;
+};
+
+export const buildAvailableParameters = sensors => {
+  const uniq = (v, i, a) => a.indexOf(v) === i;
```
AkkarinForest: Does v, i, a stand for something? If yes then could you use the full word? If they are meant to be generic then why not a, b, c? 
Also, is this the same as in line 219?

app/assets/javascripts/code/services/_sensors.js
```diff
@@ -58,7 +50,7 @@ export const sensors = (params, $http) => {
         console.log('initSelected() - sensorId is NOT null')
       }
       this.selectedParameter = self.findParameterForSensor(self.selected());
-      this.availableSensors = self.findAvailableSensorsForParameter(self.selectedParameter);
+      this.availableSensors = findAvailableSensorsForParameter(sort, this.sensors, self.selectedParameter);
```
AkkarinForest: Why do we sometimes use this and later self? `this.sensors, self.selectedParameter`. Could be just `this` everywhere in this file?

app/assets/javascripts/code/services/_sensors.js
```diff
@@ -134,7 +119,7 @@ export const sensors = (params, $http) => {
       this.selectedParameter = parameterForSensor;
     },
     buildSensorId: function(sensor) {
-      return sensor.measurement_type + "-" + sensor.sensor_name + " (" + sensor.unit_symbol + ")";
+      return buildSensorId(sensor);
```
AkkarinForest: What is the reason behind hiding this in a private function? And we could do the same for all other functions, or is this some special case?

app/assets/javascripts/code/services/_sensors.js
3v0k4: We prolly could but I'm leaving that for another round. Good catch nonetheless

app/assets/javascripts/code/services/_sensors.js
3v0k4: I needed to call `buildSensorId` from one of the private functions and I couldn't unless `buildSensorId` was private too

app/assets/javascripts/code/services/_sensors.js
3v0k4: They stand for value, index, array. I totally agree with you but this time I'll leave that for later. Again, I've refactored a lot and don't want to spend more time on cosmetics for this round

3v0k4:
Thanks for the review and sorry I haven't applied all your good points. Yeah I can. Also, in this PR you could have just reviewed the single commit that introduced most of the changes (the last) to avoid the noise

AkkarinForest:
I looked at the commit separately, but I couldn't just review the last one, I need to check the rest before I knew that they were only noise and not important changes, right? 
Anyway, it either takes more time for the author to split the changes or more time for the reviewer to compare them, so it's not really a big difference.

PR #66, opened 2018-12-14
Mobile session start point heat levels

AkkarinForest:
Starting points change colors depending on average measurements level.

### DIFF
--- a/app/models/session.rb
+++ b/app/models/session.rb
@@ -209,8 +209,10 @@ class Session < ActiveRecord::Base
       end
     end
 
-    first_measurement = fetch_first_measurement_coordinates
-    res.merge!(:startingLongitude => first_measurement.longitude, :startingLatitude => first_measurement.latitude)
+    if type == "MobileSession"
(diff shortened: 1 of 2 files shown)


app/models/session.rb
```diff
@@ -209,8 +209,10 @@ def as_json(opts=nil)
       end
     end
 
-    first_measurement = fetch_first_measurement_coordinates
-    res.merge!(:startingLongitude => first_measurement.longitude, :startingLatitude => first_measurement.latitude)
+    if type == "MobileSession"
+      first_measurement = fetch_first_measurement_coordinates
+      res.merge!(:startingLongitude => first_measurement.longitude, :startingLatitude => first_measurement.latitude, :average => measurements.average(:value))
```
3v0k4: I'd prefer to keep ruby in snake_case and js in camelCase. We should have another layer doing the translation

3v0k4: I guess this should be optimized? Also this method is screaming to be refactored!

PR #67, opened 2018-12-17
Fix parameter sensor dropdowns

3v0k4:
The tab could be opened in three different ways:
- by opening `/map` (ie clicking on the maps link in the menu)
- by switching to fixed / mobile
- by copy / pasting a url (eg `/map#/map_fixed_sessions?map=%7B%22zoom%22:5,%22lat%22:37.09024,%22lng%22:-95.71289100000001,%22mapType%22:%22terrain%22%7D&data=%7B%22sensorId%22:%22Particulate%20Matter-AirBeam2-PM2.5%20(%C2%B5g%2Fm%C2%B3)%22,%22location%22:%7B%22address%22:%22%22,%22indoorOnly%22:false,%22streaming%22:true%7D,%22tags%22:%22%22,%22usernames%22:%22%22,%22time%22:%7B%22timeFrom%22:-60,%22timeTo%22:1379,%22dayFrom%22:1,%22dayTo%22:365,%22yearFrom%22:2017,%22yearTo%22:2018%7D,%22heat%22:%7B%22highest%22:150,%22high%22:55,%22mid%22:35,%22low%22:12,%22lowest%22:0%7D%7D&selectedSessionIds=%5B%5D&tmp=%7B%22selectedSensorId%22:%22%22%7D`)

Before this PR the app always loaded as parameter / sensor the default (ie particulate matter / ab2 pm 2.5) when in the url no `sensorId` was specified or if `sensorId` was "". Unfortunately, all / all was equivalent to `sensorId === ""` which loaded particulate matter / ab2 pm 2.5 and not all / all. That happened for example if a user selected all / all and then pasted the link to navigate back to the page.

Now the code persists `sensorId === "all"` when all / all is selected, therefore pasting the link works.

Also, the smaller commit fixed another bug. In particular, on the first load the sensors weren't filtered by parameter when the page was loaded. In fact, the default selection was in place particulate matter / ab2 pm 2.5 but the sensors in the dropdown contained all of them (eg humidity, sound level)

### DIFF
--- a/app/assets/javascripts/code/services/_sensors.js
+++ b/app/assets/javascripts/code/services/_sensors.js
@@ -1,5 +1,8 @@
 import _ from 'underscore';
 
+const ALL_SENSOR = { id: "all", select_label: "All", session_count: 0 };
+const ALL_PARAMETER = { id: "all", label: "All" };
+
 export const sensors = (params, $http) => {
   var Sensors = function() {
     this.sensors = {};
@@ -50,18 +53,15 @@ export const sensors = (params, $http) => {
         console.log('initSelected() - sensorId is NOT null')
       }
       this.selectedParameter = self.findParameterForSensor(self.selected());
-      this.availableSensors = findAvailableSensorsForParameter(sort, this.sensors, self.selectedParameter);
+      this.availableSensors = [ALL_SENSOR].concat(findAvailableSensorsForParameter(sort, this.sensors, self.selectedParameter));
     },
 
-    //selected in dropdown
+    //selected sensor in dropdown
     selected: function() {
       return this.sensors[params.get('data').sensorId];
     },
     selectedId: function() {
-      if(!this.selected()){
-        return;
-      }
-      return this.selected().id;
+      return this.selected() ? this.selected().id : ALL_SENSOR.id;
     },
     //used when "all" sensors are choosen
     tmpSelected: function() {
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/code/services/_sensors.js
```diff
@@ -142,15 +143,13 @@ export const sensors = (params, $http) => {
...
-const hasChangedToAll = newValue => !newValue;
-
 const max = (valueOf, xs) => {
   const reducer = (acc, x) => valueOf(x) > valueOf(acc) ? x : acc;
   return xs.reduce(reducer, xs[0]);
 };
 
 export const findAvailableSensorsForParameter = (sort, sensors, parameter) =>
-  parameter ?
+  parameter.id !== ALL_PARAMETER.id ?
```
AkkarinForest: I'm always having troubles with reading negation in `ifs`. What do you think about doing `parameter.id == ALL_PARAMETER.id` and switching the order of lines 153, 154?

app/assets/javascripts/code/services/_sensors.js
```diff
@@ -1,5 +1,8 @@
 import _ from 'underscore';
 
+const ALL_SENSOR = { id: "all", select_label: "All", session_count: 0 };
+const ALL_PARAMETER = { id: "all", label: "All" };
```
AkkarinForest: I'm not sure if all_sensor would be understandable for someone reading this in the future and not familiar with the details of this problem. How about `ALL_SENSORS_ENTRY` or something like that?

app/assets/javascripts/code/services/_sensors.js
```diff
@@ -93,38 +93,39 @@ export const sensors = (params, $http) => {
...
       params.update({selectedSessionIds: []});
-      if (hasChangedToAll(selectedParameter)) {
-        this.availableSensors = sort(Object.values(this.sensors));
-        params.update({data: {sensorId: ""}});
+      if (selectedParameter.id === ALL_PARAMETER.id) {
+        this.availableSensors = [ALL_SENSOR].concat(sort(Object.values(this.sensors)));
+        params.update({data: {sensorId: ALL_SENSOR.id}});
       } else {
-        this.availableSensors = sort(Object.values(this.sensors).filter(sensor => sensor.measurement_type === selectedParameter.id));
+        this.availableSensors = [ALL_SENSOR].concat(sort(Object.values(this.sensors).filter(sensor => sensor.measurement_type === selectedParameter.id)));
```
AkkarinForest: this `[ALL_SENSOR].concat(` is repeated in a few places. Maybe it would make sense to extract it into a function like `addAllSensor` or something.

app/assets/javascripts/code/services/_sensors.js
```diff
@@ -93,38 +93,39 @@ export const sensors = (params, $http) => {
...
     },
     buildSensorId: function(sensor) {
       return buildSensorId(sensor);
     },
+    // dropdown
     onSensorsSelectedIdChange: function(newValue, oldValue, callback) {
       console.log("onSensorsSelectedIdChange - ", newValue, " - ", oldValue);
 
-      if(hasChangedToAll(newValue)) return;
+      if(!newValue || newValue === ALL_SENSOR.id) return;
```
AkkarinForest: is it still possible that `!newValue` is true? Since we have ALL_SENSOR default now? I don't know, just asking :)

3v0k4:
Thanks for the review. Unfortunately it seems I need to go deeper with this code. Trying to refactor just a bit solved some bugs but introduced others. `Watch`s being triggered by both `storage` and `params` is complex (and it shouldn't be that way).

In any case, I'll keep your comments in mind for the next iteration. I see that you are getting pickier and pickier. I appreciate that!

app/assets/javascripts/code/services/_sensors.js
3v0k4: Great point

app/assets/javascripts/code/services/_sensors.js
3v0k4: I agree. But for now I'm leaving like this. I feel the issue should be solved at a different level. Hopefully will come back here soon

app/assets/javascripts/code/services/_sensors.js
3v0k4: For now I prefer to leave the duplication. Not sure how this will evolve and prefer the right concretion vs an abstraction

app/assets/javascripts/code/services/_sensors.js
3v0k4: Changed this to make it clearer

3v0k4:
Closing in favour of https://github.com/HabitatMap/AirCasting/pull/72

PR #70, opened 2018-12-19
Populate coordinates

AkkarinForest:
This PR populates the start_longitude and strat_latitude columns on stream. 
1. adds a rake task to migrate the data for existing streams
2. adds logic that will add this data for new streams


app/repositories/streams_repository.rb
```diff
@@ -17,4 +17,11 @@ def calc_average_value!(stream)
     stream.average_value = stream.measurements.average(:value)
     stream.save!
   end
+
+  def add_start_coordinates!(stream)
+      first_measurement = stream.measurements.order(time: :asc).first
```
3v0k4: This is only for mobile sessions? In fixed we have a bug where sometimes the measurements' times are wrong (1970 or 2010). I think that is the affected column. Maybe you could double check with Radek if it's best to order by time or id. Also, if this is slow, depending on the indexes ordering by time or id could make a difference

app/services/measurements_creator.rb
```diff
@@ -22,6 +22,7 @@ def call(stream, measurements_attributes)
     return if stream.session.type == 'FixedSession'
     streams_repository.calc_bounding_box!(stream)
     streams_repository.calc_average_value!(stream)
+    streams_repository.add_start_coordinates!(stream)
```
3v0k4: This is prolly not changing much but just so we are aware: this adds another roundtrip to database. Maybe it would be better to do all calculations first (bounding box, average, coordinates) and save at once That is whenever possible

lib/tasks/start_coordinates.rake
```diff
@@ -0,0 +1,14 @@
+desc "Populate start coordinates of stream with data from first measurement"
+task :populate_start_coordinates => [:environment] do
+  Stream.all.each do |stream|
+    first_measurement = stream.measurements.order(time: :asc).first
+    if first_measurement
+      puts stream.id
```
3v0k4: If `Stream.all` returns in a different order than id ascending I'm not sure if this yields valuable info

lib/tasks/start_coordinates.rake
```diff
@@ -0,0 +1,14 @@
+desc "Populate start coordinates of stream with data from first measurement"
+task :populate_start_coordinates => [:environment] do
+  Stream.all.each do |stream|
+    first_measurement = stream.measurements.order(time: :asc).first
+    if first_measurement
```
3v0k4: The logic here is really similar to the one in the repo, maybe we can remove the duplication?

lib/tasks/start_coordinates.rake
AkkarinForest: It's fine, I just wanted to see the progress, but instead of dots I used id, cause I also wanted to see if there are any streams without measurements.

app/repositories/streams_repository.rb
AkkarinForest: good catch! I'll check that.

AkkarinForest: Unfortunately, using id's is not a good idea. We don't sort the measurements before saving, so the order might be random. Also if we would sort them it would be by time anyway :/ But we have an index on time, so at least this shouldn't influence the performance much.
So I'm going to use time but this is only executed for Mobile, so I guess it's ok

3v0k4: We should try to make it clear what's the path for Mobile and what's the path for Fixed. Right now reading that code is a struggle. (The code that was there already I mean)

PR #73, opened 2019-01-07
Click mobile points

AkkarinForest:
Adds the same function that fixed sessions have.
1. Users can click the session marker to select the session.
2. While hovering over the marker session name is shown.

### DIFF
--- a/app/assets/javascripts/code/services/google/_map.js
+++ b/app/assets/javascripts/code/services/google/_map.js
@@ -196,7 +196,7 @@ export const map = (
       rectangles.clear();
     },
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/code/services/google/_map.js
```diff
@@ -196,7 +196,7 @@ export const map = (
       rectangles.clear();
     },
 
-    defaultMarkerOptions: function() {
+    defaultMarkerOptions: function(session) {
       return { title: session.title, zIndex: 0 };
```
3v0k4: I guess this was throwing an error since `session.title` wasn't there?

PR #74, opened 2019-01-08
fix(heat level): show heat level marker on fixed tab

AkkarinForest:
Fixes the bug: heat level markers where not showing for currently streaming sessions on Fixed tab.

### DIFF
--- a/app/assets/javascripts/code/services/_draw_session.js
+++ b/app/assets/javascripts/code/services/_draw_session.js
@@ -23,7 +23,7 @@ export const drawSession = (
       var points = [];
       _(this.measurements(session)).each(function(measurement, idx){
         var value = Math.round(measurement.value);
-        var level = heat.getLevel(value);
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/code/services/_draw_session.js
```diff
@@ -122,6 +122,4 @@ export const drawSession = (
   return new DrawSession();
 };
 
-function calculateHeatLevel(value) {
-  return heat.getLevel(value);
-};
+const calculateHeatLevel = (heat, value) => heat.getLevel(value);
```
3v0k4: Just for the sake of philosophy this could be
```
const calculateHeatLevel = (heat, value) => heat.getLevel(value);
const calculateHeatLevel = heat => value => heat.getLevel(value);
const calculateHeatLevel = heat => heat.getLevel;
```
but I wouldn't go that far here

PR #75, opened 2019-01-08
Fix use map type from url

3v0k4:
* Do not use cookie to store map info  (zoom, lat, lng, map type); since the same info is stored in url it's not clear why that should also be in a cookie.
* Restore map type from url

Decided not to put under test / refactor cause it would have been too much work and not justified in this case imo


AkkarinForest:
Looks good to me :)

PR #76, opened 2019-01-09
Do not reset page when session deselected

3v0k4:
Problem was that in SessionsListCtrl
```
$scope.$watch("params.get('map')",
```
whenever a session is not selected and the map moves, the code 
```
$scope.page = 0;
sessions.fetch($scope.page);
```
Doing that is fine when the user moved the map. In fact, in that case we want to fetch again all sessions from the first page.

Unfortunately, that is not correct when the map is moved programmatically. In fact, in that case we want the fetch to happen but the page to stay the same. We need the fetch to happen because on mobile we need to trigger the draw of the starting points. That happens is a callback pass to the fetch. Also, this does not create perf problems because the fetch is cached (`{ cache: true }`).

This PR fixes the problem by resetting the flag only if the user moves the map (and not on a programmatic change like `fitBounds`)
```
if (!hasChangedProgrammatically) $scope.page = 0;
```

### DIFF
--- a/app/assets/javascripts/code/services/google/_map.js
+++ b/app/assets/javascripts/code/services/google/_map.js
@@ -9,6 +9,7 @@ export const map = (
   googleMaps
 ) => {
   const TIMEOUT_DELAY = process.env.NODE_ENV === 'test' ? 0 : 1000;
+  let hasChangedProgrammatically = false;
 
   var Map = function() {};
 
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/code/services/google/_map.js
```diff
@@ -92,7 +95,8 @@ export const map = (
     },
 
     listen: function(name, callback, diffmap) {
-      return google.maps.event.addListener(diffmap || this.mapObj, name, _(callback).bind(this));
+      const cb = _(callback).bind(this);
```
AkkarinForest: Just curious, is `bind(this)` necessary here?
Also `cb` is not the most creating name :D

3v0k4:
merged

app/assets/javascripts/code/services/google/_map.js
3v0k4: I'm not totally sure it's needed, I just left it there w/o investigating. It's possible that w/o that the `this` would be wrong when called in response to the event.

PR #77, opened 2019-01-09
fix(measurement#since): remove upper limit

3v0k4:
Since `DateTime.current` was returning the current time in UTC and `time` in measurements is local time, it was possible that the former was before the latter.

### DIFF
--- a/spec/models/measurement_spec.rb
+++ b/spec/models/measurement_spec.rb
@@ -32,6 +32,20 @@ describe Measurement do
     let(:measurement2) { FactoryGirl.create(:measurement, :stream => stream2) }
 
     describe "#with_tags" do
+      it "returns all measurements belonging to stream id after since_date ordered by time" do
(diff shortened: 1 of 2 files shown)


spec/models/measurement_spec.rb
```diff
@@ -31,6 +31,20 @@
...
 
+    describe "#with_tags" do
+      it "returns all measurements belonging to stream id after since_date ordered by time" do
+        now = DateTime.current
+        stream = create_stream!(session: create_session!)
+        measurement1 = create_measurement!(stream: stream, time: now)
+        measurement2 = create_measurement!(stream: stream, time: now + 2.second)
+        measurement3 = create_measurement!(stream: stream, time: now + 1.second)
+
+        actual = Measurement.since(stream_id: stream.id, since_date: now)
```
czorek: I'm just curious about the variable name here - is it by convention?

PR #80, opened 2019-01-10
Fix error 'Cannot read property 'forEach' of undefined'

AkkarinForest:
The error appeared after some refactoring I did before. I'm switching it back to how it was before. But I'm not sure if this fixes the problem, or just makes it invisible :/

### DIFF
--- a/app/assets/javascripts/code/services/_draw_session.js
+++ b/app/assets/javascripts/code/services/_draw_session.js
@@ -104,7 +104,7 @@ export const drawSession = (
     },
 
     clear: function(sessions) {
-      sessions.forEach(this.undoDraw);
+      _(sessions).each(_(this.undoDraw).bind(this));
     },
 
     clearOtherSessions: function(sessions, selectedSession) {


app/assets/javascripts/code/services/_draw_session.js
```diff
@@ -104,7 +104,7 @@ export const drawSession = (
     },
 
     clear: function(sessions) {
-      sessions.forEach(this.undoDraw);
+      _(sessions).each(_(this.undoDraw).bind(this));
```
3v0k4: Wow this line is really noisy. Well it seems there's a null somewhere. Since 
```
_(null) === []
```
this way it does not explode. Prolly
```
(sessions || []).forEach(this.undoDraw);
```
is enough? Would be good to investigate why there's a null passed in

3v0k4: That's my assumption since the error is
```
angular.js:10147 TypeError: Cannot read property 'forEach' of undefined
    at o.clear (_draw_session.js:107)
    at o.redraw (_draw_session.js:96)
```
which I guess means that the following happened
```
undefined.forEach(this.undoDraw)
```

AkkarinForest:
I will merge this, cause there I already have three other open tickets and I don't want to have too many things at once. We can revisit this in the future and investigate.

PR #81, opened 2019-01-14
Case insensitive sensors

AkkarinForest:
It turned out that mysql database used in this project is NOT case sensitive, while other parts of the code were using hashes assuming that the keys are case sensitive. In order to unify this, we transform all sensor names to lower case at the entry point from backend to frontend. Or as close to it as possible. We still need to keep the original capitalization in the db as it is used for names that are displayed for the users.

A side effect of this change is that the sensor names in the URL are now all lower case and the old links will misbehave. We agreed that it's not an issue right now. We will revisit this if we receive feedback from users that it is needed.

### DIFF
--- a/app/assets/javascripts/code/services/_sensors.js
+++ b/app/assets/javascripts/code/services/_sensors.js
@@ -1,7 +1,7 @@
 import _ from 'underscore';
 
 const buildSensorId = sensor =>
-  sensor.measurement_type + "-" + sensor.sensor_name + " (" + sensor.unit_symbol + ")";
+  sensor.measurement_type + "-" + sensor.sensor_name.toLowerCase() + " (" + sensor.unit_symbol + ")";
 
 const ALL_SENSOR = { id: "all", select_label: "All", session_count: 0, measurement_type: "all" };
 const ALL_PARAMETER = { id: "all", label: "All" };
@@ -28,6 +28,9 @@ export const sensors = (params, $http) => {
       _(data).each(function(sensor){
         sensor.id =  buildSensorId(sensor);
         sensor.label = sensor.sensor_name + " (" + sensor.unit_symbol + ")";
+
+        sensor.sensor_name = sensor.sensor_name.toLowerCase();
+
         if (sensor.label.length >= 42) {
           sensor.select_label = sensor.label.slice(0, 40) + "…";
         } else {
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/code/services/_sensors.js
```diff
@@ -164,17 +167,16 @@ export const findAvailableSensorsForParameter = (sort, sensors, parameter) =>
...
+  const ORDERS = {};
+  ORDERS[buildSensorId({measurement_type: "Particulate Matter", sensor_name: "AirBeam2-PM2.5", unit_symbol: "µg/m³"})] = 4;
+  ORDERS[buildSensorId({measurement_type: "Particulate Matter", sensor_name: "AirBeam2-PM1", unit_symbol: "µg/m³"})] = 3;
+  ORDERS[buildSensorId({measurement_type: "Particulate Matter", sensor_name: "AirBeam2-PM10", unit_symbol: "µg/m³"})] = 2;
+  ORDERS[buildSensorId({measurement_type: "Particulate Matter", sensor_name: "AirBeam-PM", unit_symbol: "µg/m³"})] = 1;
+  ORDERS[buildSensorId({measurement_type: "Humidity", sensor_name: "AirBeam2-RH", unit_symbol: "%"})] = 2;
+  ORDERS[buildSensorId({measurement_type: "Humidity", sensor_name: "AirBeam-RH", unit_symbol: "%"})] = 1;
+  ORDERS[buildSensorId({measurement_type: "Temperature", sensor_name: "AirBeam2-F", unit_symbol: "F"})] = 2;
+  ORDERS[buildSensorId({measurement_type: "Temperature", sensor_name: "AirBeam-F", unit_symbol: "F"})] = 1;
+  ORDERS[buildSensorId({measurement_type: "Sound Level", sensor_name: "Phone Microphone", unit_symbol: "dB"})] = 1;
```
3v0k4: Have you considered
```
{
   [buildSensorId({measurement_type: "Particulate Matter", sensor_name: "AirBeam2-PM2.5", unit_symbol: "µg/m³"})] = 4,
  ...
}
```
that way you don't have the mutation from `{}` to the filled one

3v0k4: Great improvement even if it's a bit messy. Nice catch

PR #82, opened 2019-01-15
Indoor privacy

3v0k4:
* Hide usernames filter when `indoorOnly` is checked
* Format sessions for list; in particular show "anonymous" as a username for indoor sessions

screen shot 2019-01-15 at 12 07 17

### DIFF
--- a/app/assets/javascripts/tests/formatSessionForList.test.js
+++ b/app/assets/javascripts/tests/formatSessionForList.test.js
@@ -1,0 +1,63 @@
+import test from 'blue-tape';
+import { formatSessionForList } from '../code/values/sessions';
+
+test('removes markers', t => {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/tests/formatSessionForList.test.js
```diff
@@ -0,0 +1,63 @@
+import test from 'blue-tape';
```
AkkarinForest: Any reasons for calling the test file differently than the code file. I liked the structure we had so far, it's intuitive to find the tests there.

3v0k4: We can improve that. Reason is that this has 0 to do with angular. But I'm open to proposals

PR #84, opened 2019-01-16
Fix: Draw selected mobile session when All Parameter selected

AkkarinForest:


app/assets/javascripts/code/services/sessions_downloader.js
```diff
@@ -48,10 +48,18 @@ angular.module("aircasting").factory("sessionsDownloader", ['$rootScope', '$http
...
 
       session.$selected = sessionIds.include(session.id);
+
+      setDefaultSessionAttributes(session);
     });
     sessions.push.apply(sessions, data);
     sessions = orderBy(sessions, '$selected', 'end_time_local');
   };
 
+  const setDefaultSessionAttributes = function (session) {
```
3v0k4: Have you considered using `session => {}`. Also, that could be private (out of the `sessions_downloader` object)

PR #87, opened 2019-01-17
Don't draw a marker if last hour average outside of scope

AkkarinForest:


app/assets/javascripts/tests/_draw_session.test.js
```diff
@@ -31,8 +31,40 @@ test('drawMobileSessionStartPoint calls drawMarker', t => {
...
+  const sensors = { anySelected: () => true , tmpSelected: () => false }
+  const heat = { getLevel: () => 1 }
+  const drawSessionMock = _drawSession({ map, sensors, heat });
+
+  const session = { last_hour_average: 1 };
+  const actual = drawSessionMock.drawFixedSession(session);
+
+  const expected = [marker]
+
+  t.deepEqual(actual, expected  );
```
3v0k4: `  ` spaces

app/assets/javascripts/tests/_draw_session.test.js
```diff
@@ -31,8 +31,40 @@ test('drawMobileSessionStartPoint calls drawMarker', t => {
...
+  const map = { drawMarker: () => marker };
+  const sensors = { anySelected: () => true , tmpSelected: () => false }
+  const heat = { getLevel: () => 1 }
+  const drawSessionMock = _drawSession({ map, sensors, heat });
+
+  const session = { last_hour_average: 1 };
+  const actual = drawSessionMock.drawFixedSession(session);
+
+  const expected = [marker]
+
```
3v0k4: I guess you can rm this line

app/assets/javascripts/tests/_draw_session.test.js
```diff
@@ -31,8 +31,40 @@ test('drawMobileSessionStartPoint calls drawMarker', t => {
...
 
-const _drawSession = ({ map }) => {
+test('drawFixedSession creates a marker if heat level is not null', t => {
+  const marker = {};
+  const map = { drawMarker: () => marker };
+  const sensors = { anySelected: () => true , tmpSelected: () => false }
+  const heat = { getLevel: () => 1 }
+  const drawSessionMock = _drawSession({ map, sensors, heat });
+
+  const session = { last_hour_average: 1 };
```
3v0k4: This line should prolly be in the block above

app/assets/javascripts/tests/_draw_session.test.js
```diff
@@ -31,8 +31,40 @@ test('drawMobileSessionStartPoint calls drawMarker', t => {
   t.end();
 });
 
-const _drawSession = ({ map }) => {
+test('drawFixedSession creates a marker if heat level is not null', t => {
```
3v0k4: Seems like you could move some noise in the builder (`drawSessionMock`) cause it's a bit noisy here

app/assets/javascripts/code/services/_draw_session.js
```diff
@@ -50,6 +50,7 @@ export const drawSession = (
 
       if (sensors.anySelected() && !sensors.tmpSelected() && session.last_hour_average) {
         level = calculateHeatLevel(heat, session.last_hour_average);
+        if (!level) { return  session.markers };
```
3v0k4: You can remove `{}`. The nested if is a bit confusing maybe we can refactor around?

app/assets/javascripts/tests/_draw_session.test.js
```diff
@@ -31,8 +31,35 @@ test('drawMobileSessionStartPoint calls drawMarker', t => {
...
+  const actual = drawSessionMock.drawFixedSession(session);
+
+  const expected = [marker]
+  t.deepEqual(actual, expected);
+
+  t.end();
+})
+
+test('drawFixedSession does not create a marker if heat level is null', t => {
+  const drawSessionMock = _drawSession({});
```
3v0k4: Have you considered passing something like `const heat = { getLevel: () => null }`. Right now the test relies on the default which is hidden in `_drawSession`.

3v0k4: Also, you may want to call it stub. Mock has another meaning. Maybe just service would do (we have been calling it service in other tests).

Stub is a test double that is instructed to return specific values.
Mock is a test double that performs expectations on calls.

app/assets/javascripts/code/services/_heat.js
```diff
@@ -0,0 +1,102 @@
...
+      } else if (value < this.getValue("highest")) {
+        return 4;
+      } else {
+        return null;
+      }
+    },
+
+    outsideOfScope: function(value) {
+      return (value < this.getValue("lowest")) || (value > this.getValue("highest"))
+    }
```
AkkarinForest: that is the only new function I added, the rest is just copied from the old heat.js

PR #94, opened 2019-01-24
New markers for mobile sessions

AkkarinForest:


app/assets/javascripts/code/services/_heat.js
```diff
@@ -93,6 +93,21 @@ export const heat = ($rootScope, params, storage) => {
...
+      } else if (value < this.getValue("mid")) {
+        return "mid";
+      } else if (value < this.getValue("high")) {
+        return "high";
+      } else if (value < this.getValue("highest")) {
+        return "highest";
+      } else {
+        return null;
+      }
+    },
```
AkkarinForest: this returns the strings 'low', 'mid' ... cause css classes are called like this. But maybe it shouldn't be coupled like that. The classes could be named green, orange etc...

app/assets/javascripts/code/services/google/_map.js
```diff
@@ -203,3 +222,76 @@ export const map = (
...
+    this.position = position;
+
+    this.anchor = document.createElement('div');
+    this.anchor.classList.add('fixed-marker');
+    this.anchor.classList.add(colorClass);
+    this.anchor.innerText = content;
+    this.anchor.addEventListener('click', function() {
+      console.log('CLICKED')
+      $rootScope.$broadcast('markerSelected', {session_id: clicableObject});
+    });
```
AkkarinForest: Is there a way to add this event listener in a different way, different place? cause it requires $rootScope, and also it's coupled with session_id and we wanted this popup to be as general as possible.

app/assets/javascripts/code/values/session.js
```diff
@@ -9,3 +9,27 @@ export const formatSessionForList = session => ({
...
+  roundedAverage() {
+    return Math.round(this.session.average)
+  };
+
+  average() {
+    return this.session.average
+  };
+
+  selectedSensorUnit(selectedSensor) {
+    this.session.streams[selectedSensor]["unit_symbol"];
```
AkkarinForest: I'm not sure this belongs to session. maybe there should be a similar class for streams???? how would you do it?

app/assets/javascripts/code/values/session.js
```diff
@@ -9,3 +9,27 @@ export const formatSessionForList = session => ({
   timeframe: session.timeframe,
   shortTypes: session.shortTypes
 });
+
+export class Session {
```
3v0k4: Personally I'd avoid using `class`.  What I'd suggest is something like
```
export const roundedAverage = session => Math.round(average(session))

export const average = session => session.average

...
```
Then you can

```
import Session from '/path'

Session.average(session)
```

3v0k4: There's a lot of ceremony with `class`. Also it introduces `this`. And it's less composable than functions

app/assets/javascripts/code/values/session.js
3v0k4: Looks ok for now. When there will be too many things related to streams in here we could extract

3v0k4: Also, the quotes are unnecessary

app/assets/javascripts/code/values/session.js
```diff
@@ -9,3 +9,27 @@ export const formatSessionForList = session => ({
...
+  average() {
+    return this.session.average
+  };
+
+  selectedSensorUnit(selectedSensor) {
+    this.session.streams[selectedSensor]["unit_symbol"];
+  };
+
+  startingLatLng(selectedSensor) {
+    const latitude = this.session.streams[selectedSensor]["start_latitude"];
```
3v0k4: quotes are unnecessary

app/assets/javascripts/code/values/session.js
```diff
@@ -9,3 +9,27 @@ export const formatSessionForList = session => ({
...
+    const longitude = this.session.streams[selectedSensor]["start_longitude"];
+    return new google.maps.LatLng(latitude, longitude);
```
3v0k4: Prolly Session shouldn't know about google maps

app/assets/javascripts/code/services/google/_map.js
```diff
@@ -1,5 +1,7 @@
 import _ from 'underscore';
 
+var Popup;
```
3v0k4: I think we can get rid of this, create the class in another module, import it and use it normally

app/assets/javascripts/code/services/google/_map.js
```diff
@@ -167,6 +169,23 @@ export const map = (
       return newMarker;
     },
 
+    drawCustomMarker: function({
+        latLng: latLng,
+        content: content,
+        colorClass: colorClass,
+        clicableObject: clicableObject
+      }) {
```
3v0k4: You don't need the things after `:`. This is ok
```
drawCustomMarker: function({ latLng, content, colorClass, clicableObject })
```
You can use the name after `:` in case you want to change the name of the const.

app/assets/javascripts/code/services/google/_map.js
```diff
@@ -167,6 +169,23 @@ export const map = (
...
+
+      definePopupClass();
```
3v0k4: Can we remove this call and just import the class?

app/assets/javascripts/code/services/google/_map.js
```diff
@@ -203,3 +222,76 @@ export const map = (
...
+   */
+  Popup = function(position, content, colorClass, $rootScope, clicableObject) {
+    this.position = position;
+
+    this.anchor = document.createElement('div');
+    this.anchor.classList.add('fixed-marker');
+    this.anchor.classList.add(colorClass);
+    this.anchor.innerText = content;
+    this.anchor.addEventListener('click', function() {
+      console.log('CLICKED')
```
3v0k4: We should prolly rm this

app/assets/javascripts/code/services/google/_map.js
```diff
@@ -203,3 +222,76 @@ export const map = (
 
   return new Map();
 };
+
+
+function definePopupClass() {
+  /**
```
3v0k4: I guess we can rm comments

app/assets/javascripts/code/services/google/_map.js
```diff
@@ -203,3 +222,76 @@ export const map = (
...
+   * @extends {google.maps.OverlayView}
+   */
+  Popup = function(position, content, colorClass, $rootScope, clicableObject) {
+    this.position = position;
+
+    this.anchor = document.createElement('div');
+    this.anchor.classList.add('fixed-marker');
+    this.anchor.classList.add(colorClass);
+    this.anchor.innerText = content;
+    this.anchor.addEventListener('click', function() {
```
3v0k4: Maybe we can inject the callback instead of the single pieces? (ie `rootScope`, `clicableObj`)

app/assets/javascripts/code/services/google/_map.js
```diff
@@ -203,3 +222,76 @@ export const map = (
 
   return new Map();
 };
+
+
+function definePopupClass() {
```
3v0k4: There's a lot of copy / pasted code in here. Do we need all of that?

3v0k4:
Lots of cool stuff going on 👍

app/assets/javascripts/code/values/session.js
AkkarinForest: I get an error when I remove them. Am I missing something? 
```ReferenceError: unit_symbol is not defined
at Session.selectedSensorUnit (session.js?d68e:27)```

app/assets/javascripts/code/values/session.js
AkkarinForest: I tried something like this, but I was getting this error:
`TypeError: Cannot read property 'roundedAverage' of undefined`
 `at MobileSessions.drawSessionInLocation `
In the end I was able to make it work with class

PR #96, opened 2019-01-28
Add heat legend to mobile tab

AkkarinForest:
Now that we have mobile markers it makes sense to have heat legend again.


AkkarinForest:
It was not reverted, I couldn't find that commit that removed the heat legend. But we had the same logic for fixed so it's copied from there.

PR #98, opened 2019-01-31
New fixed markers WIP

AkkarinForest:


app/assets/javascripts/code/services/_fixed_sessions.js
```diff
@@ -116,8 +117,54 @@ export const fixedSessions = (
     },
 
     drawSessionsInLocation: function() {
-      map.markers = [];
-      (this.get()).forEach(session => drawSession.drawFixedSession(session));
+      if (sensors.anySelected() && this.scope.params.get('data').location.streaming) {
```
3v0k4: Does `params.get('data').location.streaming` work w/o `this.scope`?

app/assets/javascripts/code/services/_fixed_sessions.js
```diff
@@ -116,8 +117,54 @@ export const fixedSessions = (
...
+      } else {
+        (this.get()).forEach(session => this.drawDefaultMarkers(session));
+      }
+    },
+
+    drawColorCodedMarkers: function(session, selectedSensor) {
+      drawSession.undoDraw(session);
+      session.markers = [];
+
+      const content = Session.lastHourAverageVauleAndUnit(session, selectedSensor);
```
3v0k4: Typo in Value ;)

app/assets/javascripts/code/services/_fixed_sessions.js
```diff
@@ -116,8 +117,54 @@ export const fixedSessions = (
...
+      const heatLevel = heat.levelName(Session.lastHourAverage(session));
+      const latLng = Session.latLng(session);
+      const callback = (id) => () => $rootScope.$broadcast('markerSelected', {session_id: id});
+
+      if (!heat.outsideOfScope(heatLevel)) {
```
3v0k4: I'd suggest an early return
```
if (heat.outsideOfScope(heatLevel)) return;
...
```

app/assets/javascripts/code/services/_fixed_sessions.js
```diff
@@ -116,8 +117,54 @@ export const fixedSessions = (
...
+        const customMarker = map.drawCustomMarker({
+            latLng: latLng,
+            content: content,
+            colorClass: heatLevel,
+            callback: callback(Session.id(session))
+          });
+        session.markers.push(customMarker);
+        map.markers.push(customMarker);
+        session.drawed = true;
```
3v0k4: We could move all of the `session.` to `Session.`. What's left now prolly relies on the mutation so we cannot really make the update like in fp.

For example we could have
```
// Session.js
const addMarker = (session, marker) => ({ ...session, markers: [...session.markers, marker] })
```
that is we return a new session with the added marker and no mutations.

Unfortunately, we cannot do that most of the times. In fact, if there's a watch keeping an eye on the session, with no mutations it wouldn't trigger. That's a limit we have in Angular

app/assets/javascripts/code/services/_fixed_sessions.js
```diff
@@ -116,8 +117,54 @@ export const fixedSessions = (
...
+      }
+    },
+
+    drawDefaultMarkers: function(session) {
```
3v0k4: Have you considered extracting the common code of `drawColorCodedMarkers` and `drawDefaultMarker` to another fnc?

app/assets/javascripts/code/services/_mobile_sessions.js
```diff
@@ -134,13 +133,14 @@ export const mobileSessions = (
       const callback = (id) => () => $rootScope.$broadcast('markerSelected', {session_id: id});
 
       if (!heat.outsideOfScope(heatLevel)) {
-        const popup = map.drawCustomMarker({
+        const customMarker = map.drawCustomMarker({
```
3v0k4: Have you considered just `marker` as a name?

app/assets/javascripts/code/services/google/_map.js
```diff
@@ -185,10 +185,8 @@ export const map = (
     },
 
     removeAllMarkers: function() {
-      var markers = this.markers;
-      _(markers).each(function(marker) {
-        marker.setMap(null);
-      });
+      this.markers.forEach(marker => marker.setMap(null));
```
3v0k4: I guess `marker.setMap(null)` would still work?

3v0k4: Is there a chance `this.markers` is `null` or something like that? Cause `_()` handles that case and now would explode.

app/assets/javascripts/tests/_map.test.js
```diff
@@ -213,6 +213,21 @@ test('when map was change programatically saveViewport calls params update with
...
 });
 
+test('removeAllMarkers removes all markers', t => {
+  let markers = [{setMap: (_) => {}}]
+  const service = _map({})
+  service.markers = markers
+  
+  service.removeAllMarkers()
+
+  const expected = [];
```
3v0k4: Not sure you want to do the same. I normally keep the expected attached to the expected block

```
 const expected = []
 t.deepEqual(actual, expected)
```

app/assets/javascripts/tests/helpers.js
```diff
@@ -10,6 +10,11 @@ export const mock = name => {
     wasCalled: () => calls1.length === 1,
     wasCalledWith: arg => deepEqual(arg, calls1[calls1.length - 1]),
     wasCalledWith2: arg => deepEqual(arg, calls2[calls2.length - 1]),
-    wasCalledWith3: arg => deepEqual(arg, calls3[calls3.length - 1])
+    wasCalledWith3: arg => deepEqual(arg, calls3[calls3.length - 1]),
+    wasCalledWithParameter: obj => {
```
3v0k4: I'd suggest naming it like in rspec (https://relishapp.com/rspec/rspec-mocks/v/3-2/docs/setting-constraints/matching-arguments). Something like `wasCalledWithObjIncluding`. Because parameter seems like it's checking the whole parameter

app/assets/javascripts/code/services/_fixed_sessions.js
AkkarinForest: Good point. I'll do that but not just now. They look similar, but they refer to different concepts and I think they will diverge further with the next tickets, so I want to hold off on extracting this until I will know how the default markers logic will change. In the spirit of duplication is better than wrong abstraction :D

PR #100, opened 2019-02-06
Sidekiq two queues

3v0k4:
* refactor and added tests coverage (previously there were no tests)
* use 2 queues instead of 1 in Sidekiq: streams with 20_000 measurements or more are pushed in the `slow` queue, streams with less are pushed in the default queue

I've left the lines adding info in the logs. I guess we will need them for some more time. At least until we really fix the problem with the slowness

### DIFF
--- a/app/models/stream.rb
+++ b/app/models/stream.rb
@@ -106,7 +106,7 @@ class Stream < ActiveRecord::Base
   def self.build!(data = {})
     measurements = data.delete(:measurements)
     stream = create!(data)
-    MeasurementsCreator.call(stream, measurements)
+    MeasurementsCreator.new.call(stream, measurements)
     stream
   end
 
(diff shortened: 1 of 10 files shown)


app/models/stream.rb
```diff
@@ -119,7 +119,7 @@ def self.build_or_update!(data = {})
     stream.set_bounding_box(latitude, longitude) unless stream.has_bounds?
     stream.save!
 
-    MeasurementsCreator.call(stream, measurements_attributes)
+    MeasurementsCreator.new.call(stream, measurements_attributes)
```
AkkarinForest: this and line 109 seem very similar, any reason why it's called measurements_attributes here and just measurements on 109 - it's a bit confusing, I was trying to figure out what's the difference, cause the name doesn't really say much

app/services/async_measurements_creator.rb
```diff
@@ -0,0 +1,23 @@
+class AsyncMeasurementsCreator
+  SLICE_SIZE = 500
+  AMOUNT_THRESHOLD = 20_000
```
AkkarinForest: I guess just `threshold` is enough? And also amount is uncountable, so it's better to use `number_threshold` if needed. Check out this for reference: https://goo.gl/images/HQGoZe 😄

3v0k4:
Thanks for the CR @AkkarinForest !

PR #104, opened 2019-02-13
Upgrade to rails 4.2.x

3v0k4:


app/models/measurement.rb
```diff
@@ -24,7 +24,6 @@ class Measurement < ActiveRecord::Base
   Y_SIZES = (1..300).map { |i| 1.2 ** i * 0.000001 }
   FIXED_MEASUREMENTS_IN_A_DAY = 1440
 
-  # belongs_to :session, :through => :stream, :inverse_of => :measurements, :counter_cache => true
```
AkkarinForest: We don't need this association anymore? Can it influence performance?

app/models/user.rb
```diff
@@ -111,4 +95,13 @@ def admin?
...
     self.username.chomp!
   end
+
+  # TokenAuthenticatable was removed from Devise in 3.1
+  # https://gist.github.com/josevalim/fb706b1e933ef01e4fb6
+  def generate_authentication_token
+    loop do
+      token = Devise.friendly_token
+      break token unless User.where(authentication_token: token).first
+    end
```
AkkarinForest: Are we sure that this always breaks? I mean is there a risk of infinite loop?

Gemfile
```diff
@@ -1,39 +1,39 @@
...
+gem 'sass-rails'
+gem 'coffee-rails'
 
-# Gems used only for assets and not required
-# in production environments by default.
-group :assets, :development do
-  gem 'coffee-rails', '~> 3.2.1'
-  gem 'sass-rails', '~> 3.2.2'
-  gem 'uglifier', '~> 1.2.7'
+group :development do
```
AkkarinForest: Any reason we need two development groups? there is another one on line 40

db/migrate/20190212125754_change_collation_for_tag_names.acts_as_taggable_on_engine.rb
```diff
@@ -0,0 +1,10 @@
+# This migration comes from acts_as_taggable_on_engine (originally 5)
+# This migration is added to circumvent issue #623 and have special characters
+# work properly
+class ChangeCollationForTagNames < ActiveRecord::Migration
+  def up
+    if ActsAsTaggableOn::Utils.using_mysql?
+      execute("ALTER TABLE tags MODIFY name varchar(255) CHARACTER SET utf8 COLLATE utf8_bin;")
+    end
+  end
+end
```
AkkarinForest: Should this migration also have a `down` method?

lib/session_builder.rb
```diff
@@ -24,10 +24,10 @@ def build_session!
     data = build_local_start_and_end_time(data)
 
     begin
-      session = Session.create!(data)
+      session = Session.create!(data.reject { |k| [:drawable, :deleted].include?(k)})
```
AkkarinForest: What is a :drawable session? I'm just curious, cause I've never seen that.

spec/models/measurement_spec.rb
```diff
@@ -16,14 +16,9 @@
...
 # You can contact the authors by email at <info@habitatmap.org>
 
-require 'spec_helper'
+require 'rails_helper'
 
 describe Measurement do
-  it { is_expected.to validate_presence_of :value }
-  it { is_expected.to validate_presence_of :longitude }
-  it { is_expected.to validate_presence_of :latitude }
-  it { is_expected.to validate_presence_of :time }
```
AkkarinForest: Have you consider rewriting them to something like 
```
 it "is invalid without a value" do
   m = Measurement.new(value: nil)
   m.valid?
   expect(m.errors[:value]).to include("can't be blank")
 end
```
Or maybe we don't need to test validations in newer rails versions?

app/controllers/api/measurement_sessions_controller.rb
```diff
@@ -20,6 +20,9 @@ module Api
   class MeasurementSessionsController < BaseController
     INT_Q_ATTRS = [:time_from, :time_to, :day_from, :day_to]
 
+    # TokenAuthenticatable was removed from Devise in 3.1
+    # https://gist.github.com/josevalim/fb706b1e933ef01e4fb6
+    before_filter :authenticate_user_from_token!, :only => :create
     before_filter :authenticate_user!, :only => :create
```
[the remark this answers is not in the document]
3v0k4: Devise removed TokenAuthenticable which was in charge of doing what I've added in this PR. Each user when created has a token (string) which is sent with each request to authenticate itself.

`authenticate_user_from_token!` tries to authenticate the user looking for the token. `authenticate_user!` tries to authenticate in a different way. That's more or less how I understand it

app/models/measurement.rb
3v0k4: It was already commented out. Just removed the comment ;)

app/models/user.rb
3v0k4: Copied from the Devise wiki. I trust that

Gemfile
3v0k4: Good catch! Fixed it

db/migrate/20190212125754_change_collation_for_tag_names.acts_as_taggable_on_engine.rb
3v0k4: Not sure really. It's generated by the gem

lib/session_builder.rb
3v0k4: With Rails 4 trying to create a model w/ an attribute that doesn't exist on the model fails. 
```
User.create(pizza: "pants") // BOOM
```
Those two are coming from the mobile app.

This should be handled in a better way (selecting what's right and not removing what's wrong). But hopefully somebody else will stumble on this and improve

spec/models/measurement_spec.rb
3v0k4: I think these tests are meaningless. The fact that `validates :....` works that way is tested by rails. I believe declarative code of a framework should not be tested.

It's like testing that 1 + 1 == 1. I trust ruby to give me a working `+`

PR #105, opened 2019-02-14
Cluster fixed sessions

AkkarinForest:
Crowd map is used for sessions that are close to each other. That only applies to the situation when the user selected currently streaming. Also, it doesn't work for all/all and indoor sessions. 
The color of a cluster is based on the last hour average. However, the info in a pop-up window when the user clicks on a cluster work the same as on mobile, meaning is based on the time filter. But there is another ticket regarding this popup, so I'll adjust it there.


lib/fixed_average_info.rb
```diff
@@ -0,0 +1,61 @@
...
+  def calculate_average(sessions)
+    sessions
+    .map { |session| last_hour_average_for(session, data[:sensor_name]) }
+    .sum / sessions.size
+  end
+
+  def last_hour_average_for(session, sensor_name)
+    stream = Stream.where(sensor_name: sensor_name, session_id: session.id).first
+
+    last_measurement_time = stream.measurements.last.time
```
AkkarinForest: turns out we have a default scope for measurements that sorts them by time: :asc, so I don't need to worry about this here. 
On the other hand, maybe we should remove this default scope, cause it probably slows down all the queries and it's not needed in some of them...

lib/fixed_average_info.rb
```diff
@@ -0,0 +1,61 @@
...
+    @grid_x ||= if data[:west] < data[:east]
+       (data[:east] - data[:west]) / data[:grid_size_x]
+    else
+      @grid_x = (180 - data[:west] + 180 + data[:east]) / data[:grid_size_x]
+    end
+  end
+
+  def grid_y
+    @grid_y ||= Y_SIZES.find { |x| x > (data[:north] - data[:south]) / data[:grid_size_y] }
+  end
```
AkkarinForest: This last two I just copied from the `lib/InfoAverage.rb` so don't ask me how they work :D And seriously I'm considering extracting the common logic, but didn't want to do to many radical changes at once, especially with the code I only half understand.

lib/fixed_average_info.rb
```diff
@@ -0,0 +1,61 @@
+class FixedAverageInfo
```
AkkarinForest: It's kinda based on `lib/AverageInfo.rb`, the surprising part is that this other class filters for tags, usernames and so on. I didn't add these here cause I think it's obsolete. Since we are getting sessions ids, and these ids come from the frontend. And the frontend fetched based on these filters already. 
Unless there is something I'm no aware of?

lib/fixed_average_info.rb
AkkarinForest: I'm reaserching that and I think we should at least have an index on Measurement.time since we have this `42   default_scope { order("time ASC") }`in `models/Measurement.rb`. More info: https://www.percona.com/blog/2006/09/01/mysql-order-by-limit-performance-optimization/

AkkarinForest: cc @3v0k4

PR #107, opened 2019-02-15
New popup for crow map for fixed sessions

AkkarinForest:


app/assets/javascripts/code/services/_fixed_sessions.js
```diff
@@ -144,7 +144,7 @@ export const fixedSessions = (
       const clusteredSessions = clusterer(sessionsToCluster, map);
       const lonelySessions = sessions.filter(isNotIn(clusteredSessions));
 
-      sessionsUtils.updateCrowdMapLayer(sessionsIds(clusteredSessions));
+      sessionsUtils.updateCrowdMapLayer(sessionsIds(clusteredSessions), "Fixed");
```
3v0k4: You could create a Constants.js file and export the constant from there. That way you are sure it has only one single point of truth.

app/assets/javascripts/code/services/google/info_window.js
```diff
@@ -8,22 +8,26 @@ angular.module("google").factory("infoWindow", ["map", "$http", "$compile", "$ro
...
-    show: function(url, data, position){
-      var self = this;
+
+    show: function(url, data, position, type){
       this.popup.setContent("working..");
       this.popup.setPosition(position);
       this.popup.open(map.get());
-      $timeout(function(){
-        $http.get(url, {params : data, cache: true}).success(_(self.onShowData).bind(self));
+      const htmlFile = (type === "Fixed") ? '/partials/fixed_info_window.html' : '/partials/info_window.html';
```
3v0k4: You could extract the file paths on line 4 into constants (eg `const FIXED_PATH`)

app/assets/javascripts/code/services/google/info_window.js
```diff
@@ -8,22 +8,26 @@ angular.module("google").factory("infoWindow", ["map", "$http", "$compile", "$ro
     get: function() {
       return this.popup;
     },
-    show: function(url, data, position){
-      var self = this;
+
+    show: function(url, data, position, type){
```
3v0k4: Have you considered something more expressive for `type`? Example `sessionType`

app/assets/javascripts/code/services/google/info_window.js
3v0k4: Have you considered a different name for `htmlFile`? I don't like it much cause it's not really a file. Maybe `partialPath` or `htmlPath`?

app/controllers/api/regions_controller.rb
```diff
@@ -45,7 +45,11 @@ def show
 
       data[:session_ids] ||= []
 
-      respond_with RegionInfo.new(data)
+      if (data[:session_ids] != [] && Session.find(data[:session_ids].first).type == "FixedSession")
```
3v0k4: Have you considered adding `fixed?` to Session, MobileSession and FixedSession?
```
# Session
def fixed?
  raise NotImplementedError, "subclass did not define #fixed?"
end

# FixedSession
def fixed?
  true
end

# MobileSession
def fixed?
  false
end
```

3v0k4: You prolly can skip the `data[:session_ids] != []` part?

spec/services/fixed_region_info_spec.rb
```diff
@@ -0,0 +1,90 @@
+require "spec_helper"
+
+describe FixedRegionInfo do
+  it "calculates average, number of contributors, top contributors and number of samples for sessions for given stearm" do
```
3v0k4: typo in stream

spec/services/fixed_region_info_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    session1 = create_session_with_stearms_and_measurements!({ id: 1, longitude: 1.0, latitude: 1.0, value: 1 })
```
3v0k4: typo in streams

3v0k4: Have you considered having defaults in for `longitude` and `latitude`?

spec/services/fixed_region_info_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      session_ids: [session1.id, session2.id],
+      sensor_name: "AirBeam2-F",
+    }
+
+    region_info = FixedRegionInfo.new.call(data)
+
+    expect(region_info).to eq({
+        average: 1.5,
+        number_of_contributors: 2,
+        top_contributors: (["Test User1", "Test User2"]),
```
3v0k4: It's difficult to understand why the users are called that way. You could make that explicit by creating the users in the test and having `username1` and `username2` in common

spec/services/fixed_region_info_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+        number_of_samples: 120,
```
3v0k4: The fact that this should be 120 is hidden down below. Maybe you could make it more explicit in the test?

spec/services/fixed_region_info_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+private
+
+def create_session_with_stearms_and_measurements!(attributes)
+  session = create_session!({
+    id: attributes.fetch(:id),
+    longitude: attributes.fetch(:longitude),
+    latitude: attributes.fetch(:latitude),
+  })
+  stream = create_stream!({ session: session })
+  create_measurements!({ stream: stream, value: attributes.fetch(:value) })
```
3v0k4: In ruby when you are passing a hash as an argument you can skip the `{}` ;)

spec/services/fixed_region_info_spec.rb
3v0k4: I'd suggest to try to refactor this test a bit. It covers the production code but does not really explain how the input is related to the output. You can think of tests as documentation and ask yourself, does the code and description of this test explain why the output is
```
average: 1.5,
number_of_contributors: 2,
top_contributors: (["Test User1", "Test User2"]),
number_of_instruments: 2,
number_of_samples: 120,
```
?

spec/services/fixed_region_info_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  session
+end
+
+def create_session!(attributes)
```
3v0k4: We are prolly doing this kind of things in many tests. We should start extracting them

PR #108, opened 2019-02-19
Improve performance for fixed info windows

AkkarinForest:
More descriptions in commits body if needed :)

### DIFF
--- a/app/services/fixed_region_info.rb
+++ b/app/services/fixed_region_info.rb
@@ -21,7 +21,7 @@ class FixedRegionInfo
   end
 
   def last_measurement_time(streams_ids)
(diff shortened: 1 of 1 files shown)


app/services/fixed_region_info.rb
```diff
@@ -21,7 +21,7 @@ def streams_ids(sessions, sensor_name)
   end
 
   def last_measurement_time(streams_ids)
-    Measurement.with_streams(streams_ids).last.time
+    Measurement.with_streams(streams_ids).select("MAX(time) as last_time").first.last_time
```
3v0k4: Cooool stuff. And great gains! Good job!

PR #109, opened 2019-02-19
Don't center markers based on pixels

AkkarinForest:
That was causing more problems than it was adding value. Now the top left corner of a marker is placed in the given location, but in my opinion, it's much different than if the center of the central dot was placed in the location.


3v0k4:
As long as Michael knows and stuff works out fine I guess this is a really good simplification!

Maybe you could add a comment somewhere in the code that explains how the markers are positioned?

AkkarinForest:
In the end, we found a better solution to position the markers correctly, by using a parent div.

PR #110, opened 2019-02-19
When average is on a threshold it should have a colour of a lower range.

AkkarinForest:


3v0k4:
The fact that the logic changed but no tests did is smelly ;)

AkkarinForest:
Sorry! I added more precise tests.

PR #124, opened 2019-03-04
New filters

AkkarinForest:
I added a second elm app and used it to create profile and tags filter for fixed sessions. We now have some duplication between them, I extracted part of it to Labels module, but if you have any suggestions on how to clean it further let me know. Or maybe it's ok to keep the duplication since it's temporary until we connect the elm apps together.
Also, testing gets more tricky with all these abstraction and duplications, so I'm not sure if now I cover all the stuff that should be covered, or if I'm testing some parts multiple times...

Also, if you don't want to type too much we can always have a skype session or you can record me a voice message :D


app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
```diff
@@ -1,5 +1,5 @@
 import _ from 'underscore';
-import { Elm } from '../../../elm/src/Main.elm';
+import { Elm } from '../../../elm/src/MobileSessionsFilters.elm';
```
3v0k4: Great move renaming this too!

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -94,4 +96,61 @@ export const FixedSessionsMapCtrl = (
     }, true);
 
   $scope.setDefaults();
+
+  if (process.env.NODE_ENV !== 'test') {
```
3v0k4: We could consider extracting the same parts into an abstraction but I'm not sure it wouldn't be wasted effort. Especially cause there's already the mobile / fixed duplication. Let's just not forget. If you want on my wall there's an "eject" section where I was collecting things to be done before wrapping up

app/javascript/angular/code/directives/autocomplete.js
```diff
@@ -1,55 +0,0 @@
-angular.module("aircasting").directive('autocomplete', function (){
```
3v0k4: Great! Less Angular stuff!

app/javascript/elm/tests/MobileSessionsFiltersTests.elm
```diff
@@ -133,53 +101,21 @@ tagsArea =
...
                     defaultModel
                         |> update (AddProfile activityValue)
                         |> Tuple.first
                         |> view
                         |> Query.fromHtml
                         |> Query.has [ Slc.text activityValue ]
-            , fuzz string "new profile label has a button" <|
-                \profileName ->
-                    { defaultModel | profiles = Labels.fromList [ profileName ] }
+            , test "new profile label has a button" <|
```
3v0k4: Was the `fuzz` a problem on this one?

public/partials/usernames_and_tags.html
```diff
@@ -1,30 +0,0 @@
-<div ng-controller="UsernamesAndTagsFiltersCtrl">
```
3v0k4: Awesome

app/javascript/elm/tests/MobileSessionsFiltersTests.elm
AkkarinForest: Yes, cause it sometimes produces an empty string. I think at some point I will create a custom `fuzz` that produces strings that I need and I'll revisit these tests.

PR #127, opened 2019-03-06
Refactor labels

3v0k4:
Take a look. If you like it feel free to merge (there are prolly conflicts) @AkkarinForest 

This is a complex abstraction https://www.youtube.com/watch?v=RN2_NchjrJQ (maybe justified?). Let's try to keep abstractions as simple as possible in general


Gemfile
```diff
@@ -32,6 +32,7 @@ gem 'sass-rails'
 gem 'coffee-rails'
 gem 'activesupport-json_encoder'
 gem 'webpacker', git: 'https://github.com/rails/webpacker.git', branch: '3-x-stable'
+gem 'rb-readline'
```
3v0k4: On my personal Mac I couldn't setup the project w/o. We can remove this for now I guess

app/javascript/elm/src/FixedSessionFilters.elm
```diff
@@ -44,44 +44,33 @@ init flags =
...
+    | ProfileLabels LabelsInput.Msg
 
 
 update : Msg -> Model -> ( Model, Cmd Msg )
 update msg model =
     case msg of
-        UpdateTagsSearch content ->
-            ( { model | tags = Labels.updateCandidate model.tags content }, Cmd.none )
+        TagsLabels subMsg ->
+            updateLabels subMsg model.tags Ports.updateTags TagsLabels (\tags -> { model | tags = tags })
```
3v0k4: This should make easier to spot a problem where we use a "profile" thing instead of a "tag" thing (eg `ProfileLabels` instead of `TagLabels`). In other words, the differences are here and the noise is in `updateLabels`. We could one day refactor in a way that the types make tags and profiles different

AkkarinForest:
This is pretty cool! And you're right there will be conflicts, but that's good - means I really need to understand this if I want to merge :D 
Thanks for doing this refactor! Even if we don't really need that in the project it's awesome to see and learn these patterns :)

Sorry for late response, I only had time to look at this today.

PR #128, opened 2019-03-07
Prepare TimeRange data structure

AkkarinForest:
There is a lot more to come on the js side, but I tried to divide that feature into smaller PRs :)

### DIFF
--- a/app/javascript/elm/src/FixedSessionFilters.elm
+++ b/app/javascript/elm/src/FixedSessionFilters.elm
@@ -1,9 +1,11 @@
-module FixedSessionFilters exposing (init)
+module FixedSessionFilters exposing (Msg(..), defaultModel, update, view)
 
 import Browser
 import Html exposing (Html, div)
+import Json.Encode as Encode
 import Labels exposing (Labels)
 import Ports
+import TimeRange exposing (TimeRange)
 
 
 
@@ -13,6 +15,7 @@ import Ports
 type alias Model =
     { tags : Labels
     , profiles : Labels
+    , timeRange : TimeRange
     }
 
 
@@ -20,12 +23,14 @@ defaultModel : Model
 defaultModel =
     { tags = Labels.empty
     , profiles = Labels.empty
+    , timeRange = TimeRange.defaultTimeRange
     }
 
 
 type alias Flags =
     { tags : List String
     , profiles : List String
+    , timeRange : Encode.Value
     }
 
 
@@ -34,6 +39,7 @@ init flags =
     ( { defaultModel
         | tags = Labels.fromList flags.tags
         , profiles = Labels.fromList flags.profiles
+        , timeRange = TimeRange.update defaultModel.timeRange flags.timeRange
       }
     , Cmd.none
     )
@@ -50,6 +56,7 @@ type Msg
     | UpdateProfileSearch String
     | AddProfile String
     | RemoveProfile String
+    | UpdateTimeRange Encode.Value
 
 
 update : Msg -> Model -> ( Model, Cmd Msg )
@@ -73,6 +80,13 @@ update msg model =
         RemoveProfile profile ->
             Labels.removeLabel profile model.profiles (updateProfiles model) Ports.updateProfiles
 
+        UpdateTimeRange value ->
+            let
+                newTimeRange =
+                    TimeRange.update model.timeRange value
+            in
+            ( { model | timeRange = newTimeRange }, Ports.updateTimeRange value )
+
 
 updateProfiles : { a | profiles : Labels } -> Labels -> { a | profiles : Labels }
 updateProfiles labelled labels =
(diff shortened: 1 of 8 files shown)


app/javascript/elm/src/TimeRange.elm
```diff
@@ -0,0 +1,56 @@
+module TimeRange exposing (TimeRange(..), defaultTimeRange, update, viewTimeFilter)
```
AkkarinForest: I have to expose `TimeRange(..)` to be able to create expected values in the `TimeRangeTests`. Or is there a better way to do this?

app/javascript/elm/src/MobileSessionsFilters.elm
```diff
@@ -27,6 +30,7 @@ defaultModel =
     , isCrowdMapOn = False
     , tags = Labels.empty
     , profiles = Labels.empty
+    , timeRange = TimeRange.defaultTimeRange
     }
```
AkkarinForest: Is the default model useful here? Cause it's supposed to be overwritten by flags anyway.

app/javascript/elm/src/TimeRange.elm
```diff
@@ -0,0 +1,56 @@
...
+    , timeTo : Int
+    }
+
+
+defaultTimeRange : TimeRange
+defaultTimeRange =
+    TimeRange
+        { timeFrom = 1
+        , timeTo = 1
+        }
```
AkkarinForest: This is actually never used cause it's overwritten by flags, I'm thinking of deleting the default model totally.

app/javascript/elm/src/MobileSessionsFilters.elm
```diff
@@ -92,6 +99,13 @@ update msg model =
         RemoveProfile profile ->
             Labels.removeLabel profile model.profiles (updateProfiles model) Ports.updateProfiles
 
+        UpdateTimeRange value ->
+            let
+                newTimeRange =
+                    TimeRange.update model.timeRange value
+            in
+            ( { model | timeRange = newTimeRange }, Ports.updateTimeRange value )
```
3v0k4: I'm not sure I understand the need of the port going back to JS. Doesn't JS already know the value of the range since it was selected there? Maybe we don't need to tell JS back?

app/javascript/elm/src/MobileSessionsFilters.elm
3v0k4: That's a good question. Maybe we don't need a default model

app/javascript/elm/src/TimeRange.elm
3v0k4: I'd suggest keeping in this module only the data structure `TimeRange`. I think the `view` we could inline in the filters modules (we are going to rm the duplication hopefully).

Also, the `update` is a bit strange cause it's not a normal Elm "update" (ie `update : Msg -> Model -> (Model, Cmd Msg)`). I think the decoder and something like `decodeWithDefault : TimeRange -> Encode.Value -> TimeRange` is enough here.

3v0k4: I'd suggest always starting with the concrete code (inlined in the place using it) and then abstracting to another module if and only if there's a reason ;)

3v0k4: Have you considered DateTimeRange?

app/javascript/elm/src/TimeRange.elm
```diff
@@ -0,0 +1,56 @@
...
+update timeRange newValue =
+    let
+        result =
+            Decode.decodeValue timeRangeDecoder newValue
+    in
+    case result of
+        Ok decodedValue ->
+            TimeRange
+                { timeFrom = decodedValue.timeFrom
+                , timeTo = decodedValue.timeTo
```
3v0k4: I think in this module `epochFrom` and `epochTo` could be a good option for the naming

app/javascript/elm/src/FixedSessionFilters.elm
```diff
@@ -13,19 +15,22 @@ import Ports
...
     { tags = Labels.empty
     , profiles = Labels.empty
+    , timeRange = TimeRange.defaultTimeRange
     }
 
 
 type alias Flags =
     { tags : List String
     , profiles : List String
+    , timeRange : Encode.Value
```
3v0k4: Great that you started playing with decoders. As you said we should start stepping up the game and make the compiler protect us more with types

app/javascript/elm/src/TimeRange.elm
AkkarinForest: I'm not sure why? Epoch just means the date (https://en.wikipedia.org/wiki/Epoch_(computing)) so epochFrom is a bit confusing. Did you mean something like `unixTimeForm`, `unixTimeTo`?

3v0k4: Something like that 👍

PR #129, opened 2019-03-07
New jquery library for time filter

AkkarinForest:
It's not as big as it looks like, I just deleted a lot of stuff from the old libraries :D


app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -59,7 +59,9 @@ export const FixedSessionsMapCtrl = (
       sensorId,
       location: {address: "", indoorOnly: false, streaming: true},
       tags: "",
-      usernames: ""
+      usernames: "",
+      timeFrom: moment().utc().startOf('day').subtract(1, 'year').format('X'),
+      timeTo: moment().utc().endOf('day').format('X')
```
3v0k4: I remember somewhere else we were adjusting the offset after calling `utc()` or something like that. You may want to double check if this or the other should be changed

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -101,10 +103,16 @@ export const FixedSessionsMapCtrl = (
...
+        timeFrom: $scope.params.get('data').timeFrom,
+        timeTo: $scope.params.get('data').timeTo,
+      };
+
       const flags = {
         tags: $scope.params.get('data').tags.split(', ').filter((tag) => tag !== "") || [],
-        profiles: $scope.params.get('data').usernames.split(', ').filter((tag) => tag !== "") || []
-      }
+        profiles: $scope.params.get('data').usernames.split(', ').filter((tag) => tag !== "") || [],
+        timeRange: timeRange
```
3v0k4: When key and value have the same name you can just have `timeRange` (instead of `timeRange: timeRange`)

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -129,10 +137,38 @@ export const FixedSessionsMapCtrl = (
...
+
+      elmApp.ports.updateTimeRange.subscribe((timeRange) => {
+        params.update({data: {timeFrom: timeRange.timeFrom, timeTo: timeRange.timeTo}});
+        $scope.sessions.fetch();
+      });
     });
   }
 }
 
+const setTimeRangeFilter = (params, elmApp) => {
```
3v0k4: What do you say about using "setup" instead of "set" in the naming?

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -129,10 +137,38 @@ export const FixedSessionsMapCtrl = (
...
+  if (document.getElementById("daterange")) {
+    $('input[id="daterange"]').daterangepicker({
```
3v0k4: Have you considered using the same selector in the if and here? I guess the id could be enough

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -129,10 +137,38 @@ export const FixedSessionsMapCtrl = (
...
+      opens: 'left',
+      timePicker: true,
+      startDate: moment.unix((params.get('data').timeFrom)).utc().format('DD/MM/YYYY hh:mm A'),
+      endDate: moment.unix((params.get('data').timeTo)).utc().format('DD/MM/YYYY hh:mm A'),
+      locale: {
+        format: 'DD/MM/YYYY hh:mm A'
+      }
+    }, function(timeFrom, timeTo, _) {
```
3v0k4: `function(timeFrom, timeTo, _)` == `function(timeFrom, timeTo)` in JS

app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
```diff
@@ -152,10 +161,38 @@ export const MobileSessionsMapCtrl = (
         params.update({data: {usernames: profiles.join(", ")}});
         $scope.sessions.fetch();
       });
+
+      setTimeRangeFilter(params, elmApp)
+
+      elmApp.ports.updateTimeRange.subscribe((timeRange) => {
```
3v0k4: You could destructure in here
```
({timeFrom, timeTo} ) => {
```
and then only mention the keys afterwards
```
{timeFrom, timeTo}
```

app/javascript/angular/libs/jquery_ui/jquery.ui.daterangepicker.js
```diff
@@ -0,0 +1,1527 @@
+/**
```
3v0k4: This would be great if we could have it in the package.json

public/partials/fixed_sessions_map.html
```diff
@@ -55,3 +54,5 @@ <h4>Heat Legend</h4>
     </div>
   </div>
 </div>
+
+<link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/daterangepicker/daterangepicker.css" />
```
3v0k4: Maybe we could have this in the layout that is used for both mobile and fixed map?

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -129,10 +137,38 @@ export const FixedSessionsMapCtrl = (
...
   }
 }
 
+const setTimeRangeFilter = (params, elmApp) => {
+  if (document.getElementById("daterange")) {
+    $('input[id="daterange"]').daterangepicker({
+      opens: 'left',
+      timePicker: true,
+      startDate: moment.unix((params.get('data').timeFrom)).utc().format('DD/MM/YYYY hh:mm A'),
+      endDate: moment.unix((params.get('data').timeTo)).utc().format('DD/MM/YYYY hh:mm A'),
```
3v0k4: May be better to pass `timeFrom` and `timeTo` instead of `params` to `setTimeRangeFilter`. The api of that module (ie `get(String).key`) looks ugly. If we keep it closer to Angular it's gonna be easier to get rid of

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
AkkarinForest: Explanation mostly for documentation purposes: I removed the offset adjusting totally. Now the only place we do the adjusting is here and in the setTimeRangeFilter below. These are the places where we want to make sure that the timezones are not taken into account. Cause in general we just want the time since epoch and we don't care about the time zones during sessions fetching and we don't want to switch time zones while sending links to users in other regions, we want to preserve the same hour.

PR #130, opened 2019-03-07
Filter sessions based on new time range.

AkkarinForest:
Some more info in the commits.

### DIFF
--- a/app/controllers/api/regions_controller.rb
+++ b/app/controllers/api/regions_controller.rb
@@ -23,8 +23,6 @@ module Api
     FLOAT_Q_ATTRS = [:north, :south, :east, :west]
     INT_Q_ATTRS = [
       :time_from, :time_to,
-      :day_from, :day_to,
-      :year_from, :year_to,
       :grid_size_x, :grid_size_y
     ]
 
@@ -34,14 +32,8 @@ module Api
       FLOAT_Q_ATTRS.each { |key| data[key] = data[key].to_f if data.key?(key) }
       INT_Q_ATTRS.each { |key| data[key] = data[key].to_i if data.key?(key) }
 
-      data[:time_from] = data[:time_from] || 0
-      data[:time_to] = data[:time_to] || 2359
-
-      data[:day_from] = data[:day_from] || 0
-      data[:day_to] = data[:day_to] || 365
-
-      data[:year_from] = data[:year_from] || 2010
-      data[:year_to] = data[:year_to] || 2050
+      data[:time_from] = data[:time_from] || Time.new(2010).to_i
+      data[:time_to] = data[:time_to] || Time.new(2100).end_of_year.to_i
 
       data[:session_ids] ||= []
 
--- a/app/models/fixed_session.rb
+++ b/app/models/fixed_session.rb
@@ -43,6 +43,7 @@ class FixedSession < Session
   end
 
   def update_end_time!
+    # Measurement.time is a local time, so this are both local end times:
     self.end_time = self.measurements.maximum('time')
     self.end_time_local = self.measurements.maximum('time')
     self.last_measurement_at = DateTime.current
--- a/app/models/measurement.rb
+++ b/app/models/measurement.rb
@@ -32,9 +32,6 @@ class Measurement < ActiveRecord::Base
 
   prepare_range(:longitude_range, 'measurements.longitude')
   prepare_range(:latitude_range, 'measurements.latitude')
-  prepare_range(:time_range, "(EXTRACT(HOUR FROM time) * 60 + EXTRACT(MINUTE FROM time))")
-  prepare_range(:day_range, "(DAYOFYEAR(DATE_ADD(time, INTERVAL (YEAR(NOW()) - YEAR(time)) YEAR)))")
-  prepare_range(:year_range, :time)
 
   geocoded_by :address # field doesn't exist, call used for .near scope inclusion only
 
@@ -73,9 +70,23 @@ class Measurement < ActiveRecord::Base
   end)
 
   scope(:with_time, lambda do |data|
-    day_range(data[:day_from], data[:day_to]).
-      time_range(data[:time_from], data[:time_to]).
-      year_range(Date.new(data[:year_from].to_i),Date.new(data[:year_to].to_i + 1) - 1)
+    time_from = Time.strptime(data[:time_from].to_s, '%s')
+    time_to = Time.strptime(data[:time_to].to_s, '%s')
+
+    where("DATE(time) >= ?", time_from.beginning_of_day)
+      .where("DATE(time) <= ?",  time_to.end_of_day)
+      .minutes_range(Session.minutes_of_day(time_from), Session.minutes_of_day(time_to))
+  end)
+
+  scope(:minutes_range, lambda do |minutes_from, minutes_to|
+    if !Session.whole_day?(minutes_from, minutes_to)
+      field_in_minutes = lambda { |field|
+         "(EXTRACT(HOUR FROM #{field}) * 60 + EXTRACT(MINUTE FROM #{field}))"
+       }
+
+      where("#{field_in_minutes.call('time')} >= ?", minutes_from)
+        .where("#{field_in_minutes.call('time')} <= ?",  minutes_to)
+    end
   end)
 
   def as_indexed_json(options={})
--- a/app/models/session.rb
+++ b/app/models/session.rb
@@ -39,8 +39,6 @@ class Session < ActiveRecord::Base
   validates :url_token, :uuid, :uniqueness => true
   validates :type, :presence => :true
 
-  prepare_range(:start_year_range, :start_time)
-
   accepts_nested_attributes_for :notes, :streams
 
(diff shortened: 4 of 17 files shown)


app/models/fixed_session.rb
```diff
@@ -43,6 +43,7 @@ def after_measurements_created
   end
 
   def update_end_time!
+    # Measurement.time is a local time, so this are both local end times:
     self.end_time = self.measurements.maximum('time')
     self.end_time_local = self.measurements.maximum('time')
     self.last_measurement_at = DateTime.current
```
AkkarinForest: We should change that if we add the utc column to measurement. Or just delete `start_time` and `end_time` from Session.

app/controllers/api/regions_controller.rb
```diff
@@ -34,14 +32,8 @@ def show
...
-      data[:time_from] = data[:time_from] || 0
-      data[:time_to] = data[:time_to] || 2359
-
-      data[:day_from] = data[:day_from] || 0
-      data[:day_to] = data[:day_to] || 365
-
-      data[:year_from] = data[:year_from] || 2010
-      data[:year_to] = data[:year_to] || 2050
+      data[:time_from] = data[:time_from] || Time.new(2010).to_i
+      data[:time_to] = data[:time_to] || Time.new(2100).end_of_year.to_i
```
AkkarinForest: I'm not sure why the defaults here are necessary, I added them just in case, but maybe we don't need them at all.

app/models/fixed_session.rb
3v0k4: Could you elaborate more on this please?

app/models/session.rb
```diff
@@ -125,10 +115,28 @@ def self.from_location(data)
...
+      (:time_from BETWEEN start_time_local AND end_time_local)",
+      :time_from => time_from, :time_to => time_to)
+      .local_minutes_range(minutes_of_day(time_from), minutes_of_day(time_to))
+  end
+
   def self.whole_day?(time_from, time_to)
     time_from == FIRST_MINUTE_OF_DAY && time_to == LAST_MINUTE_OF_DAY
   end
 
+  def self.minutes_of_day(time)
```
3v0k4: Have you considered having this in another module? It seems like it's not related to Session

spec/models/session_spec.rb
```diff
@@ -1,52 +1,97 @@
...
+
+    it "returns sessions that started before the range and ended after the range" do
+      session = create_session(start_time: Time.new(2010, 1, 1), end_time: Time.new(2010, 1, 4))
+
+      actual = Session.filter_by_time_range(Session.all, Time.new(2010, 1, 2).to_i, Time.new(2010, 1, 3).to_i)
+
+      expect(actual.all).to eq([session])
+    end
+
+    it "does not return sessions that dont everlap with the range" do
```
3v0k4: everlap

spec/models/session_spec.rb
```diff
@@ -1,52 +1,97 @@
...
+
+    it "returns sessions that started before the range and ended after the minutes range" do
+      session = create_session(start_time: Time.new(2010, 1, 2, 1, 1), end_time: Time.new(2010, 1, 2, 1, 4))
+
+      actual = Session.filter_by_time_range(Session.all, Time.new(2010, 1, 1, 1, 2).to_i, Time.new(2010, 1, 3, 1, 3).to_i)
+
+      expect(actual.all).to eq([session])
+    end
+
+    it "does not return sessions that dont everlap with the minutes range even when they are in the date range" do
```
3v0k4: everlap

app/models/session.rb
```diff
@@ -49,26 +47,25 @@ class Session < ActiveRecord::Base
...
-
-    where "
-      (#{field_in_minutes.call('start_time_local')} BETWEEN :start_minutes AND :end_minutes)
-      OR
-      (#{field_in_minutes.call('end_time_local')} BETWEEN :start_minutes AND :end_minutes)
-      OR
-      (:start_minutes BETWEEN #{field_in_minutes.call('start_time_local')} AND #{field_in_minutes.call('end_time_local')})
-    ", :start_minutes => start_minutes, :end_minutes => end_minutes
+  scope :local_minutes_range, lambda { |minutes_from, minutes_to|
+    if !whole_day?(minutes_from, minutes_to)
```
3v0k4: Have you considered `unless`?

3v0k4: `whole_day?` is another thing that doesn't sound related to Session

app/models/session.rb
```diff
@@ -49,26 +47,25 @@ class Session < ActiveRecord::Base
...
+      field_in_minutes = lambda { |field|
```
3v0k4: Wow, we need to come back to this query at some point

app/models/session.rb
```diff
@@ -125,10 +115,28 @@ def self.from_location(data)
...
+    time_to = Time.strptime(time_to.to_s, '%s')
+
+    sessions.where(
+      "(start_time_local BETWEEN :time_from AND :time_to)
+      OR
+      (end_time_local BETWEEN :time_from AND :time_to)
+      OR
+      (:time_from BETWEEN start_time_local AND end_time_local)",
+      :time_from => time_from, :time_to => time_to)
+      .local_minutes_range(minutes_of_day(time_from), minutes_of_day(time_to))
```
3v0k4: Would it be possible to have one query instead of two? I mean merging `filter_by_time_range` and `local_minutes_range` together?

app/models/session.rb
```diff
@@ -125,10 +115,28 @@ def self.from_location(data)
     where(:streams => {:id => streams_ids.uniq})
   end
 
+  def self.filter_by_time_range(sessions, time_from, time_to)
+    time_from = Time.strptime(time_from.to_s, '%s')
+    time_to = Time.strptime(time_to.to_s, '%s')
```
3v0k4: We should not allow url params to get this deep into the code unchecked. The controller should take all the params, validate them and convert to what deeper objects need. We should start moving in that direction asap.

This is an outdated example of that (https://github.com/3v0k4/rails_sane_architecture/blob/master/app/forms/comment_form.rb). The readme explains the structure https://github.com/3v0k4/rails_sane_architecture. I'd prolly use https://dry-rb.org/gems/dry-validation/ and https://dry-rb.org/gems/dry-struct/ nowadays for that

app/models/measurement.rb
```diff
@@ -73,9 +70,23 @@ class Measurement < ActiveRecord::Base
...
-      year_range(Date.new(data[:year_from].to_i),Date.new(data[:year_to].to_i + 1) - 1)
+    time_from = Time.strptime(data[:time_from].to_s, '%s')
+    time_to = Time.strptime(data[:time_to].to_s, '%s')
+
+    where("DATE(time) >= ?", time_from.beginning_of_day)
+      .where("DATE(time) <= ?",  time_to.end_of_day)
+      .minutes_range(Session.minutes_of_day(time_from), Session.minutes_of_day(time_to))
+  end)
+
+  scope(:minutes_range, lambda do |minutes_from, minutes_to|
```
3v0k4: Do we need to keep this query separated from `with_time` or could they be one scope?

app/models/fixed_session.rb
AkkarinForest: Sure. So we do:
```
self.end_time = self.measurements.maximum('time')
self.end_time_local = self.measurements.maximum('time')
```

And this `maximum('time')` comes from `Measurement.time` column, which as we know is saved as local time. So the `Session.end_time` is the same as `Session.end_local_time` here. 
I think there is no reason to have this data duplicated in the database. So we should:
- either remove the `end_time` column from the db
- or change `end_time` to store the time in utc. (eg by adding the `utc_time` column to Measurements and using it to create `self.end_time = self.measurements.maximum('utc_time')`)

And I think there is a similar problem might be with `start_time` and `start_local_time`.

app/models/session.rb
AkkarinForest: I'm not sure how to make it just one query since the `local_minutes_range` is only performed `unless whole_day?`.

But can I ask why you suggested that? Is it about readability? Or about performance? Cause in the latter case I think it doesn't change anything, rails take care of that. They don't execute every sql line immediately. They collect all the code that looks like sql, put it together and only execute when the resulting object is actually needed, right?

app/models/fixed_session.rb
3v0k4: Would you mind creating a ticket somewhere and documenting this please?

app/models/session.rb
3v0k4: Yeah, the query builder is lazy that's correct.

I've stared at the code for some time and I couldn't really get why there were two queries when the column is one. Thus readability is something we could work on.

The perf is something we can take a look later on I guess. But in case we wanted to use indexes here, at the moment the queries are working on
- `end_time_local`
- `start_time_local`
- `EXTRACT(HOUR FROM end_time_local)`
- `EXTRACT(HOUR FROM start_time_local)`
When a SQL function is applied to a column then the index can't be used. In other words, an index on `end_time_local` would not be used for `EXTRACT(HOUR FROM end_time_local)`

PR #133, opened 2019-03-13
Clustering fixed sessions

AkkarinForest:
We have a new design for clustering fixed sessions. It is a standard approach so I'm using a MarkerClustererPlus library. 
The ticket: https://llp.kanbanery.com/projects/8040/board/tasks/2392185
Some files are just restored from the previous clustering. Of course, you can still look at them if you see some room for improvement ;)


app/javascript/angular/code/services/google/_map.js
```diff
@@ -149,15 +151,35 @@ export const map = (
...
+        calculator: (markers, stylesCount) => {
+          const average = markers.reduce((sum, marker) => sum + marker.value(), 0) / markers.length
+          return { text: "", index: heat.getLevel(Math.round(average)) }
+        }
+      };
+
+      const markerClusterer = new MarkerClusterer(this.mapObj, this.markers, options);
+
+      googleMaps.listen(markerClusterer, 'clusterclick', onClick);
+    },
```
AkkarinForest: I couldn't figure out how to test it, even using sinon. I can't find a way to stub MarkerClusterer. On the other hand, this method just basically creates some settings and passes it to the library, which I'm assuming is tested by google.

app/javascript/angular/code/services/google/_map.js
```diff
@@ -149,15 +151,35 @@ export const map = (
...
+          { url: '/assets/marker2.png', height: 10, width: 10 },
+          { url: '/assets/marker3.png', height: 10, width: 10 },
+          { url: '/assets/marker4.png', height: 10, width: 10 },
+        ],
+        zoomOnClick: false,
+        calculator: (markers, stylesCount) => {
+          const average = markers.reduce((sum, marker) => sum + marker.value(), 0) / markers.length
+          return { text: "", index: heat.getLevel(Math.round(average)) }
+        }
+      };
```
AkkarinForest: I guess it's not easy to understand how this works. In short the index returned by the calculator is used to choose a style from the styles array. It's documented here: https://htmlpreview.github.io/?https://github.com/googlemaps/v3-utility-library/blob/master/markerclustererplus/docs/reference.html#ClusterIconStyle

app/javascript/angular/code/services/google/custom_marker.js
```diff
@@ -55,5 +55,16 @@ export function buildCustomMarker(latLng, content, colorClass, callback, type) {
           });
         });
   };
-  return new CustomMarker(latLng, content, colorClass, callback, type);
+
+  CustomMarker.prototype.getPosition = function() {
+    return new google.maps.LatLng({ lat: this.position.lat(), lng: this.position.lng() });
+  };
+
+  CustomMarker.prototype.getDraggable = () => {};
```
AkkarinForest: `getPosition` and `getDraggable` are required to make the clusterer library work.

app/javascript/angular/code/services/google/info_window.js
```diff
@@ -1,29 +1,39 @@
+import constants from '../../constants'
```
AkkarinForest: This file is just restored from the previous clustering.

app/services/fixed_region_info.rb
```diff
@@ -0,0 +1,53 @@
+class FixedRegionInfo
```
AkkarinForest: This file is just restored from the previous clustering.

spec/services/fixed_region_info_spec.rb
```diff
@@ -0,0 +1,125 @@
+require "rails_helper"
```
AkkarinForest: This file is just restored from the previous clustering.

package.json
```diff
@@ -6,6 +6,7 @@
     "test": "NODE_ENV=test blue-tape -r babel-register app/javascript/angular/tests/**/*.test.js"
   },
   "dependencies": {
+    "@google/markerclustererplus": "^2.1.11",
```
3v0k4: Great that you could put the dependency in here!

app/controllers/api/regions_controller.rb
```diff
@@ -28,16 +28,19 @@ class RegionsController < BaseController
...
       data = ActiveSupport::JSON.decode(params[:q]).symbolize_keys
+      data[:session_ids] ||= []
 
-      FLOAT_Q_ATTRS.each { |key| data[key] = data[key].to_f if data.key?(key) }
-      INT_Q_ATTRS.each { |key| data[key] = data[key].to_i if data.key?(key) }
-
-      data[:time_from] = data[:time_from] || Time.new(2010).to_i
-      data[:time_to] = data[:time_to] || Time.new(2100).end_of_year.to_i
+      if (data[:session_ids] != [] && Session.find(data[:session_ids].first).fixed?)
+        respond_with FixedRegionInfo.new.call(data)
```
3v0k4: `FixedRegionInfo` is used only for fixed sessions and `RegionInfo` only for mobile ones? Have you considered renaming the latter?

app/controllers/api/regions_controller.rb
```diff
@@ -28,16 +28,19 @@ class RegionsController < BaseController
...
+      else
+        FLOAT_Q_ATTRS.each { |key| data[key] = data[key].to_f if data.key?(key) }
+        INT_Q_ATTRS.each { |key| data[key] = data[key].to_i if data.key?(key) }
 
-      data[:session_ids] ||= []
+        data[:time_from] = data[:time_from] || Time.new(2010).to_i
+        data[:time_to] = data[:time_to] || Time.new(2100).end_of_year.to_i
 
-      respond_with RegionInfo.new(data)
+        respond_with RegionInfo.new(data)
```
3v0k4: I'm not sure what's the difference in the resulting json when the code enters the first or the second branch. Maybe this should be two endpoints w/o ifs?

3v0k4: I don't have an up-to-date gist on how I'd structure Rails apps at the moment. I was thinking I could do that next week. Maybe I could update https://github.com/3v0k4/rails_sane_architecture#the-meat-architecture and make it happen in one endpoint in AirCasting as a proof of concept.

If we like the result we could start moving towards that every timer we touch a controller.

app/controllers/api/regions_controller.rb
```diff
@@ -28,16 +28,19 @@ class RegionsController < BaseController
...
     def show
       data = ActiveSupport::JSON.decode(params[:q]).symbolize_keys
+      data[:session_ids] ||= []
 
-      FLOAT_Q_ATTRS.each { |key| data[key] = data[key].to_f if data.key?(key) }
-      INT_Q_ATTRS.each { |key| data[key] = data[key].to_i if data.key?(key) }
-
-      data[:time_from] = data[:time_from] || Time.new(2010).to_i
-      data[:time_to] = data[:time_to] || Time.new(2100).end_of_year.to_i
+      if (data[:session_ids] != [] && Session.find(data[:session_ids].first).fixed?)
```
3v0k4: Have you thought if there's any way of avoiding the query? `Session.find(data[:session_ids].first).fixed?`

app/javascript/angular/code/services/google/_map.js
```diff
@@ -149,15 +151,37 @@ export const map = (
       return newMarker;
     },
 
-    drawCustomMarker: function({ latLng, content, colorClass, callback, type }) {
-      const customMarker = buildCustomMarker(latLng, content, colorClass, callback, type);
+    drawCustomMarker: function({ latLng, content, colorClass, callback, type, objectId, value }) {
+      const customMarker = buildCustomMarker(latLng, content, colorClass, callback, type, objectId, value );
```
3v0k4: Have you thought about passing the entire obj instead of destructuring and passing each as a single arg?

app/javascript/angular/code/services/google/_map.js
```diff
@@ -149,15 +151,37 @@ export const map = (
...
+      const options = {
+        styles: [
+          { url: '/assets/marker1.png', height: 10, width: 10 },
+          { url: '/assets/marker2.png', height: 10, width: 10 },
+          { url: '/assets/marker3.png', height: 10, width: 10 },
+          { url: '/assets/marker4.png', height: 10, width: 10 },
+        ],
+        zoomOnClick: false,
+        gridSize: 20,
+        calculator: (markers, stylesCount) => {
```
3v0k4: I guess you could remove `stylesCount`

app/javascript/angular/code/services/google/_map.js
3v0k4: I guess a comment in code would be a good first step

app/javascript/angular/code/services/google/custom_marker.js
3v0k4: If you feel like it would help ourselves in the future maybe you could add a comment in the code to say that

app/javascript/angular/code/services/google/info_window.js
```diff
@@ -1,29 +1,39 @@
...
+    show: function(url, data, position, sessionType){
       this.popup.setContent("working..");
       this.popup.setPosition(position);
       this.popup.open(map.get());
-      $timeout(function(){
-        $http.get(url, {params : data, cache: true}).success(_(self.onShowData).bind(self));
+      const htmlPath = (sessionType === constants.fixedSession) ? FIXED_INFO_WINDOW_PATH : MOBILE_INFO_WINDOW_PATH;
+
+      $timeout(() => {
+        $http.get(url, {params : data, cache: true}).success((data, status, headers, config) => this.onShowData(data, status, headers, config, htmlPath));
```
3v0k4: Are all of `data, status, headers, config` used?

app/javascript/angular/tests/_fixed_sessions.test.js
```diff
@@ -360,6 +361,20 @@ test('drawSessionsInLocation draws colorcoded marker for currently streaming ses
   t.end();
 });
 
+test('drawSessionsInLocation calls map.clusterMarkers for currently streaming sessions when sensor selected', t => {
+  const clusterMarkers = sinon.spy();
+  const map = { clusterMarkers }
+  const sensors = { anySelected: () => true, selectedSensorName: () => "sensorName"};
+  const data = buildData({ location: { streaming: true } });
+
```
3v0k4: I guess yiou can rm this line

app/javascript/angular/tests/_fixed_sessions.test.js
```diff
@@ -360,6 +361,20 @@ test('drawSessionsInLocation draws colorcoded marker for currently streaming ses
   t.end();
 });
 
+test('drawSessionsInLocation calls map.clusterMarkers for currently streaming sessions when sensor selected', t => {
+  const clusterMarkers = sinon.spy();
+  const map = { clusterMarkers }
+  const sensors = { anySelected: () => true, selectedSensorName: () => "sensorName"};
+  const data = buildData({ location: { streaming: true } });
```
3v0k4: Must be `streaming: true` for the test to pass?

app/javascript/angular/tests/_fixed_sessions.test.js
```diff
@@ -360,6 +361,20 @@ test('drawSessionsInLocation draws colorcoded marker for currently streaming ses
   t.end();
 });
 
+test('drawSessionsInLocation calls map.clusterMarkers for currently streaming sessions when sensor selected', t => {
+  const clusterMarkers = sinon.spy();
+  const map = { clusterMarkers }
+  const sensors = { anySelected: () => true, selectedSensorName: () => "sensorName"};
```
3v0k4: Are both keys needed for the test to pass?

spec/services/fixed_region_info_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+
+ def create_session_with_streams_and_measurements!(attributes)
+  session = create_session!(id: attributes.fetch(:id), user: attributes.fetch(:user))
+  stream = create_stream!(session: session, min_latitude: attributes.fetch(:min_latitude, 1.0), max_latitude: attributes.fetch(:max_latitude, 1.0), min_longitude: attributes.fetch(:min_longitude, 1.0), max_longitude: attributes.fetch(:max_longitude, 1.0))
+  create_measurements!(stream: stream, value: attributes.fetch(:value), count: attributes.fetch(:count))
+
+   session
+end
+
+ def create_session!(attributes)
```
3v0k4: We should prolly be extracting those `create_*` to a common place

app/javascript/angular/tests/_map.test.js
```diff
@@ -229,6 +229,19 @@ test('removeAllMarkers removes all markers', t => {
   t.end();
 });
 
+test('removeAllMarkers clears the clustering', t => {
+  const clearMarkers = sinon.spy()
+  const service = _map({})
+  service.clusterer = { clearMarkers }
```
3v0k4: This is a bit hacky. Maybe extracting the "clusterer" to another file would make it more testable but not sure

app/controllers/api/regions_controller.rb
AkkarinForest: I considered creating a new separate endpoint for fixed sessions, but I wasn't sure if that's a good idea. 
I guess I could also send a flag from the frontend like `data[:session_type]`?

app/controllers/api/regions_controller.rb
AkkarinForest: Sure, it's a good idea. The endpoints in aircasting are a mess and we should definitely have some plan to organize them and start moving in this direction.
In that topic - have you ever used GraphQL? It might be something that we could use to organize and lower the number of our endpoints.

spec/services/fixed_region_info_spec.rb
AkkarinForest: I thought that keeping them in the same file gives us the advantage of seeing where everything comes from and what exactly is happening. 
If we move them somewhere else then why not just use the factory girl (or upgrade to factory bot)?

app/controllers/api/regions_controller.rb
3v0k4: When it comes to methods there's a smell for cases with a "control parameter" that does different things depending on a flag
```
def call(flag)
  flag ? do_something : do_something_else
end
```
https://www.rubydoc.info/github/troessner/reek/Reek/Smells/ControlParameter

Not totally sure this is as valid when it comes to endpoint. Personally I'd consider two different endpoints. I think it's more explicit

app/controllers/api/regions_controller.rb
3v0k4: GraphQL is cool. And I thought of you when I've seen that in Elm Europe they are gonna run a workshop about it (http://2019.elmeurope.org/#schedule).

I think that would come with some different challenges but I'm totally open to explore the idea!

spec/services/fixed_region_info_spec.rb
3v0k4: I think it's explicit enough what happens: we explicitly pass the parameters that we need in the created object.

I'd personally advise against FactoryGirl. Also the authors of the gem say
> My general advice, though, is to avoid Factory Girl as much as is reasonably possible. Not because it’s bad or unreliable software (Factory Girl is very reliable; we’ve used it successfully since 2008), but because its inherent persistence mechanism is calling #save! on the object, which will always take longer than not persisting data.
https://thoughtbot.com/blog/speed-up-tests-by-selectively-avoiding-factory-girl

app/controllers/api/regions_controller.rb
AkkarinForest: All right, for now, I divided it into two endpoints. But we should go back to this topic and create some plan to organize routes.

app/javascript/angular/tests/_fixed_sessions.test.js
AkkarinForest: Yes, we only want to cluster for currently streaming sessions. In other cases, the method returns early.

spec/services/fixed_region_info_spec.rb
AkkarinForest: I'm still don't see why using factory bot would be a bad thing? In this file, we are using `create!` everywhere, which executes `save`, using not using factory girl doesn't improve the tests speed?

It's probably a sign of a bad code structure, cause we are executing sql queries all over the code. If we ware using a repository in that code, we could inject the repo in this test and the just do `Object.new`. But that would be a bigger refactor.

3v0k4: The problem with FactoryGirl is that it hides database queries (too) conveniently. In many projects I've seen situations where a single Factory.create! fired 5+ queries when actually only one was needed. That situation times many unit tests makes the entire suite incredibly slow (for no reasons).

PR #135, opened 2019-03-14
Validate params

AkkarinForest:
I added the `dry-validation` gem to see how it will work. What do you think? Is that the direction we want to take?


3v0k4:
Let's have a chat tomorrow about this! Thanks for the initiative!

PR #136, opened 2019-03-19
Copy URLs button

AkkarinForest:
I still have to add the same changes to mobile sessions, but I wanted to get your opinion first. 

Also, I am not sure how to create a popup window in elm. I mean something similar to what we have now for copying links. I used `if else` in the view but that doesn't seem elegant?

Deploy together with https://github.com/LunarLogic/playing-zone/pull/253


app/views/layouts/map.html.haml
```diff
@@ -3,6 +3,7 @@
   %head
     %meta(name='viewport' content='width=device-width,initial-scale=0.5')
     %meta(name='description' content='AirCasting map showing health and environmental data like air quality, temperature, humidity, sound levels and more.')
+    %script{:src => "https://cdn.jsdelivr.net/npm/clipboard@2/dist/clipboard.min.js"}
```
3v0k4: I see that this is in package.json too. Is this line needed?

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -148,6 +149,15 @@ export const FixedSessionsMapCtrl = (
       }
 
       setupTimeRangeFilter(elmApp, $scope.sessions, callback,  params.get('data').timeFrom, params.get('data').timeTo);
+
+      new Clipboard('.copy-link');
+      elmApp.ports.requestCurrentUrl.subscribe(() => {
+        const longUrl = window.location.href
+
+        $http.get("api/short_url", { params: { longUrl: longUrl }})
```
3v0k4: Please use `fetch` in place of `$http`. Since we are moving away from angular better not to use its helpers

3v0k4: `{ longUrl: longUrl }` == { longUrl }`

app/services/url_shortener.rb
```diff
@@ -0,0 +1,27 @@
...
+    begin
+      result = http_utils.post(
+        url: "https://api-ssl.bitly.com/v4/shorten",
+        token: A9n.bitly_access_token,
+        body: { "long_url" => link },
+      )
+      if http_utils.successful?(result)
+        link = http_utils.body(result)["link"]
+      end
+    rescue
```
3v0k4: This returns `nil`, I guess you wanted the have the link if that fails?

app/services/url_shortener.rb
```diff
@@ -0,0 +1,27 @@
+require 'net/http'
+require 'json'
+
+class UrlShortener
+  def initialize(http_utils: HttpUtils.new)
+    @http_utils = http_utils
```
3v0k4: What do you think of just naming it Http w/o utils?

app/services/url_shortener.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def call(link)
+    begin
+      result = http_utils.post(
+        url: "https://api-ssl.bitly.com/v4/shorten",
+        token: A9n.bitly_access_token,
+        body: { "long_url" => link },
+      )
+      if http_utils.successful?(result)
+        link = http_utils.body(result)["link"]
```
3v0k4: Could you avoid the mutation of link here?

app/services/url_shortener.rb
3v0k4: Also, this should prolly be handled by the http service

app/services/url_shortener.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def call(link)
+    begin
+      result = http_utils.post(
+        url: "https://api-ssl.bitly.com/v4/shorten",
+        token: A9n.bitly_access_token,
+        body: { "long_url" => link },
+      )
+      if http_utils.successful?(result)
```
3v0k4: This obliges the client of `HttpUtils` to call twice on that. What if you return a `Result` as in https://github.com/HabitatMap/AirCasting/pull/137 ?

3v0k4: Or you could do something like `post_or_default(default: long_link)` then http would return either the short or the long

3v0k4: If `Result` had a method `or_value(default)` that gives back the value if `Success` or default if `Failure` then you could do
```
http_utils.post(...).or_value(long_link) 
```
like `Maybe.with_default`

PR #137, opened 2019-03-19
validate / coerce http params

AkkarinForest:
**PROOF OF CONCEPT**

More comments in the files changed.

This is the errors returned as an example. The errors returned are not perfect, we could improve later. The point here is that we have a more structured way of handling incoming params

curl -H "Accept: application/json" -H "Content-type: application/json" "http://localhost:3000/api/fixed_region"
-> {"sensor_name":["sensor_name is missing"],"session_ids":["session_ids is missing"]}

curl -H "Accept: application/json" -H "Content-type: application/json" "http://localhost:3000/api/fixed_region?q=%7B%22west%22:-105.46874
999999999,%22east%22:-104.64154411764706,%22south%22:39.42835968965721,%22north%22:40.15181583075184,%22time_from%22:%221521072000%22,%22time_to%22:%221552694399%22,%22grid_
size_x%22:68.6695278969957,%22grid_size_y%22:25,%22tags%22:%22%22,%22usernames%22:%22%22,%22sensor_name%22:%22airbeam2-pm2.5%22,%22measurement_type%22:%22Particulate+Matter%
22,%22unit_symbol%22:%22%C2%B5g%2Fm%C2%B3%22,%22session_ids%22:%5B83643,83642,83641,83638,83637,83635,83613,83598,83596,83590,83589,83588,83587,83586,83584,83583,83582,83581
,83580,83579,83578,83577,83576,83575,83574,83573,83572,83570,83566,83565,83564,83563,83562,83561,83559,83558,83557,83555,83554,83553,83552,83551,83550,83549,83548,83546,8354
5,83543,83539,83538%5D%7D"
-> 
{"average":0.09244588161614373,"number_of_contributors":4,"number_of_samples":3522,"number_of_instruments":48}

curl -H "Accept: application/json" -H "Content-type: application/json" "http://localhost:3000/api/fixed_region?q=\{\"sensor_name\":\"name
\",\"session_ids\":\[\]TTTTTTTTTTTTTTTTT\}"
->
could not parse request


config/environments/development.rb
```diff
@@ -13,7 +13,7 @@
   config.eager_load = false
 
   # Show full error reports and disable caching.
-  config.consider_all_requests_local       = true
+  config.consider_all_requests_local       = false
```
3v0k4: This should be removed. I've used it to have Rails return the reals errors with curl instead of the stacktrace and all of that

app/models/api/fixed_region.rb
```diff
@@ -0,0 +1,18 @@
+require 'dry-validation'
```
3v0k4: The idea is that for each "resource" we are expecting to get via http we just create a new file like this with
- validation / schema
- coercion / struct

The rest of the PR is reusable infrastructure

app/services/form.rb
```diff
@@ -0,0 +1,38 @@
...
+    @schema.call(params).messages(full: true)
+  end
+
+  def to_h
+    @struct.new(
+      @schema.call(params).to_h
+    )
+  end
+
+  def add_error(error)
```
3v0k4: Not used for now but it's useful in case we want from the service to return added errors

PR #141, opened 2019-03-25
Decouple storage

AkkarinForest:
I'm trying to disconnect stuff from the storage. When we finish rewriting location filter and heat map I think we will be able to remove storage totally.
The side effect is that we have to access stuff directly from params, but I'm hoping this is temporary, cause we are planning to move params to elm, so I didn't want to put a lot of effort in optimizing that part.

Let me know if you think this is a good idea and if it's going in the right direction.

### DIFF
--- a/app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
+++ b/app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
@@ -54,7 +54,7 @@ export const MobileSessionsMapCtrl = (
       .get('data', { sensorId: sensors.defaultSensorId })
       .sensorId;
 
-    storage.updateDefaults({
+    const defaults = {
       sensorId,
       location: {address: ""},
       tags: "",
@@ -63,11 +63,11 @@ export const MobileSessionsMapCtrl = (
       crowdMap: false,
       timeFrom: moment().utc().startOf('day').subtract(1, 'year').format('X'),
       timeTo: moment().utc().endOf('day').format('X')
-    });
+    };
 
     if (!params.get('data').heat) sensors.fetchHeatLevels();
 
-    storage.updateFromDefaults();
+    params.updateFromDefaults(defaults);
   };
 
   $scope.searchSessions = function() {
--- a/app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
+++ b/app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
@@ -86,9 +86,9 @@ export const MobileSessionsMapCtrl = (
     console.log("watch - params.get('data').heat - ", newValue, " - ", oldValue);
     if (newValue === oldValue) return;
 
-    if (storage.isCrowdMapLayerOn() && mobileSessions.noOfSelectedSessions() === 0) {
+    if ($scope.params.get('data').crowdMap && mobileSessions.noOfSelectedSessions() === 0) {
       sessionsUtils.updateCrowdMapLayer(mobileSessions.sessionIds());
-    } else if (storage.isCrowdMapLayerOn() && mobileSessions.noOfSelectedSessions() === 1) {
+    } else if ($scope.params.get('data').crowdMap && mobileSessions.noOfSelectedSessions() === 1) {
       sessionsUtils.updateCrowdMapLayer($scope.params.get('selectedSessionIds'));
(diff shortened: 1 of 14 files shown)


app/javascript/angular/code/services/_mobile_sessions.js
```diff
@@ -67,7 +67,7 @@ export const mobileSessions = (
 
 
     onSessionsFetch: function() {
-      if (!storage.isCrowdMapLayerOn()) {
+      if (!params.get('data').crowdMap) {
```
gitjul: I'd change `crowdMap` to `isCrowdMap` since it's a boolean

3v0k4: I'd suggest adding something like `isCrowdMapOn` cause that is a feature that is either on or off, correct?

app/javascript/angular/code/services/params.js
```diff
@@ -34,8 +34,25 @@ angular.module("aircasting").factory('params', ['$location', '$rootScope', 'util
...
+    },
   };
   return new Params();
 }]);
 
+
+const filterAlreadySetParams = (defaults, currentParamsData = {}) => {
+  return Object.keys(defaults).reduce((acc, key) => {
+    if (paramNotSet(key, currentParamsData)) {
+      acc[key] = defaults[key];
```
3v0k4: Please try to avoid mutations when possible

app/javascript/angular/code/services/params.js
```diff
@@ -34,8 +34,25 @@ angular.module("aircasting").factory('params', ['$location', '$rootScope', 'util
...
+    },
+    updateData: function(newData) {
+      this.update({data: utils.merge(this.paramsData.data || {}, newData)})
+    },
   };
   return new Params();
 }]);
 
+
+const filterAlreadySetParams = (defaults, currentParamsData = {}) => {
```
3v0k4: Is this the same as `{ ...defaults, ...currentParamsData }`?

app/javascript/angular/code/services/_update_crowd_map_layer.js
```diff
@@ -15,7 +14,7 @@ export const updateCrowdMapLayer = (
 ) => ({
   call: (sessionIds) => {
     map.clearRectangles();
-    if (!storage.isCrowdMapLayerOn()) return;
+    if (!params.get('data').crowdMap) return;
```
gitjul: how about using the `isCrowdMapOn` function in all places like that?

app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
```diff
@@ -86,9 +86,9 @@ export const MobileSessionsMapCtrl = (
     console.log("watch - params.get('data').heat - ", newValue, " - ", oldValue);
     if (newValue === oldValue) return;
 
-    if (storage.isCrowdMapLayerOn() && mobileSessions.noOfSelectedSessions() === 0) {
+    if ($scope.params.get('data').crowdMap && mobileSessions.noOfSelectedSessions() === 0) {
```
gitjul: how about using the `isCrowdMapOn` function in all places like that?

app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
```diff
@@ -86,9 +86,9 @@ export const MobileSessionsMapCtrl = (
...
       sessionsUtils.updateCrowdMapLayer(mobileSessions.sessionIds());
-    } else if (storage.isCrowdMapLayerOn() && mobileSessions.noOfSelectedSessions() === 1) {
+    } else if ($scope.params.get('data').crowdMap && mobileSessions.noOfSelectedSessions() === 1) {
```
gitjul: like above ^

app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
```diff
@@ -131,13 +131,13 @@ export const MobileSessionsMapCtrl = (
 
       const elmApp = Elm.MobileSessionsFilters.init({ node: node, flags: flags });
 
-      elmApp.ports.toggleCrowdMap.subscribe(() => {
-        storage.toggleCrowdMapData();
+      elmApp.ports.toggleCrowdMap.subscribe((crowdMap) => {
```
gitjul: shown offline - just a note: `(param) =>` can be written as `param =>`

PR #142, opened 2019-03-26
Sessions list in Elm

3v0k4:
This is the first PR on the sessions list. It introduces what was in the proof of concept with added tests. In particular:
- list in Elm
- radio buttons instead of checkboxes

### DIFF
--- a/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
+++ b/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
@@ -122,18 +122,6 @@ export const SessionsListCtrl = (
     });
   }, true);
 
-  $scope.toggleAll = function(){
-    if(sessions.hasSelectedSessions()) {
-      sessions.deselectAllSessions();
-    } else {
-      flash.set(CANNOT_SELECT_MULTIPLE_SESSIONS);
-    }
-  };
-
-  $scope.allSelectionText = function() {
-    return sessions.hasSelectedSessions() ? "none" : "all";
-  };
-
   $scope.toggleSession = function(sessionId, markerSelected) {
     if(this.isSessionDisabled(sessionId)){
       flash.set(CANNOT_SELECT_MULTIPLE_SESSIONS);
--- a/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
+++ b/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
@@ -1,5 +1,6 @@
 import _ from 'underscore';
 import { formatSessionForList } from '../values/session'
+import { Elm } from '../../../elm/src/SessionsList.elm';
 
 export const SessionsListCtrl = (
   $scope,
@@ -18,6 +19,7 @@ export const SessionsListCtrl = (
 ) => {
   let sessions;
   let singleSession;
+  let elmApp;
   const CANNOT_SELECT_MULTIPLE_SESSIONS = "You can't select multiple sessions";
 
   $scope.setDefaults = function() {
@@ -93,6 +95,7 @@ export const SessionsListCtrl = (
   $scope.$watch("newSessionsForList()", function(newSessions, oldSessions) {
     console.log("newSessionsForList()", newSessions, oldSessions);
     $scope.sessionsForList = newSessions;
+    if (elmApp) elmApp.ports.updateSessions.send(newSessions.map(formatSessionForElm));
   }, true);
 
   $scope.$on('markerSelected', function(event, data){
@@ -172,4 +175,26 @@ export const SessionsListCtrl = (
   };
 
   $scope.setDefaults();
-};
+
+  if (process.env.NODE_ENV !== 'test') {
+    angular.element(document).ready(() => {
+      const node = document.getElementById('sessions-bottom-elm');
+      const flags = $scope.sessions.get().map(formatSessionForList).map(formatSessionForElm);
+      elmApp = Elm.SessionsList.init({ node, flags });
+
+      elmApp.ports.checkedSession.subscribe(({ selected, deselected }) => {
+        if (deselected) $scope.toggleSession(deselected, true);
+        if (selected) $scope.toggleSession(selected, true);
+        $scope.$apply();
+      });
+
+      elmApp.ports.loadMoreSessions.subscribe(() => {
+        $scope.updateSessionsPage();
+        $scope.$apply();
+      });
+    });
+  };
+}
+
+const formatSessionForElm = s =>
+  ({ ...s , shortTypes: s.shortTypes.map(({ name, type }) => ({ name, type_: type })) });
--- a/app/javascript/elm/src/SessionsList.elm
+++ b/app/javascript/elm/src/SessionsList.elm
@@ -7,6 +7,7 @@ import Html.Events as Events
 import String exposing (fromInt)
 
 
+
 ---- MODEL ----
 
 
@@ -69,6 +70,7 @@ update msg model =
                 ( { model | selectedSessionId = Nothing }
                 , checkedSession { deselected = model.selectedSessionId, selected = Nothing }
                 )
+
             else
                 ( { model | selectedSessionId = Just id }
                 , checkedSession { deselected = model.selectedSessionId, selected = Just id }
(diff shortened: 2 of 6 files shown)


app/javascript/elm/src/SessionsList.elm
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+
+type alias Session =
+    { title : String
+    , id : Int
+    , timeframe : String
+    , username : String
+    , shortTypes : List ShortType
+    , selected : Bool
```
3v0k4: ?

app/javascript/elm/src/SessionsList.elm
```diff
@@ -0,0 +1,188 @@
+port module SessionsList exposing (Model, Msg(..), Session, ShortType, checkedSession, init, main, subscriptions, update, updateSessions, view, viewSession, viewShortType, loadMoreSessions)
+
+import Browser
+import Html exposing (Html, button, dd, div, dl, dt, h1, img, input, label, li, section, span, text, ul)
+import Html.Attributes exposing (checked, class, for, id, src, style, type_)
+import Html.Events exposing (onClick)
+import String exposing (fromInt)
```
AkkarinForest: It would be cool if we used the same importing/exposing style, so it will be easier to merge later. In other elm files we have:
- import Html exposing (whatever you need)
- import Html.Attributes as Attr
- import Html.Events as Events

but I'm open to using a different strategy as long as it's consistent. What do you think?

app/javascript/elm/src/SessionsList.elm
```diff
@@ -0,0 +1,188 @@
...
+
+type alias Model =
+    { sessions : List Session
+    , selectedSessionId : Maybe Int
+    }
+
+
+init : List Session -> ( Model, Cmd Msg )
+init flags =
+    ( { sessions = flags, selectedSessionId = Nothing }, Cmd.none )
```
AkkarinForest: Does this mean that sending a link with preselected session doesn't work? We should add that before deploying to production.

app/javascript/elm/src/SessionsList.elm
```diff
@@ -0,0 +1,192 @@
...
+
+
+
+---- VIEW ----
+
+
+viewSession : Maybe Int -> Session -> Html Msg
+viewSession selectedSessionId session =
+    li
+        [ style "padding" "5px 10px"
```
AkkarinForest: What do you think about keeping styling in some css file and just using classes here?

app/javascript/elm/src/SessionsList.elm
```diff
@@ -0,0 +1,192 @@
...
+        )
+
+
+viewLoadMore : Int -> Html Msg
+viewLoadMore sessionCount =
+    if sessionCount /= 0 && modBy 50 sessionCount == 0 then
+        li [] [ button [ onClick LoadMoreSessions ] [ text "Load More..." ] ]
+
+    else
+        text ""
```
AkkarinForest: So if we have exactly 50 sessions available we show the load more button, even though there is nothing else to load? Maybe that's fine, I know this loading more sessions is tricky and it might be impossible to get do it correctly right now. Just wanted to notice that, so we remember to fix it if this is simplified in the future.

app/javascript/elm/tests/SessionsListTests.elm
```diff
@@ -0,0 +1,201 @@
...
+                        { sessions = [ sessionWithId id, sessionWithId (id + 1) ], selectedSessionId = Just (id + 1) }
+
+                    expected =
+                        { model | selectedSessionId = Nothing }
+                in
+                model
+                    |> update (ToggleSessionSelection (id + 1))
+                    |> Tuple.first
+                    |> Expect.equal expected
+        , fuzz int "when another session was selected ToggleSessionSelection selects the new one" <|
```
gitjul: I don't see that happening, I'm seeing a "You can't select multiple sessions" message

gitjul: I guess because it wasn't changed on the js side?

gitjul: perhaps it would be nice to just get rid of that part of code from js, the new behaviour makes much more sense

app/javascript/elm/src/SessionsList.elm
3v0k4: We are gonna remove that when moving to the new design so I didn't want to overthink the css.

app/javascript/elm/src/SessionsList.elm
3v0k4: Yeah, this sucks. I just brought over what we have in js for now

app/javascript/elm/tests/SessionsListTests.elm
3v0k4: Were you running that on your machine? If so I guess you have the old partials cached

app/javascript/elm/src/SessionsList.elm
3v0k4: Will use that, thanks!

app/javascript/elm/src/SessionsList.elm
3v0k4: Nice comment. I had to check to be sure but it works because the requests for the sessions is fired after the init. The selected one and the new sessions come in through the port

app/javascript/elm/tests/SessionsListTests.elm
```diff
@@ -0,0 +1,201 @@
...
+import Expect
+import Fuzz exposing (int, intRange, string)
+import Html
+import Json.Encode as Encode
+import Ports
+import SessionsList exposing (..)
+import Test exposing (..)
+import Test.Html.Query as Query
+import Test.Html.Selector as Slc
+import TimeRange
```
AkkarinForest: do you need Encode, Ports, TimeRange here?

app/javascript/elm/tests/SessionsListTests.elm
```diff
@@ -0,0 +1,201 @@
...
+                    |> Query.findAll [ Slc.all [ Slc.tag "input", Slc.checked True ] ]
+                    |> Query.count (Expect.equal 1)
+        , fuzz (intRange 1 10) "with modulo 50 sessions in the model the load more button is shown" <|
+            \times ->
+                { sessions = List.repeat (50 * times) session, selectedSessionId = Nothing }
+                    |> view
+                    |> Query.fromHtml
+                    |> Query.contains [ Html.text "Load More..." ]
+        , test "with 0 sessions in the model the load more button is not shown" <|
+            \times ->
```
AkkarinForest: Unused variable `times`

app/javascript/elm/tests/SessionsListTests.elm
```diff
@@ -0,0 +1,201 @@
...
+                    newSessions =
+                        [ selectedSession, unselectedSession ]
+                in
+                model
+                    |> update (UpdateSessions newSessions)
+                    |> Tuple.first
+                    |> .selectedSessionId
+                    |> Expect.equal (Just id)
+        , fuzz int "LoadMoreSessions delegates to javascript" <|
+            \id ->
```
AkkarinForest: Unused variable `id`

app/javascript/elm/tests/SessionsListTests.elm
gitjul: that's right.

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -18,6 +19,7 @@ export const SessionsListCtrl = (
 ) => {
   let sessions;
   let singleSession;
+  let elmApp;
   const CANNOT_SELECT_MULTIPLE_SESSIONS = "You can't select multiple sessions";
```
gitjul: is this still used?

app/javascript/elm/tests/SessionsListTests.elm
3v0k4: Just so you know, until we move out of what we have we always need to do what's documented here https://github.com/HabitatMap/AirCasting#assets-version-bumping

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
3v0k4: Unfortunately yes

PR #143, opened 2019-03-26
Location filter

AkkarinForest:
New location filter, fully working for the mobile tab. On fixed tab I still left the old one parallel to the new one, so we can have it as a guide when implementing the checkboxes. When that's done we should remove it from the HTML and clean js from the function that will no longer be used - cc Jul :)

Also, I'm wondering if changing `location: {address: "", indoorOnly: false, streaming: true}` into separate params is a good idea. On one hand, it makes the code cleaner and makes more sense since `indoorOnly`, `streaming` has nothing to do with the location filter. But the downside is that it breaks the old links that user might have saved. And supporting them might add a lot of work and lines of code. Of course I'll ask Micheal if he thinks that it's necessary, but still... Any thoughts?


app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -59,7 +59,9 @@ export const FixedSessionsMapCtrl = (
 
     const defaults = {
       sensorId,
-      location: {address: "", indoorOnly: false, streaming: true},
+      location: '',
+      indoorOnly: false,
+      streaming: true,
```
AkkarinForest: of course, this breaks the checkboxes for now, we need are planning to rewrite them next.

app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
```diff
@@ -141,6 +131,14 @@ export const MobileSessionsMapCtrl = (
         sessionsUtils.updateCrowdMapLayer($scope.sessions.allSessionIds());
       });
 
+      elmApp.ports.findLocation.subscribe(location => {
+        FiltersUtils.findLocation(location, params, map);
+      });
+
+      map.onPanOrZoom(() => {
+        FiltersUtils.clearLocation(elmApp, params);
```
3v0k4: Have you thought keeping the port action on this side? Passing the entire elm app to a util function seems like delegating too much responsibility

app/javascript/angular/code/filtersUtils.js
```diff
@@ -95,3 +95,13 @@ const updateTooltipContent = (link, tooltip) => {
     window.setTimeout(tooltip.hide, 1000);
   });
 };
+
```
3v0k4: I guess all of this file is untested, it could be a good exercise to see if there's any pure testable function to be extracted (in the same file or not) and test them

app/javascript/angular/tests/filtersUtils.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+  sinon.assert.called(send);
+
+  t.end();
+});
+
+test("clearLocation changes location in the params to empty string", t => {
+  const update = sinon.spy();
+  const params = { update };
+  const elmApp = { ports: { locationCleared: { send: () => {}}}};
```
3v0k4: Sometimes testing highlights smells. This is to me a symptom of the fact that `FiltersUtils` is coupled with Elm: it has to navigate `.ports...send`

app/javascript/elm/src/FixedSessionFilters.elm
```diff
@@ -62,6 +69,12 @@ type Msg
 update : Msg -> Model -> ( Model, Cmd Msg )
 update msg model =
     case msg of
+        UpdateLocationInput newValue ->
+            ( { model | location = newValue }, Cmd.none )
```
3v0k4: This is picky so feel free to totally disregard. When updating values in a record I'd use `newKEY`. In this case `newLocation`

app/javascript/elm/src/FixedSessionFilters.elm
```diff
@@ -101,7 +114,8 @@ updateLabels msg model toSubCmd mapper updateModel =
 view : Model -> Html Msg
 view model =
     div []
-        [ Html.map ProfileLabels <| LabelsInput.view model.profiles "Profile Names" "profiles-search"
+        [ LocationFilter.view model.location UpdateLocationInput SubmitLocation
```
3v0k4: Have you considered keeping that in this file? In my experience with Elm it's better to move to a new file only if there's a strong argument. 

I know this is gonna be used by mobile too but we have duplication in any case. We will hopefully solve that when merging the two

app/javascript/elm/tests/MobileSessionsFiltersTests.elm
```diff
@@ -15,6 +15,54 @@ import Test.Html.Selector as Slc
...
+                    |> update (UpdateLocationInput locationValue)
+                    |> Tuple.first
+                    |> view
+                    |> Query.fromHtml
+                    |> Query.find [ Slc.id "location-filter" ]
+                    |> Query.has [ Slc.attribute <| Attr.value locationValue ]
+        , test "when Enter key is pressed SubmitLocation is triggered" <|
+            \_ ->
+                let
+                    simulatedEnterKeydown : Encode.Value
```
3v0k4: Have you considered another name here? This is an encoded value, `simulated` seems not correct

app/javascript/elm/src/LocationFilter.elm
```diff
@@ -0,0 +1,33 @@
...
+onEnter msg =
+    let
+        isEnter code =
+            if code == 13 then
+                Decode.succeed msg
+
+            else
+                Decode.fail "not ENTER"
+    in
+    Events.on "keydown" (Decode.andThen isEnter Events.keyCode)
```
3v0k4: Cool! `andThen` is a monadic operation, you are using monads feel free to brag!

3v0k4:
As regards the links, I'd err on the side of breaking. It's not a good thing to do but the way they are built makes it difficult to do things compatible with the old format. Let's have a good format for the new big deploy and be good citizens from then on.

PR #146, opened 2019-03-29
Parameter filter

AkkarinForest:
I still need to apply this to fixed, but I wanted to get your opinion first, cause it's already a bit complex.

This is the version with popups. The one with dropdown is very the same in logic, just has different views, we can deploy it first if we are not yet able to use this one. I'll push it on a separate branch.

**UPDATE**
I extracted the popup to a separate module and applied the changes to the fixed tab. But I couldn't figure out an easy way to move `isPopupExtended` flag with that module, unless not without complex abstractions. I think I'll move to sensors filter now and then see if something will be clearer then.

### DIFF
--- a/app/javascript/elm/src/MobileSessionsFilters.elm
+++ b/app/javascript/elm/src/MobileSessionsFilters.elm
@@ -1,13 +1,14 @@
-module MobileSessionsFilters exposing (Msg(..), Popups(..), defaultModel, update, view, viewPopup)
+module MobileSessionsFilters exposing (Msg(..), defaultModel, update, view)
 
 import Browser
 import Browser.Events
-import Html exposing (Html, button, div, h4, input, label, li, option, p, select, span, text, ul)
+import Html exposing (Html, button, div, h4, input, label, p, span, text)
 import Html.Attributes as Attr
 import Html.Events as Events
 import Json.Decode as Decode
 import Json.Encode as Encode
 import LabelsInput
+import Popup
 import Ports
 import TimeRange exposing (TimeRange)
 
@@ -17,9 +18,9 @@ import TimeRange exposing (TimeRange)
 
 
 type alias Model =
-    { popup : Popups
+    { popup : Popup.Popup
     , isPopupExtended : Bool
-    , parameters : Items
+    , parameters : Popup.Items
     , selectedParameter : String
     , location : String
     , tags : LabelsInput.Model
@@ -30,18 +31,9 @@ type alias Model =
     }
 
 
-type Popups
-    = SelectFromItems Items
-    | None
-
-
-type alias Items =
-    { main : List String, other : Maybe (List String) }
-
-
 defaultModel : Model
 defaultModel =
-    { popup = None
+    { popup = Popup.None
     , isPopupExtended = False
     , parameters =
         { main = [ "Particulate Matter", "Humidity", "Temperature", "Sound Levels" ]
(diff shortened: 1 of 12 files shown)


app/javascript/elm/src/MobileSessionsFilters.elm
```diff
@@ -142,6 +195,68 @@ view model =
...
+viewPopup onSelect isPopupExtended items =
+    let
+        numberOfMainItems =
+            4
+
+        mainItems =
+            List.take numberOfMainItems items
+
+        moreItems =
+            List.drop numberOfMainItems items
```
AkkarinForest: is there an easier way to split a list into two parts on a given index?

app/javascript/elm/src/MobileSessionsFilters.elm
```diff
@@ -142,6 +195,68 @@ view model =
...
+selectableItems : List String -> (String -> Msg) -> Html Msg
+selectableItems items onSelect =
+    items
+        |> List.map (\item -> li [] [ button [ Events.onClick (onSelect item) ] [ text item ] ])
+        |> ul []
+
+
+clickWithoutDefault : Msg -> Html.Attribute Msg
+clickWithoutDefault msg =
+    Events.custom "click" (Decode.map preventDefault (Decode.succeed msg))
```
AkkarinForest: This overrides the default onClick which is defined in subscriptions ` Browser.Events.onClick (Decode.succeed ClosePopup)` - this way clicking anywhere on the page closes the popup, except for places where we want to override that

app/javascript/elm/tests/MobileSessionsFiltersTests.elm
```diff
@@ -15,6 +15,109 @@ import Test.Html.Selector as Slc
...
+                    |> Query.findAll [ Slc.tag "li" ]
+                    |> Query.count (Expect.equal 5)
+        , test "clicking on an item executed select function" <|
+            \_ ->
+                [ "item" ]
+                    |> viewPopup SelectParameter False
+                    |> Query.fromHtml
+                    |> Query.find [ Slc.tag "button", Slc.containing [ Slc.text "item" ] ]
+                    |> Event.simulate Event.click
+                    |> Event.expect (SelectParameter "item")
```
AkkarinForest: Can I somehow create a fake select function? I had pass `SelectParameter ` cause `viewPopup` expects `(String -> Msg)`. But it should be able to accept any function of this type, so it's not cool that I had to use a specific one, right?

app/javascript/elm/src/MobileSessionsFilters.elm
```diff
@@ -129,7 +175,14 @@ updateLabels msg model toSubCmd mapper updateModel =
...
 view model =
     div []
-        [ viewLocation model.location
+        [ viewParameterFilter model.selectedParameter
+        , case model.popup of
+            SelectFromItems items ->
+                viewPopup SelectParameter model.isPopupExtended items
+
+            None ->
+                div [] []
```
3v0k4: Maybe here you want `text ""`?

app/javascript/elm/src/MobileSessionsFilters.elm
3v0k4: Would this work? https://package.elm-lang.org/packages/elm-community/list-extra/latest/List-Extra#splitAt

app/javascript/elm/tests/MobileSessionsFiltersTests.elm
3v0k4: I guess you could change the type to `(String -> msg)` in the function and pass from the test anything. Not sure though

PR #147, opened 2019-03-29
Refresh time range filter

AkkarinForest:


app/javascript/elm/tests/TimeRangeTests.elm
```diff
@@ -4,17 +4,24 @@ import Expect
...
 import Test exposing (..)
+import Test.Html.Event as Event
 import Test.Html.Query as Query
 import Test.Html.Selector as Slc
 import TimeRange
 
 
+type MsgStub
+    = MsgStub
+
```
AkkarinForest: Is that a good idea for stubbing a message that I need on line 52?

AkkarinForest:
In the end, we decided with Pina that it's easier to deploy it only to experimental for now. So I cherry-picked the commits there and I'm deleting this PR.

PR #149, opened 2019-04-02
New design with Elm on top

3v0k4:
Put Elm on top of Angular. In other words, put Elm in charge and delegate to old code in Angular. That is the opposite of what we've been doing: Angular on top delegating to multiple Elm apps

Also, this introduces the new design


app/javascript/packs/application.js
```diff
@@ -7,38 +7,43 @@
...
+  require('../angular/libs/jquery_ui/jquery.ui.core');
+  require('../angular/libs/jquery_ui/jquery.ui.widget');
+  require('../angular/libs/jquery_ui/jquery.ui.mouse');
+  require('../angular/libs/jquery_ui/jquery.ui.position');
+  require('../angular/libs/jquery_ui/jquery.ui.slider');
+  require('../angular/libs/jquery_ui/jquery.ui.autocomplete');
+  require('../angular/libs/jquery_ui/jquery.ui.dialog');
+  require('../angular/libs/jquery_ui/jquery.ui.daterangepicker');
+  require('angular');
+  require('angular-route');
```
3v0k4: remove

app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
```diff
@@ -113,7 +103,7 @@ export const MobileSessionsMapCtrl = (
       };
 
       const flags = {
-        parametersList: buildAvailableParameters(sensorsList),
```
AkkarinForest: remove me

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -140,9 +141,7 @@ export const SessionsListCtrl = (
 
   if (process.env.NODE_ENV !== 'test') {
     angular.element(document).ready(() => {
-      const node = document.getElementById('sessions-bottom-elm');
       const flags = $scope.sessions.get().map(formatSessionForList).map(formatSessionForElm);
```
AkkarinForest: do we need these flags?

AkkarinForest:
This is amazing. It will be so much easier to add new stuff in elm now. Thanks for taking care of that, it looks like it was very difficult.

I noticed a few things, you probably don't want to fix them in this PR, but I'll list what I noticed, maybe you can create a separate tickets if these are bigger problems, or maybe they will be fixed as we implement more elm logic, so we just have to remember to check if it works before going to production:
- selected tags lost their styling
- moving the resolution slider doesn't work when the crowd map is not selected. we should probably hide it then
- there is a copy link button right below the timerange filter and there is a separate placeholder at the bottom of the page - we need to merge that
- we need to fix the styling for home, about, donate pages
- there's something wrong with drawing the markers - only one is drawn and it has some black halo that shouldn't be there
- there are no buttons to zoom in and out on the map
- user can move the map when the spinner is shown - this is different than what we had before - do we want that change?

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
3v0k4: Nope! Thanks

3v0k4:
@AkkarinForest thanks a lot for the CR, it helped spotting a lot of things.

When it comes to the list you've mentioned:

FIXED selected tags lost their styling

ASKED PINA moving the resolution slider doesn't work when the crowd map is not selected. we should probably hide it then

ASKED PINA we need to fix the styling for home, about, donate pages

ASKED PINA user can move the map when the spinner is shown - this is different than what we had before - do we want that change?

FIXED there is a copy link button right below the timerange filter and there is a separate placeholder at the bottom of the page - we need to merge that

LET'S DISCUSS THIS there's something wrong with drawing the markers - only one is drawn and it has some black halo that shouldn't be there

WILL CREATE TICKET MYSELF there are no buttons to zoom in and out on the map

PR #150, opened 2019-04-03
Export sessions in Elm

3v0k4:
Move the feature from Angular to Elm


app/javascript/elm/tests/MainTests.elm
```diff
@@ -468,6 +468,28 @@ viewTests =
...
+            \id ->
+                let
+                    expected =
+                        exportPath ++ "?session_ids[]=" ++ String.fromInt id ++ "&session_ids[]=" ++ String.fromInt (id + 1)
+                in
+                { defaultModel | sessions = [ sessionWithId id, sessionWithId (id + 1) ] }
+                    |> view
+                    |> Query.fromHtml
+                    |> Query.find [ Slc.containing [ Slc.text "export sessions" ] ]
+                    |> Query.has [ Slc.attribute <| Attr.href expected ]
```
AkkarinForest: what happens with 0 sessions in the model?

3v0k4: It explodes because Rails doesn't have a default ;)

Added some info here https://llp.kanbanery.com/projects/8040/board/tasks/2365429 If that's fine I'd leave this to that ticket

PR #151, opened 2019-04-04
Sensor filter

AkkarinForest:
The expected behaviour is:
- when opening parameter filter user sees 4 default parameters and has a possibility to see more parameters
- other parameters are ordered alphabetically
- when selecting a parameter a sensor with most sessions is automatically selected
- when opening sensors filter user sees all sensors for given parameter sorted alphabetically

There is still a lot of js left. I was hoping that I'll be able to remove staff from _sensors.js, but it turns out that there are many placed in the js code that use selected `sensor_name` instead of selected `sensorId`. And it seemed like to much to change all of them in this PR. But it's that we should do in the near future. Cause right now I need to create a sensors list both in js and elm. The old one in js is used to find the selected sensor in:
```   
 selected: function() {
        return this.sensors[params.get('data').sensorId || DEFAULT_SENSOR_ID];
 }
```


app/javascript/elm/src/Sensors.elm
```diff
@@ -0,0 +1,154 @@
...
+            parameterSensorPairs
+                |> List.filter (\pair -> pair.id_ == sensorId)
+                |> List.head
+    in
+    case maybePair of
+        Just pair ->
+            pair.label
+
+        Nothing ->
+            ""
```
AkkarinForest: This should never happen, but still, should I have some error logged here?

3v0k4:
> when opening sensors filter user sees all sensors for given parameter sorted alphabetically
Shouldn't this have a few hardcoded on top? For example PM has the AB2 on top

AkkarinForest:
Well yes, but that's coming in the next version :D I listed what is expected behavior in this version in the description.

app/javascript/elm/src/Sensors.elm
```diff
@@ -0,0 +1,154 @@
...
+
+
+allParameters : List ParameterSensorPair -> List String
+allParameters pairs =
+    List.map (\pair -> pair.parameter) pairs
+        |> Set.fromList
+        |> Set.toList
+
+
+allParametersWithPrioritization : List ParameterSensorPair -> { main : List String, other : Maybe (List String) }
```
3v0k4: I see you are using "prioritization" and "main". Maybe you could introduce only one term for that concept

app/javascript/elm/src/Sensors.elm
```diff
@@ -0,0 +1,154 @@
...
+
+        maybeOtherParameters =
+            if List.isEmpty otherParameters then
+                Nothing
+
+            else
+                Just otherParameters
+    in
+    { main = prioritizeParameters
+    , other = maybeOtherParameters
```
3v0k4: Have you considered `others`?

app/javascript/elm/src/Sensors.elm
3v0k4: Is `Nothing` the same as `[]` for `other`? Maybe you don't need the maybe here?

app/javascript/elm/src/Sensors.elm
```diff
@@ -0,0 +1,154 @@
...
+                    pairs
+                        |> List.filter (\pair -> pair.label == key)
+                        |> List.head
+            in
+            case maybeSensorPair of
+                Just pair ->
+                    pair.id_
+
+                Nothing ->
+                    "particulate matter-airbeam2-pm2.5 (µg/m³)"
```
3v0k4: Have you considered using `map` and `withDefault`? I'm pretty sure you can flat this fnc a lot

app/javascript/elm/src/Sensors.elm
```diff
@@ -0,0 +1,154 @@
...
+type alias Decodable =
+    { sensor : String
+    , parameter : String
+    , unit : String
+    , session_count : Int
+    }
+
+
+labelsForParameterInId : List ParameterSensorPair -> String -> List String
+labelsForParameterInId pairs sensorId =
```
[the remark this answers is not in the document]
AkkarinForest: but it's actually sensors labels for a given parameter
but I added `InId` in the name to indicate that this function actually expects a sensorId and not a parameter

app/javascript/elm/src/Sensors.elm
```diff
@@ -0,0 +1,154 @@
...
+
+        Nothing ->
+            ""
+
+
+allParameters : List ParameterSensorPair -> List String
+allParameters pairs =
+    List.map (\pair -> pair.parameter) pairs
+        |> Set.fromList
+        |> Set.toList
```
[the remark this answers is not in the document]
AkkarinForest: yes, that's the only way in elm I found

app/javascript/elm/src/Sensors.elm
AkkarinForest: not really, the popup that shows a list of parameters or sensors uses that maybe to decide it if should have a "show more" button

app/javascript/angular/tests/_sensors.test.js
```diff
@@ -211,7 +22,7 @@ test('selected with no sensor id in the url returns the default sensor with adde
 
   const expected = {
     ...defaultSensor,
-    id: 'Particulate Matter-airbeam2-pm2.5 (µg/m³)',
+    id: 'particulate matter-airbeam2-pm2.5 (µg/m³)',
```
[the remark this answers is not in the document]
AkkarinForest: Pina decided on that with Micheal. But it's a good idea to confirm during the call, I'll do that.
The side effect is that old links with parameters capitalized will not work anymore, but since we are changing bigger things in the API, I guess we can accept that.

app/javascript/elm/src/Sensors.elm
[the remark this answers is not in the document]
3v0k4: So maybe `byId`?

PR #152, opened 2019-04-04
Toggle session selection

3v0k4:
* Refactor: toggling session (brought more stuff in Elm)
* Feature: deselect session when dismissing graph (see "X" in img)

Screen Shot 2019-04-04 at 17 54 15

### DIFF
--- a/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
+++ b/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
@@ -100,11 +100,11 @@ export const SessionsListCtrl = (
     var session = sessions.find(sessionId);
     if(sessions.isSelected(session)) {
       params.update({selectedSessionIds: []});
-      session.$selected = false;
+      elmApp.ports.toggleSessionSelection.send(null);
     } else if($scope.canSelectSession(sessionId)) {
       params.update({selectedSessionIds: [sessionId]});
       $scope.markerSelected.set(markerSelected);
-      session.$selected = true;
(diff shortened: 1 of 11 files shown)


app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -100,11 +100,11 @@ export const SessionsListCtrl = (
     var session = sessions.find(sessionId);
     if(sessions.isSelected(session)) {
       params.update({selectedSessionIds: []});
-      session.$selected = false;
+      elmApp.ports.toggleSessionSelection.send(null);
     } else if($scope.canSelectSession(sessionId)) {
       params.update({selectedSessionIds: [sessionId]});
       $scope.markerSelected.set(markerSelected);
-      session.$selected = true;
+      elmApp.ports.toggleSessionSelection.send(sessionId);
```
AkkarinForest: We should investigate if we can get rid of the watch on line 111 at some point. As far as I can see it's responsible for calling `sessions.deselectSession` or `sessions.selecetSession` so maybe you can just call it from the `if else` directly?
That probably depends on how many other places mutate the `selectedSessionIds`, but I guess it shouldn't be mutated anywhere else so we can try to clean that up. It depends on how we will handle the use case when user change the sensors while having a selected session...

3v0k4: Tried to remove but breaks some stuff. Maybe next round. Thanks for the comment!

3v0k4:
Great catch on the refresh with selected session. While fixing I removed some more dead code!

PR #154, opened 2019-04-05
Show priority sensors for priority parameters

AkkarinForest:
Now the sensors popup for prioritized parameters has a section with prioritized sensors.

### DIFF
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -131,7 +131,7 @@ type Msg
     | UpdateTimeRange Encode.Value
     | RefreshTimeRange
     | ShowCopyLinkTooltip
-    | ShowExpandableSelectFromPopup
+    | ShowExpandableSelectFromPopup Popup.Items String
     | ShowSelectFormPopup
     | SelectSensorId String
     | ClosePopup
@@ -181,8 +181,8 @@ update msg model =
         ShowCopyLinkTooltip ->
             ( model, Ports.showCopyLinkTooltip () )
 
-        ShowExpandableSelectFromPopup ->
-            ( { model | popup = Popup.ExpandableSelectFrom (Sensor.allParametersWithPrioritization model.sensors) }, Cmd.none )
+        ShowExpandableSelectFromPopup items itemType ->
+            ( { model | popup = Popup.ExpandableSelectFrom items itemType, isPopupExtended = False }, Cmd.none )
 
         ShowSelectFormPopup ->
             ( { model | popup = Popup.SelectFrom (Sensor.sensorLabelsForParameterInId model.sensors model.selectedSensorId) }, Cmd.none )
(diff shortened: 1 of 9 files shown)


app/javascript/elm/src/Popup.elm
```diff
@@ -7,36 +7,34 @@ import Json.Decode as Decode
...
-
-                    Nothing ->
-                        text ""
+                , if List.isEmpty items.others then
+                    text ""
+
+                  else if isPopupExtended then
+                    div []
+                        [ selectableItems items.others onSelect
+                        , togglePopupStateButton ("less " ++ itemType) toggle
```
3v0k4: Do we need to allow users to make the list short again? Maybe we could remove that feature?

app/javascript/elm/src/Main.elm
```diff
@@ -177,8 +177,8 @@ update msg model =
         ShowCopyLinkTooltip ->
             ( model, Ports.showCopyLinkTooltip () )
 
-        ShowExpandableSelectFromPopup ->
-            ( { model | popup = Popup.ExpandableSelectFrom (Sensor.allParametersWithPrioritization model.sensors) }, Cmd.none )
+        ShowExpandableSelectFromPopup items itemType ->
+            ( { model | popup = Popup.ExpandableSelectFrom items itemType, isPopupExtended = False }, Cmd.none )
```
3v0k4: Maybe we could have a `popupState : { isExtended : Bool, items : ...}` so that it's clear that both are popup stuff?

app/javascript/elm/src/Main.elm
```diff
@@ -177,8 +177,8 @@ update msg model =
         ShowCopyLinkTooltip ->
             ( model, Ports.showCopyLinkTooltip () )
 
-        ShowExpandableSelectFromPopup ->
-            ( { model | popup = Popup.ExpandableSelectFrom (Sensor.allParametersWithPrioritization model.sensors) }, Cmd.none )
+        ShowExpandableSelectFromPopup items itemType ->
```
3v0k4: Not sure I understand this well but what do you think of renaming `ShowExpandableSelectFromPopup` to `ShowPopup`?

app/javascript/elm/src/Popup.elm
AkkarinForest: Pina would like to keep it.
Would removing it give us some significant benefits?

app/javascript/elm/src/Popup.elm
```diff
@@ -7,36 +7,34 @@ import Json.Decode as Decode
...
-    , others : Maybe (List String)
+    , others : List String
     }
 
 
 view : msg -> (String -> msg) -> Bool -> Popup -> Html msg
 view toggle onSelect isPopupExtended popup =
     case popup of
-        ExpandableSelectFrom items ->
+        ExpandableSelectFrom items itemType ->
```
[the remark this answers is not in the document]
AkkarinForest: These are different concepts. `ExpandableSelectFrom` has two states: expanded and not expanded. `SelectFrom` is just a list of items.

AkkarinForest:
Sorry it was difficult to understand. I refactored it so maybe not it will be clearer. But it's still a bit complex. The idea is that there are 3 ways the popups may look :
- shows main items and a "show more" button - aka not expanded
- shows main and other items - aka expanded
- just shows the list of items, no buttons - this is for sensors that don't have any "prioritized" elements

app/javascript/elm/src/Main.elm
AkkarinForest: I'm not sure about that. The popup can have two states (shown or hidden named `SelectFrom`, `None`), that is already in the model under `popup`.  Then only `SelectFrom` is interested in `isExtender` and `items`.
But in general, I think it's a good idea to connect them, just didn't have time to figure it out today.

PR #155, opened 2019-04-05
Single session info

3v0k4:
Sorry for the size of the PR. I guess you can review backend and frontend separately.

This PR adds the session details to the left of the graph when a session is selected. See below

Screen Shot 2019-04-10 at 12 39 26

### DIFF
--- a/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
+++ b/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
@@ -132,4 +132,4 @@ export const SessionsListCtrl = (
 }
 
 const formatSessionForElm = s =>
-  ({ ...s , shortTypes: s.shortTypes.map(({ name, type }) => ({ name, type_: type })) });
+  ({ ...s , timeRange: s.timeframe, shortTypes: s.shortTypes.map(({ name, type }) => ({ name, type_: type })) });
--- a/app/javascript/elm/src/Data/SelectedSession.elm
+++ b/app/javascript/elm/src/Data/SelectedSession.elm
@@ -1,0 +1,78 @@
+module Data.SelectedSession exposing (SelectedSession, decoder, fetch, sensorNameFromId, view)
+
+import Data.Page exposing (Page(..))
+import Html exposing (Html, div, p, text)
+import Html.Attributes as Attr
+import Http
+import Json.Decode as Decode exposing (Decoder(..))
+import RemoteData exposing (RemoteData(..), WebData)
+
+
+type alias SelectedSession =
+    { title : String
+    , username : String
+    , sensorName : String
+    , average : Float
+    , min : Float
+    , max : Float
+    , timeRange : String
+    , measurements : List Float
+    , id : Int
+    }
+
+
+decoder : Decoder SelectedSession
+decoder =
+    Decode.map7 toSelectedSession
+        (Decode.field "title" Decode.string)
+        (Decode.field "username" Decode.string)
+        (Decode.field "sensor_name" Decode.string)
+        (Decode.field "average" (Decode.nullable Decode.float) |> Decode.map (Maybe.withDefault -1))
+        (Decode.field "timeRange" Decode.string)
+        (Decode.field "measurements" (Decode.list Decode.float))
+        (Decode.field "id" Decode.int)
+
+
+toSelectedSession : String -> String -> String -> Float -> String -> List Float -> Int -> SelectedSession
+toSelectedSession title username sensorName average timeRange measurements sessionId =
+    { title = title
+    , username = username
+    , sensorName = sensorName
+    , average = average
+    , min = List.minimum measurements |> Maybe.withDefault -1
+    , max = List.maximum measurements |> Maybe.withDefault -1
+    , timeRange = timeRange
+    , measurements = measurements
+    , id = sessionId
+    }
+
+
+sensorNameFromId : String -> String
+sensorNameFromId =
+    String.split "-" >> List.drop 1 >> String.join "-" >> String.split " " >> List.head >> Maybe.withDefault ""
+
+
+fetch : String -> Page -> Int -> (WebData SelectedSession -> msg) -> Cmd msg
+fetch sensorId page id toMsg =
+    Http.get
+        { url =
+            if page == Mobile then
+                "/api/mobile/sessions/" ++ String.fromInt id ++ ".json?sensor_name=" ++ sensorNameFromId sensorId
+
+            else
+                "/api/fixed/sessions/" ++ String.fromInt id ++ ".json?sensor_name=" ++ sensorNameFromId sensorId
+        , expect = Http.expectJson (RemoteData.fromResult >> toMsg) decoder
+        }
+
+
+view : SelectedSession -> Html msg
+view session =
+    div []
+        [ p [ Attr.class "single-session-TODO" ] [ text session.title ]
+        , p [ Attr.class "single-session-TODO" ] [ text session.username ]
+        , p [ Attr.class "single-session-TODO" ] [ text session.sensorName ]
+        , p [ Attr.class "single-session-TODO" ] [ text <| String.fromFloat session.min ]
+        , p [ Attr.class "single-session-TODO" ] [ text <| String.fromFloat session.max ]
+        , p [ Attr.class "single-session-TODO" ] [ text <| String.fromFloat session.average ]
+        , p [ Attr.class "single-session-TODO" ] [ text session.timeRange ]
+        ]
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -1,19 +1,22 @@
-module Main exposing (Msg(..), Page(..), defaultModel, exportPath, update, view)
+module Main exposing (Msg(..), defaultModel, exportPath, update, view)
 
 import Browser exposing (..)
 import Browser.Events
 import Browser.Navigation
+import Data.Page exposing (Page(..))
+import Data.SelectedSession as SelectedSession exposing (SelectedSession)
 import Data.Session exposing (..)
 import Html exposing (Html, a, button, dd, div, dl, dt, form, h2, h3, img, input, label, li, main_, nav, p, span, text, ul)
 import Html.Attributes as Attr
 import Html.Attributes.Aria exposing (..)
 import Html.Events as Events
-import Json.Decode as Decode
+import Json.Decode as Decode exposing (Decoder(..))
 import Json.Encode as Encode
 import LabelsInput
 import Maybe exposing (..)
 import Popup
 import Ports
+import RemoteData exposing (RemoteData(..), WebData)
 import Sensor exposing (Sensor)
 import String exposing (fromInt)
 import TimeRange exposing (TimeRange)
@@ -29,16 +32,11 @@ exportPath =
 ---- MODEL ----
 
 
-type Page
-    = Fixed
-    | Mobile
-
-
 type alias Model =
     { page : Page
     , key : Maybe Browser.Navigation.Key
     , sessions : List Session
-    , selectedSessionId : Maybe Int
+    , selectedSession : WebData SelectedSession
     , isHttping : Bool
     , popup : Popup.Popup
     , isPopupExtended : Bool
@@ -59,7 +57,6 @@ defaultModel =
     { page = Mobile
     , key = Nothing
     , sessions = []
-    , selectedSessionId = Nothing
     , isHttping = False
     , popup = Popup.None
     , isPopupExtended = False
@@ -72,6 +69,7 @@ defaultModel =
     , crowdMapResolution = 25
     , timeRange = TimeRange.defaultTimeRange
     , isIndoor = False
+    , selectedSession = NotAsked
     }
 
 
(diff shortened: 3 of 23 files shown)


app/javascript/elm/src/Main.elm
```diff
@@ -354,16 +426,44 @@ view model =
                                             ]
 
                                         Just _ ->
-                                            [ div [ Attr.class "single-session-container" ]
-                                                [ div [ Attr.class "single-session-info" ]
-                                                    [ p [ Attr.class "single-session-owner" ] [ text "NYCEJA" ] ]
-                                                , div
-                                                    [ Attr.class "single-session-graph", Attr.id "graph-box" ]
-                                                    [ div [ Attr.id "graph" ] []
+                                            case model.selectedSession of
```
3v0k4: use remote data and remove double nesting?

spec/services/api/form_spec.rb
```diff
@@ -99,14 +85,30 @@ class Struct < Dry::Struct
...
-      form = Api::Form.new(json: json, schema: schema, struct: struct)
+      form = Api::ParamsForm.new(params: params, schema: schema, struct: struct)
       form.add_error("added error")
 
       expect(form.errors).to eq([ "name is missing", "added error" ])
     end
   end
 end
+
+describe Api::JsonForm do
```
3v0k4: The rest is already tested in `ParamsForm`

spec/controllers/api/mobile/sessions_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+        "timeRange": "01.10.2000 02:03 - 04.11.2001 05:06",
+        "id": session.id,
+      }
+      expect(json_response).to eq(expected)
+    end
+  end
+
+  private
+
+  def create_user!(username:)
```
3v0k4: I'll extract the factories in a future PR

app/javascript/elm/src/Data/SelectedSession.elm
```diff
@@ -0,0 +1,78 @@
+module Data.SelectedSession exposing (SelectedSession, decoder, fetch, sensorNameFromId, view)
+
+import Data.Page exposing (Page(..))
+import Html exposing (Html, div, p, text)
+import Html.Attributes as Attr
```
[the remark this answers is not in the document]
gitjul: update: did that for other elm files on `experimental`

app/javascript/elm/src/Main.elm
```diff
@@ -249,15 +241,50 @@ update msg model =
...
+                Success selectedSession ->
+                    if selectedSession.id == id then
+                        ( { model | selectedSession = NotAsked }
+                        , Ports.checkedSession { deselected = Just selectedSession.id, selected = Nothing }
+                        )
+
+                    else
+                        ( { model | selectedSession = NotAsked }
+                        , Cmd.batch
+                            [ Ports.checkedSession { deselected = Just selectedSession.id, selected = Just id }
```
gitjul: I'm wondering whether another name wouldn't be better. "checked" is ambiguous, I'd consider using "selected" which is consistent with the other naming (if it makes sense) or sth like "chosen" or "toggled". I feel like for sure I've had this conversation before but don't remember the outcome :P but it may be different this time anyway

app/javascript/elm/src/Main.elm
```diff
@@ -351,29 +378,7 @@ view model =
...
-                                                , div
-                                                    [ Attr.class "single-session-graph", Attr.id "graph-box" ]
-                                                    [ div [ Attr.id "graph" ] []
-                                                    ]
-                                                , div [ Attr.class "single-session-close" ] [ button [ Events.onClick DeselectSession ] [ text "X" ] ]
-                                                ]
-                                            ]
-                                    )
-                                ]
+                                [ div [ Attr.attribute "ng-controller" "SessionsListCtrl" ] (viewSessionsOrSelectedSession model.selectedSession model.sessions) ]
```
gitjul: how cute :)

elm and angular sitting in a tree
k-i-s-s-i-n-g

app/javascript/elm/tests/MainTests.elm
```diff
@@ -525,132 +528,173 @@ sessionWithTitle title =
...
                         exportPath ++ "?session_ids[]=" ++ String.fromInt id
                 in
-                { defaultModel | sessions = [ sessionWithId id ] }
+                { defaultModel | sessions = [ sessionWithId id ], selectedSession = NotAsked }
                     |> view
                     |> Query.fromHtml
                     |> Query.find [ Slc.containing [ Slc.text "export sessions" ] ]
                     |> Query.has [ Slc.attribute <| Attr.href expected ]
-        , fuzz int "with 2 sessions in the model the export link is correctly generated" <|
+        , fuzz int "with no selection and 2 sessions in the model the export link is correctly generated" <|
```
gitjul: is this test needed? or the one with 1 session?

app/javascript/elm/tests/MainTests.elm
```diff
@@ -525,132 +528,173 @@ sessionWithTitle title =
...
-        , fuzz (intRange 1 10) "with modulo 50 sessions in the model the load more button is shown" <|
+        , fuzz (intRange 1 10) "with no selection and modulo 50 sessions in the model the load more button is shown" <|
             \times ->
-                { defaultModel | sessions = List.repeat (50 * times) session, selectedSessionId = Nothing }
+                { defaultModel | sessions = List.repeat (50 * times) session, selectedSession = NotAsked }
                     |> view
                     |> Query.fromHtml
                     |> Query.contains [ Html.text "Load More..." ]
-        , test "with 0 sessions in the model the load more button is not shown" <|
+        , test "with no selection and 0 sessions in the model the load more button is not shown" <|
```
gitjul: is this case is true for 0-49 sessions?

app/javascript/elm/tests/MainTests.elm
```diff
@@ -525,132 +528,173 @@ sessionWithTitle title =
...
-        , fuzz int "with no selections ToggleSessionSelection tells javascript what was selected" <|
+                { defaultModel | selectedSession = Success selectedSession_ }
+                    |> view
+                    |> Query.fromHtml
+                    |> Query.has [ Slc.all expected ]
+        ]
+
+
+updateTests : Test
+updateTests =
```
gitjul: I think I kinda like this way of organizing tests better, will try that out next time

3v0k4:
> I don't have a good understanding of how viewing the sessions worked before, and sorry if this comment is dumb, but you added some nice code that retrieves the sessions that I would think replaces a lot of old code since this functionality is in place. but not much seems to have been removed. it seems weird to me and I'm sure I'm missing something. I don't have anything actionable thing to say in this matter, we can talk offline :)

Not sure I get what code you are referring to as "old". But it's possible I left some endpoint around. I have a stickie to remember about doing that. Thing is it's not clear what we can remove from the API cause it's public. Also, there's some stuff on my wall with all the mobile / web endpoints. I'm available to chat about this if we want to figure out something at the moment instead of deferring.

> I think I'd like to add rubocop at some point, if you like the idea too. this is a general comment related to the fact I'm looking at Ruby code, not this neat PR. I wouldn't necessarily apply any pre-deploy hooks, just to use when needed/wanted for now, so we don't have to fix the whole existing code.

+100

> looking at the network tab, showing a session now makes 3 requests instead of one, 2 of which look exactly the same - don't know if this duplication is intentional or simply ok to happen or a bug. couple requests instead of one isn't bad I think, especially if nothing else is happening, it's only been a problem for me in the past when they were blocking each other and slowing down the load. just an observation.
 
Good catch :) It's done once in Elm and once in Angular. We will hopefully remove that asap when doing more work. It's a tradeoff I accepted to enable smaller iterations while bringing more things in Elm.

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -132,4 +132,4 @@ export const SessionsListCtrl = (
 }
 
 const formatSessionForElm = s =>
-  ({ ...s , shortTypes: s.shortTypes.map(({ name, type }) => ({ name, type_: type })) });
+  ({ ...s , timeRange: s.timeframe, shortTypes: s.shortTypes.map(({ name, type }) => ({ name, type_: type })) });
```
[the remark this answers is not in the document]
3v0k4: Ok. I'll start doing that from now on

app/javascript/elm/src/Data/SelectedSession.elm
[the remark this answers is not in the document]
3v0k4: I'll do that from now on

app/services/api/to_session_hash.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    Success.new(
+      title: session.title,
+      username: session.user.username,
+      sensor_name: session.streams.first.sensor_name,
+      average: session.streams.first.average_value,
+      measurements: session.streams.first.measurements.map(&:value),
+      timeRange: "#{format_datetime(session.start_time_local)} - #{format_datetime(session.end_time_local)}",
+      id: session.id,
+    )
```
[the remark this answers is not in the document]
3v0k4: I thought about that and thought that the added complexity would not justify the move as you said

spec/controllers/api/v2/data/sessions_controller_spec.rb
```diff
@@ -1,14 +1,12 @@
 require 'rails_helper'
 
 describe Api::V2::Data::SessionsController do
-  describe "GET 'last'" do
-    before do
+  describe "GET last" do
+    it 'returns id' do
       FactoryGirl.create(:mobile_session)
```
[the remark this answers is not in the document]
3v0k4: I'm for 2. This is a quote from the authors of the gem which I agree with

> My general advice, though, is to avoid Factory Girl as much as is reasonably possible. Not because it’s bad or unreliable software (Factory Girl is very reliable; we’ve used it successfully since 2008), but because its inherent persistence mechanism is calling #save! on the object, which will always take longer than not persisting data.
(https://thoughtbot.com/blog/speed-up-tests-by-selectively-avoiding-factory-girl)

app/javascript/elm/tests/MainTests.elm
```diff
@@ -706,27 +750,14 @@ updateTests =
                     |> update DeselectSession
                     |> Tuple.second
                     |> Expect.equal expected
-        , fuzz int "with Just ToggleSessionSelectionFromAngular selects the passed id" <|
```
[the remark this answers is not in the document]
3v0k4: Cause I believe there's no way of testing `Cmd`s in Elm and I believe that `Ports` call is together with a `Cmd`

app/javascript/elm/src/Data/SelectedSession.elm
```diff
@@ -0,0 +1,78 @@
...
+
+fetch : String -> Page -> Int -> (WebData SelectedSession -> msg) -> Cmd msg
+fetch sensorId page id toMsg =
+    Http.get
+        { url =
+            if page == Mobile then
+                "/api/mobile/sessions/" ++ String.fromInt id ++ ".json?sensor_name=" ++ sensorNameFromId sensorId
+
+            else
+                "/api/fixed/sessions/" ++ String.fromInt id ++ ".json?sensor_name=" ++ sensorNameFromId sensorId
```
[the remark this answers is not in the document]
3v0k4: Yeah I don't like it much too. Thing is I'm yet not sure how to organize api calls. Maybe we could have one module for "url builders"?

app/javascript/elm/src/Main.elm
3v0k4: Good one! Renamed to `toggleSession`. At some point we may want to have a chat about naming ports. Not sure if it's better the imperative or past tense form (ie toggleSession vs sessionToggled)

app/javascript/elm/src/Main.elm
3v0k4: We should add React and make it a threesome ;)

app/javascript/elm/tests/MainTests.elm
3v0k4: I'm not sure how much I like it. It's risky to forget about exposing (and running) some tests

spec/controllers/api/mobile/sessions_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  def create_measurement!(stream:, sensor_name:, value:)
+    Measurement.create!(
+      time: DateTime.current,
+      latitude: 123,
+      longitude: 123,
+      value: 123,
+      milliseconds: 123,
+      stream: stream
+    )
+  end
```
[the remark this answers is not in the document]
3v0k4: Could you elaborate on this please? I don't get it

app/javascript/elm/tests/MainTests.elm
3v0k4: I think so. But 0 is an edge case cause it's modulo 50 that's why I tested

app/javascript/elm/tests/MainTests.elm
3v0k4: I guess it's a bit of overtesting. The rationale was that in one case there's `?` and stuff and in the other `?...&`. I'm fine with removing if you prefer

3v0k4:
Wow! You spent a lot of time on this. Thanks :) Feel free to do less if it takes too much time!

spec/controllers/api/v2/data/sessions_controller_spec.rb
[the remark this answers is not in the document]
gitjul: esattamente

PR #156, opened 2019-04-09
Disable profile filter for indoor sessions

gitjul:
* disable input
* clear selected profiles

@paulinamaterna 
- [ ] style disabled inputs

### DIFF
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -508,8 +508,8 @@ viewMobileFilters model =
         , viewSensorFilter model.sensors model.selectedSensorId
         , viewLocation model.location model.isIndoor
         , TimeRange.view RefreshTimeRange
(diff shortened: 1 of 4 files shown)


app/javascript/elm/src/Main.elm
```diff
@@ -240,7 +240,13 @@ update msg model =
             ( { model | isHttping = isHttpingNow }, Cmd.none )
 
         ToggleIndoor ->
-            ( { model | isIndoor = not model.isIndoor }, Ports.toggleIndoor (not model.isIndoor) )
+            if model.isIndoor then
+                ( { model | isIndoor = not model.isIndoor }, Ports.toggleIndoor (not model.isIndoor) )
```
3v0k4: Since `isIndoor` must be True I'd put `False` when updating the model. It's less cryptic than `not`. The same for the else branch

gitjul: yeah was considering it. I'll do it directly on experimental

PR #157, opened 2019-04-10
Streaming toggle

gitjul:
Ongoing/ended streaming toggle.

55633575-fbb63000-57bc-11e9-8013-9602fda53caf

### DIFF
--- a/app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
+++ b/app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
@@ -51,7 +51,7 @@ export const FixedSessionsMapCtrl = (
       sensorId,
       location: '',
       isIndoor: false,
-      streaming: true,
+      isStreaming: true,
       tags: "",
       usernames: "",
       timeFrom: oneYearAgo,
@@ -89,6 +89,11 @@ export const FixedSessionsMapCtrl = (
 
       elmApp.ports.toggleIndoor.subscribe(isIndoor => {
         params.update({ data: { isIndoor: isIndoor }});
+        $scope.sessions.fetch();
+      });
+
+      elmApp.ports.toggleStreaming.subscribe(isStreaming => {
+        params.update({ data: { isStreaming: isStreaming }});
         $scope.sessions.fetch();
       });
 
--- a/app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
+++ b/app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
@@ -1,10 +1,18 @@
-import _ from 'underscore';
-import moment from 'moment'
+import _ from "underscore";
+import moment from "moment";
 
-import * as FiltersUtils from '../filtersUtils'
+import * as FiltersUtils from "../filtersUtils";
 
-const endOfToday = moment().utc().endOf('day').format('X');
-const oneYearAgo = moment().utc().startOf('day').subtract(1, 'year').format('X');
+const endOfToday = moment()
+  .utc()
+  .endOf("day")
+  .format("X");
+
+const oneYearAgo = moment()
+  .utc()
+  .startOf("day")
+  .subtract(1, "year")
+  .format("X");
 
 export const FixedSessionsMapCtrl = (
   $scope,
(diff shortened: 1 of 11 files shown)


app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -92,6 +92,11 @@ export const FixedSessionsMapCtrl = (
         $scope.sessions.fetch();
       });
 
+      elmApp.ports.toggleStreaming.subscribe(isStreaming => {
+        params.update({ data: { isStreaming: isStreaming }});
```
3v0k4: `isStreaming: isStreaming` == `isStreaming` if you prefer

app/javascript/elm/tests/MainTests.elm
```diff
@@ -501,6 +501,56 @@ toggleIndoorFilter =
...
+                    |> Query.find [ Slc.attribute <| ariaLabel "active" ]
+                    |> Event.simulate Event.click
+                    |> Event.expect ToggleStreaming
+        , test "when isStreaming is true ToggleStreaming triggers Ports.toggleStreaming with False" <|
+            \_ ->
+                { defaultModel | page = Fixed }
+                    |> update ToggleStreaming
+                    |> Tuple.second
+                    |> Expect.equal (Ports.toggleStreaming False)
+        , test "when isStreaming is false ToggleStreaming triggers Ports.toggleStreaming with False" <|
```
3v0k4: I guess this and the previous test are using the same case (`isStreaming = True`) cause that's the default in `defaultModel`

gitjul: yes, you're totally right, mixup

gitjul:
thanks for reviewing! this is actually not 100% complete, there is one more todo with locking the time range on 1 last hour. forgot to mark this as in progress :)

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
gitjul: is this better?

3v0k4: I like it

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -1,19 +1,6 @@
...
-const endOfToday = moment()
-  .utc()
-  .endOf("day")
-  .format("X");
-
-const oneYearAgo = moment()
-  .utc()
-  .startOf("day")
-  .subtract(1, "year")
-  .format("X");
```
gitjul: this was setting static values once instead of every time we needed to refresh. with endOfToday and oneYearAgo the difference may have not been noticeable but with the 1 hour span it was obvious we need functions

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
gitjul: why do you like it better? what's the +?

app/javascript/angular/code/filtersUtils.js
```diff
@@ -1,5 +1,46 @@
...
+    .format("X");
+
+export const oneHourAgo = () =>
+  moment()
+    .utc()
+    .subtract(1, "hour")
+    .format("X");
+
+export const defaultTimeFrom = isStreaming => {
+  if (isStreaming) {
```
3v0k4: You could also use the ternary
```
const fnc = isStreaming =>
  isStreaming ? a() : b()
```
I'd suggest bubbling up the conditional. Reek for example [reports that as a smell](https://github.com/troessner/reek/blob/master/docs/Boolean-Parameter.md)

app/javascript/angular/code/filtersUtils.js
```diff
@@ -1,5 +1,46 @@
...
+    return oneHourAgo();
+  } else {
+    return oneYearAgo();
+  }
+};
+
+export const defaultTimeTo = isStreaming => {
+  if (isStreaming) {
```
3v0k4: As above :)

app/javascript/elm/src/Main.elm
```diff
@@ -670,7 +686,7 @@ viewMobileFilters model =
         [ viewParameterFilter model.sensors model.selectedSensorId
         , viewSensorFilter model.sensors model.selectedSensorId
         , viewLocation model.location model.isIndoor
-        , TimeRange.view RefreshTimeRange
+        , TimeRange.view RefreshTimeRange False
```
3v0k4: I wonder if creating a custom type instead of the boolean would simplify or not the code. Something like
```
type Streaming
  = Active
  | Dormant
```
so that here you could use
```
 TimeRange.view RefreshTimeRange Dormant
```

app/javascript/elm/tests/MainTests.elm
```diff
@@ -521,6 +535,56 @@ toggleIndoorFilter =
...
+                    |> Query.has [ Slc.attribute <| class "toggle-button--pressed" ]
+        , test "clicking dormant button triggers ToggleStreaming" <|
+            \_ ->
+                { defaultModel | page = Fixed }
+                    |> view
+                    |> Query.fromHtml
+                    |> Query.find [ Slc.attribute <| ariaLabel "dormant" ]
+                    |> Event.simulate Event.click
+                    |> Event.expect ToggleStreaming
+        , test "clicking active button triggers ToggleStreaming" <|
```
3v0k4: Have you considered just keeping one test for triggering `ToggleStreaming`?

app/javascript/elm/src/Main.elm
gitjul: I'll consider changing that, for sure it would be more meaningful, I didn't really like passing the boolean like that because we don't know what it represents until we look for it

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
3v0k4: I guess less code? I don't have a strong opinion

gitjul: hmmm then I guess I didn't get the change, I'll ask offline

PR #160, opened 2019-04-12
Heat

3v0k4:
* fetch heat levels for sensor on app init
* refactor sensor to remove `label` and `id_`. The rationale is that those two can be calculated with the other properties of a `Sensor`. "Caching" them poses the risk of getting out of sync

### DIFF
--- a/app/controllers/api/thresholds_controller.rb
+++ b/app/controllers/api/thresholds_controller.rb
@@ -21,7 +21,7 @@ module Api
     respond_to :json
 
     def show
-      respond_with Stream.thresholds(params[:id], params[:unit_symbol])
+      render json: Stream.thresholds(params[:id], params[:unit_symbol])
     end
   end
 end
--- a/app/javascript/elm/src/Data/HeatMapThresholds.elm
+++ b/app/javascript/elm/src/Data/HeatMapThresholds.elm
@@ -1,0 +1,59 @@
+module Data.HeatMapThresholds exposing (HeatMapThresholds, fetch)
+
+import Http
+import Json.Decode as Decode exposing (Decoder(..))
+import Sensor exposing (Sensor)
+
+
+type alias HeatMapThresholds =
+    { h1 : Int
+    , h2 : Int
+    , h3 : Int
+    , h4 : Int
+    , h5 : Int
+    }
+
+
+fetch : List Sensor -> String -> (Result Http.Error HeatMapThresholds -> msg) -> Maybe (Cmd msg)
+fetch sensors sensorId toCmd =
+    let
+        maybeSensorName =
+            Sensor.nameForSensorId sensorId sensors
+
+        maybeUnit =
+            Sensor.unitForSensorId sensorId sensors
+
+        fetch_ sensorName unit =
+            Http.get
+                { url = "/api/thresholds/" ++ sensorName ++ "?unit_symbol=" ++ unit
+                , expect = Http.expectJson toCmd decoder
+                }
+    in
+    Maybe.map2 fetch_ maybeSensorName maybeUnit
+
+
+maybeIntDecoder : Maybe Int -> Decoder Int
+maybeIntDecoder m =
+    case m of
+        Nothing ->
+            Decode.fail "threshold must be an Int"
+
+        Just i ->
+            Decode.succeed i
+
+
+intAsStringDecoder : Decoder Int
+intAsStringDecoder =
+    Decode.string
+        |> Decode.map String.toInt
+        |> Decode.andThen maybeIntDecoder
+
+
+decoder : Decoder HeatMapThresholds
+decoder =
+    Decode.map5 HeatMapThresholds
+        (Decode.index 0 intAsStringDecoder)
+        (Decode.index 1 intAsStringDecoder)
+        (Decode.index 2 intAsStringDecoder)
+        (Decode.index 3 intAsStringDecoder)
+        (Decode.index 4 intAsStringDecoder)
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -119,13 +119,18 @@ init flags url key =
         , logoNav = flags.logoNav
         , linkIcon = flags.linkIcon
       }
-    , case flags.selectedSessionId of
+    , fetchSelectedSession flags.selectedSessionId flags.selectedSensorId page
+    )
+
+
+fetchSelectedSession : Maybe Int -> String -> Page -> Cmd Msg
+fetchSelectedSession maybeId selectedSensorId page =
+    case maybeId of
         Nothing ->
             Cmd.none
 
         Just id ->
-            SelectedSession.fetch flags.selectedSensorId page id GotSession
-    )
+            SelectedSession.fetch selectedSensorId page id GotSession
 
 
 
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -3,6 +3,7 @@ module Main exposing (Msg(..), defaultModel, exportPath, update, view)
 import Browser exposing (..)
 import Browser.Events
 import Browser.Navigation
+import Data.HeatMapThresholds as HeatMapThresholds exposing (HeatMapThresholds)
 import Data.Page exposing (Page(..))
 import Data.SelectedSession as SelectedSession exposing (SelectedSession)
 import Data.Session exposing (..)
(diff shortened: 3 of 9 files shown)


app/controllers/api/thresholds_controller.rb
```diff
@@ -21,7 +21,7 @@ class ThresholdsController < BaseController
     respond_to :json
 
     def show
-      respond_with Stream.thresholds(params[:id], params[:unit_symbol])
+      render json: Stream.thresholds(params[:id], params[:unit_symbol])
```
3v0k4: This way it always returns json. The alternatives were
- add "accept" json to the client (since this controller is supposed to do json only this seems skippable)
- adding .json in the request url (the route uses a regexp that would consider `.json` as part of the sensor id)

app/javascript/elm/src/Data/HeatMapThresholds.elm
```diff
@@ -0,0 +1,54 @@
...
+    , h2 : Int
+    , h3 : Int
+    , h4 : Int
+    , h5 : Int
+    }
+
+
+fetch : List Sensor -> String -> (Result Http.Error HeatMapThresholds -> msg) -> Cmd msg
+fetch sensors sensorId toCmd =
+    case ( Sensor.nameForSensorId sensors sensorId, Sensor.unitForSensorId sensors sensorId ) of
```
3v0k4: ?

app/javascript/elm/src/Main.elm
```diff
@@ -3,13 +3,15 @@ module Main exposing (Msg(..), defaultModel, exportPath, update, view)
...
 import Browser.Navigation
+import Data.HeatMapThresholds as HeatMapThresholds exposing (HeatMapThresholds)
 import Data.Page exposing (Page(..))
 import Data.SelectedSession as SelectedSession exposing (SelectedSession)
 import Data.Session exposing (..)
 import Html exposing (Html, a, button, dd, div, dl, dt, form, h2, h3, img, input, label, li, main_, nav, p, span, text, ul)
 import Html.Attributes exposing (attribute, autocomplete, checked, class, classList, disabled, for, href, id, max, min, name, placeholder, rel, src, target, type_, value)
 import Html.Attributes.Aria exposing (ariaLabel)
 import Html.Events as Events
+import Http
```
3v0k4: ?

app/javascript/elm/src/Data/HeatMapThresholds.elm
```diff
@@ -0,0 +1,54 @@
+module Data.HeatMapThresholds exposing (HeatMapThresholds, fetch)
+
+import Http
+import Json.Decode as Decode exposing (Decoder(..))
+import Sensor exposing (Sensor)
+
+
+type alias HeatMapThresholds =
```
3v0k4: alias

app/javascript/elm/src/Main.elm
```diff
@@ -119,13 +123,21 @@ init flags url key =
         , logoNav = flags.logoNav
         , linkIcon = flags.linkIcon
       }
-    , case flags.selectedSessionId of
+    , Cmd.batch
+        [ fetchSelectedSession flags.selectedSessionId flags.selectedSensorId page
+        , HeatMapThresholds.fetch (Sensor.decodeSensors flags.sensors) flags.selectedSensorId (RemoteData.fromResult >> UpdateHeatMapThresholds)
```
3v0k4: (Sensor.decodeSensors flags.sensors)

app/javascript/elm/src/Sensor.elm
```diff
@@ -184,3 +186,19 @@ idForParameterOrLabel parameterOrLabel oldSensorId sensors =
...
+
+nameForSensorId : List Sensor -> String -> Maybe String
+nameForSensorId sensors id =
+    sensors
+        |> List.filter (\sensor -> sensor.id_ == id)
+        |> List.head
+        |> Maybe.map .name
+
+
+unitForSensorId : List Sensor -> String -> Maybe String
```
3v0k4: test

app/javascript/elm/src/Sensor.elm
```diff
@@ -184,3 +186,19 @@ idForParameterOrLabel parameterOrLabel oldSensorId sensors =
 
         _ ->
             defaultSensorId
+
+
+nameForSensorId : List Sensor -> String -> Maybe String
```
3v0k4: test

app/javascript/elm/src/Main.elm
```diff
@@ -264,7 +280,7 @@ update msg model =
         ToggleSessionSelectionFromAngular maybeId ->
             case maybeId of
                 Just id ->
-                    ( model, SelectedSession.fetch model.selectedSensorId model.page id GotSession )
+                    ( model, SelectedSession.fetch model.selectedSensorId model.page id (RemoteData.fromResult >> GotSession) )
```
gitjul: what do you think of using `<<` to make it read better? (same for other places)

app/javascript/elm/src/Sensor.elm
```diff
@@ -50,7 +52,7 @@ mainSensors =
    { id_ = "Particulate Matter-airbeam2-pm2.5 (µg/m³)"
    , label = "AirBeam2-PM2.5 (µg/m³)"
    , parameter = "Particulate Matter"
-   , sensor = "AirBeam2-PM2.5"
+   , name = "AirBeam2-PM2.5"
```
gitjul: 👌

app/javascript/packs/elm.js
```diff
@@ -69,5 +69,9 @@ const setupHeatMap = () => {
     for (var i = 0; i < connect.length; i++) {
       connect[i].classList.add(classes[i]);
     }
+
+    window.__elmApp.ports.updateHeatMapThresholds.subscribe(({ h1, h2, h3, h4, h5 }) => {
+      node.noUiSlider.set([h2, h3, h4]);
```
gitjul: neat! I like this plugin already

app/javascript/elm/tests/SensorTests.elm
```diff
@@ -145,38 +129,34 @@ all =
...
-                    |> nameForSensorId "id"
+                let
+                    sensor1 =
+                        { sensor | name = "name" }
+
+                    sensor2 =
+                        { sensor | name = "other name" }
+                in
+                [ sensor1, sensor2 ]
+                    |> nameForSensorId (toId sensor1)
```
gitjul: hm, I feel like when I'm reading such a test description I'd expect the test to call `nameForSensorId` on a `sensorId` instead of going deeper and calling `toId`. then `toId` and `toLabel` could be tested separately. I guess this is a more "integration-like" approach? in that case, since the input is actually a `sensor` not a `sensorId`, the test description seems off. I'm probably nitpicking.

app/javascript/elm/src/Main.elm
```diff
@@ -103,6 +106,9 @@ init flags url key =
 
                 _ ->
                     Mobile
+
+        sensors =
+            Sensor.decodeSensors flags.sensors
```
gitjul: just to leave a trace you were talking about changing this to `decode` and refactoring the code around this

app/javascript/elm/src/Main.elm
3v0k4: You'd prefer ` GotSession << RemoteData.fromResult` instead of `RemoteData.fromResult >> GotSession` ?

app/javascript/elm/tests/SensorTests.elm
3v0k4: I see what you mean but in this case it's much more convenient to "integrate". Otherwise I would have to write a correct id by hand or copy the logic from the production code which seems smelly

app/javascript/packs/elm.js
3v0k4: Yeah it's cool!

app/javascript/elm/src/Main.elm
3v0k4: Thanks!

app/javascript/elm/src/Main.elm
gitjul: this style was suggested as possibly better / easier to read in Elm in Action and at first I didn't know why but now it does make sense - I'm passing the result of the function to another one like I would pass an argument, so this order is more comfortable to read, I think. wdyt?

3v0k4: I'm all in for that

3v0k4: I'll be doing it from now on and refactor whenever I see composition

PR #161, opened 2019-04-16
Heat 1

3v0k4:
* fix encoding problem in query params
* update heat map when sensor is selected
* change extremes with min and max input

### DIFF
--- a/app/javascript/elm/tests/Data/HeatMapThresholdsTests.elm
+++ b/app/javascript/elm/tests/Data/HeatMapThresholdsTests.elm
@@ -1,0 +1,27 @@
+module Data.HeatMapThresholdsTests exposing (suite)
+
+import Data.HeatMapThresholds exposing (extremes, updateMaximum, updateMinimum)
+import Expect
+import Fuzz exposing (int, list, string)
+import Json.Encode as Encode
+import Sensor exposing (..)
+import Test exposing (..)
+
+
+suite : Test
+suite =
+    describe "Data.HeatMapThresholds"
+        [ fuzz2 int int "updateMinimum, updateMaximum and extremes" <|
+            \h1 h5 ->
+                let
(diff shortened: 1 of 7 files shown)


3v0k4:
Merged

app/javascript/elm/tests/Data/HeatMapThresholdsTests.elm
```diff
@@ -0,0 +1,27 @@
...
+import Fuzz exposing (int, list, string)
+import Json.Encode as Encode
+import Sensor exposing (..)
+import Test exposing (..)
+
+
+suite : Test
+suite =
+    describe "Data.HeatMapThresholds"
+        [ fuzz2 int int "updateMinimum, updateMaximum and extremes" <|
```
gitjul: (note to self?) what happens if h5 is smaller than h1? is this ever the case?

3v0k4: I'm answering this as part of the PR conversation ;)

3v0k4:
When it comes to changing the extremes we had a chat with Michael and decided that it's fine not to handle the case where one gets equal or bigger / smaller than the other. Rationale is that it's an edge case and that the user can still "recover". Or at least I didn't notice any situation where by setting values the user could blow up the app in a way that they can't go back. But I could have missed something!

If you feel like it's something that is worth investing time into, please feel free to add a ticket / discuss further with Michael.

gitjul:
I think it does explode sometimes i.e. you can get to a point where this functionality (the slider) isn't usable, and reloading doesn't help. I'll add a ticket and I'll show you.

gitjul:
I don't really like the situation experience but not enough to pressure us to fix it. We can get back to it if it becomes a problem.

PR #162, opened 2019-04-17
Heat

3v0k4:
Add "reset to default" button in heatmap.

I've decided to be more conservative with testing. I feel that a lot of testing I'm doing in Elm is just rewriting stuff twice. And that comes at the cost of having to maintain lots of tests. I'll try to test only critical things or stuff that justifies having a test behind.

### DIFF
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -181,6 +181,7 @@ type Msg
     | UpdateHeatMapThresholds (WebData HeatMapThresholds)
     | UpdateHeatMapMinimum String
     | UpdateHeatMapMaximum String
+    | ResetHeatMapToDefaults
 
 
 update : Msg -> Model -> ( Model, Cmd Msg )
@@ -329,7 +330,7 @@ update msg model =
                 cmd =
                     case heatMapThresholds of
                         Success thresholds ->
-                            Ports.updateHeatMapThresholds thresholds
+                            Ports.updateHeatMapThresholds <| HeatMapThresholds.toValues thresholds
 
                         _ ->
                             Cmd.none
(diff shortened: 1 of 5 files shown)


app/javascript/elm/src/Main.elm
```diff
@@ -467,6 +478,7 @@ viewHeatMapSlider heatMapThresholds sensorUnit =
         [ div [ class "heatmap-input" ] [ viewHeatMapInput "min" threshold1 sensorUnit UpdateHeatMapMinimum ]
         , div [ id "heatmap" ] []
         , div [ class "heatmap-input" ] [ viewHeatMapInput "max" threshold5 sensorUnit UpdateHeatMapMaximum ]
+        , button [ Events.onClick ResetHeatMapToDefaults ] [ text "D" ]
```
gitjul: couldn't find that button :D why not sth like "Reset heatmap"?

app/javascript/elm/tests/MainTests.elm
```diff
@@ -22,6 +23,26 @@ import Test.Html.Selector as Slc
 import TimeRange
 
 
+heatMapThresholdsWithMinimum : Int -> HeatMapThresholds
+heatMapThresholdsWithMinimum value =
+    { threshold1 = { value = value, default = 1 }
```
gitjul: what if value is higher than one or all of the other thresholds?

app/javascript/elm/tests/MainTests.elm
```diff
@@ -22,6 +23,26 @@ import Test.Html.Selector as Slc
...
+    }
+
+
+heatMapThresholdsWithMaximum : Int -> HeatMapThresholds
+heatMapThresholdsWithMaximum value =
+    { threshold1 = { value = 1, default = 1 }
+    , threshold2 = { value = 2, default = 2 }
+    , threshold3 = { value = 3, default = 3 }
+    , threshold4 = { value = 4, default = 4 }
+    , threshold5 = { value = value, default = 5 }
```
gitjul: what if value is lower than one or all of the other thresholds?

app/javascript/elm/src/Main.elm
gitjul: aaa... https://llp.kanbanery.com/projects/8040/board/tasks/2403287/attachments
:)

app/javascript/elm/tests/MainTests.elm
3v0k4: I guess for the purpose of the tests I'm just looking at that value regardless of the others. It's a good point though. We could have a smart constructor that does the checking and returns a maybe in case we needed to have that in the production code. Something like
```
build : Maybe Heat
build val =
  if isOk val then
    Just ..
  else
    Nothing
```

app/javascript/elm/tests/MainTests.elm
3v0k4: as above

PR #165, opened 2019-04-18
Load heat values from url

3v0k4:
If there's any values in the URL use them (both as current values and defaults) otherwise fetch them depending on the selected sensor

### DIFF
--- a/app/javascript/packs/elm.js
+++ b/app/javascript/packs/elm.js
@@ -20,6 +20,14 @@ document.addEventListener('DOMContentLoaded', () => {
     console.warn(params);
     console.warn(data);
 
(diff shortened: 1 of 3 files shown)


app/javascript/packs/elm.js
```diff
@@ -20,6 +20,14 @@ document.addEventListener('DOMContentLoaded', () => {
     console.warn(params);
     console.warn(data);
 
+    const heatMapThresholdValues = data.heat ? {
+      threshold1: data.heat.lowest,
```
gitjul: what happens if those are undefined? (can they?)

3v0k4: I think they are either there or the entire hash in `data.heat` is missing

PR #166, opened 2019-04-19
Format dates

3v0k4:
* Whenever dates are on the same date (ie same month, year, day) show "MM/DD/YYYY mm:hh mm:hh" instead of "MM/DD/YYYY mm:hh MM/DD/YYYY mm:hh"

To make that happen I've decided to pass to Elm the time in milliseconds and decide in Elm what to do. Since the times come from Angular for the sessions list and from Rails for the single session this is the best way to avoid rewriting the logic in both Rails and Angular. Also, this way we have more options in the future when it comes to manipulate times in the frontend. 
Read the [readme for elm/time](https://package.elm-lang.org/packages/elm/time/latest/) to see why using Posix is a good idea.

Up until now we were passing `List Session` though the port and leaving to the Elm runtime to translate the JSON into a value of type `List Session`. Now that we are passing times in milliseconds the easy way to get sessions with time in posix is to decode them. Plus having the decoder gives us more options when it comes to handling failures.

### DIFF
--- a/app/javascript/angular/code/services/sessions_downloader.js
+++ b/app/javascript/angular/code/services/sessions_downloader.js
@@ -37,9 +37,8 @@ angular.module("aircasting").factory("sessionsDownloader", ['$rootScope', '$http
 
     _(data).each(function(session){
       if(session.start_time_local && session.end_time_local) {
-        times = [moment(session.start_time_local, "YYYY-MM-DDTHH:mm:ss"), moment(session.end_time_local, "YYYY-MM-DDTHH:mm:ss")];
-        session.startTime = times[0].format('MM/DD/YYYY, HH:mm');
-        session.endTime = times[1].format('MM/DD/YYYY, HH:mm');
+        session.startTime = moment(session.start_time_local).utc().valueOf();
+        session.endTime = moment(session.end_time_local).utc().valueOf();
       }
 
       session.streams = keysToLowerCase(session.streams)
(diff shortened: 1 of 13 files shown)


app/javascript/angular/code/services/sessions_downloader.js
```diff
@@ -37,9 +37,8 @@ angular.module("aircasting").factory("sessionsDownloader", ['$rootScope', '$http
 
     _(data).each(function(session){
       if(session.start_time_local && session.end_time_local) {
-        times = [moment(session.start_time_local, "YYYY-MM-DDTHH:mm:ss"), moment(session.end_time_local, "YYYY-MM-DDTHH:mm:ss")];
-        session.startTime = times[0].format('MM/DD/YYYY, HH:mm');
-        session.endTime = times[1].format('MM/DD/YYYY, HH:mm');
+        session.startTime = moment(session.start_time_local).utc().valueOf();
```
3v0k4: `valueOf()` returns POSIX in milliseconds (what we need later on in Elm)

app/javascript/elm/tests/Data/SelectedSessionTests.elm
```diff
@@ -1,9 +1,7 @@
 module Data.SelectedSessionTests exposing (suite)
 
-import Data.SelectedSession exposing (sensorNameFromId)
+import Data.SelectedSession exposing (..)
```
gitjul: why do you prefer to do this?

3v0k4: Don't know. I'll refactor to use an explicit import

PR #167, opened 2019-04-23
Avoid session path below graph when selecting a session

3v0k4:

PR #168, opened 2019-04-23
Avoid session path below graph when selecting a session

3v0k4:
Turns out https://github.com/HabitatMap/AirCasting/pull/167 does not work. But this will cause turns out [`fitBounds` accepts a (padding as a second arg](https://developers.google.com/maps/documentation/javascript/reference/map#Map.fitBounds)

### DIFF
--- a/app/javascript/angular/code/services/google/_map.js
+++ b/app/javascript/angular/code/services/google/_map.js
@@ -66,7 +66,7 @@ export const map = (
         if (!googleMaps.wasGeocodingSuccessful(status)) return;
 
         const latLngBounds = results[0].geometry.viewport;
-        this._fitBoundsWithoutPanOrZoomCallback(this.mapObj, latLngBounds);
+        this._fitBoundsWithoutPanOrZoomCallback(latLngBounds);
       };
 
       geocoder.get(address, callback);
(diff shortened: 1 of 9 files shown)


app/javascript/angular/code/services/google/_map.js
```diff
@@ -95,6 +95,20 @@ export const map = (
     },
 
     fitBounds: function(bounds, zoom) {
+      const fnc = (latLngBounds, zoom) => () =>
```
gitjul: not a fan of these `fnc` functions, even though they are used near their definitions

app/javascript/angular/code/services/google/google_maps.js
```diff
@@ -14,6 +16,14 @@ angular.module("google").factory("googleMaps", [
 
       fitBounds: (mapObj, viewport) => mapObj.fitBounds(viewport),
 
+      fitBoundsWithBottomPadding: (mapObj, viewport) =>
+        mapObj.fitBounds(viewport, {
+          bottom: SELECTED_SESSION_DIV_HEIGHT,
+          top: 0,
+          left: 0,
+          right: 0
+        }),
```
gitjul: nice solution, so much better

app/javascript/angular/code/services/google/_map.js
3v0k4: If you have a better idea please share!

gitjul: would if I had, sorry

PR #176, opened 2019-04-30
Heatmap ui tweaks

gitjul:
- [x] Use proper grey in colors.css
- [x] Style heat map inputs
- [x] Tweak slider to fix width issues
- [x] Style "reset to default" button in heat map
- [x] Add alts to images in elm


app/assets/stylesheets/config/typography.css.scss
```diff
@@ -4,7 +4,6 @@ body {
   color: black;
   -moz-osx-font-smoothing: grayscale;
   -webkit-font-smoothing: antialiased;
-  font-smoothing: antialiased;
```
gitjul: there's no such property (there's [`font-smooth`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth) but it's non standard and I wouldn't use it.)

btw there are reason's not to use smoothing, it's not always the best idea, e.g. contrast is worse
- http://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/
- https://www.zachleat.com/web/font-smooth/

app/assets/stylesheets/modules/buttons-links.css.scss
```diff
@@ -1,22 +1,22 @@
...
   font-weight: $font-stack-bold;
+  height: 33px;
+  text-align: center;
 }
 
-.circular-button {
+.link-button {
   height: 33px;
+  margin-top: 3px;
   width: 33px;
```
gitjul: this is pretty specific and there's nothing else using this class now, so I'm merging those two classes

app/assets/stylesheets/modules/buttons-links.css.scss
```diff
@@ -1,22 +1,22 @@
...
-}
 
-.link-button {
-  marign-top: 3px;
+  img {
+    vertical-align: middle;
```
gitjul: fixes alignment. note: changing line-height messes with this, but there's no reason to set line-height if no text

app/assets/stylesheets/config/typography.css.scss
3v0k4: Cool stuff! Thanks for sharing

PR #177, opened 2019-04-30
Graph Tweaks

3v0k4:
- [x] render empty graph with "loading data" before firing the request for the measurements
- [x] use `requestAnimationFrame` to avoid Highcharts not finding the div
- [x] when hovering on graph the bubble with the details of each measurements should be on top and caret should be pointing down
- [x] use correct colors in graph and not greys
- [x] redraw on heat map change
- [x] avoid the double draw caused by ToggleSession and ToggleSessionFromAngular

Review commit by commit if you prefer to check smaller changes

### DIFF
--- a/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
+++ b/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
@@ -110,11 +110,14 @@ export const SessionsListCtrl = (
   );
 
   $scope.$on("markerSelected", function(event, data) {
-    $scope.toggleSession(data.session_id, true);
+    $scope.toggleSession(
+      data.session_id,
+      elmApp.ports.toggleSessionSelection.send
+    );
     $scope.$apply();
   });
 
-  $scope.toggleSession = function(sessionId, markerSelected) {
+  $scope.toggleSession = function(sessionId, callback) {
     if (this.isSessionDisabled(sessionId)) {
       flash.set(CANNOT_SELECT_MULTIPLE_SESSIONS);
       return;
@@ -122,11 +125,11 @@ export const SessionsListCtrl = (
     var session = sessions.find(sessionId);
     if (sessions.isSelected(session)) {
       params.update({ selectedSessionIds: [] });
-      elmApp.ports.toggleSessionSelection.send(null);
+      callback(null);
     } else if ($scope.canSelectSession(sessionId)) {
       params.update({ selectedSessionIds: [sessionId] });
-      $scope.markerSelected.set(markerSelected);
-      elmApp.ports.toggleSessionSelection.send(sessionId);
+      $scope.markerSelected.set(true);
+      callback(sessionId);
     }
   };
 
@@ -144,8 +147,8 @@ export const SessionsListCtrl = (
   if (process.env.NODE_ENV !== "test") {
     angular.element(document).ready(() => {
       elmApp.ports.toggleSession.subscribe(({ selected, deselected }) => {
-        if (deselected) $scope.toggleSession(deselected, true);
-        if (selected) $scope.toggleSession(selected, true);
+        if (deselected) $scope.toggleSession(deselected, () => {});
+        if (selected) $scope.toggleSession(selected, () => {});
         $scope.$apply();
       });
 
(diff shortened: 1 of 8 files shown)


app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -110,23 +110,26 @@ export const SessionsListCtrl = (
...
+    $scope.toggleSession(
+      data.session_id,
+      elmApp.ports.toggleSessionSelection.send
+    );
     $scope.$apply();
   });
 
-  $scope.toggleSession = function(sessionId, markerSelected) {
+  $scope.toggleSession = function(sessionId, callback) {
     if (this.isSessionDisabled(sessionId)) {
```
AkkarinForest: do we still need that if we have radio buttons now?

3v0k4: Good catch. Prolly not but will leave this to another round of removals

app/javascript/elm/src/Main.elm
```diff
@@ -317,11 +317,18 @@ update msg model =
...
+                        ( { model | selectedSession = NotAsked }, Cmd.none )
+
+                    else
+                        ( model, SelectedSession.fetch model.sensors model.selectedSensorId model.page id (RemoteData.fromResult >> GotSession) )
+
+                ( _, Just id ) ->
                     ( model, SelectedSession.fetch model.sensors model.selectedSensorId model.page id (RemoteData.fromResult >> GotSession) )
 
-                Nothing ->
+                ( _, Nothing ) ->
```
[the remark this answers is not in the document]
3v0k4: I see what you mean. At the same time, I'm sure sure how to simplify that w/o making it more complex.

I'm not sure it's an impossible state. I think that from Angular we could expect anything coming at this point.

3v0k4:
@AkkarinForest thanks for the CR!

app/javascript/packs/elm.js
```diff
@@ -140,7 +140,7 @@ const setupHeatMap = () => {
 };
 
 const draw = fnc => ({ times, streamId, heat, sensor }) =>
-  fnc({ sensor, heat, times, streamId });
+  window.requestAnimationFrame(() => fnc({ sensor, heat, times, streamId }));
```
gitjul: interesting, didn't know that trick...

app/javascript/elm/src/Main.elm
```diff
@@ -448,10 +448,10 @@ toGraphParams thresholds selectedSession sensors selectedSensorId =
...
 
         levels =
-            [ { from = threshold1, to = threshold2, color = "#aaa" }
-            , { from = threshold2, to = threshold3, color = "#bbb" }
-            , { from = threshold3, to = threshold4, color = "#ccc" }
-            , { from = threshold4, to = threshold5, color = "#ddd" }
+            [ { from = threshold1, to = threshold2, className = "first-band" }
+            , { from = threshold2, to = threshold3, className = "second-band" }
+            , { from = threshold3, to = threshold4, className = "third-band" }
+            , { from = threshold4, to = threshold5, className = "fourth-band" }
```
gitjul: why `-band`? (just trying to understand the meaning, what it reflects)

gitjul: (in case I wanna change sth here as you suggested)

gitjul: another words: what's a graph band?

app/javascript/packs/elm.js
```diff
@@ -126,11 +126,9 @@ const setupHeatMap = () => {
       console.warn("heatmap updated", Object.values(toValues(node.noUiSlider)));
 
       // changing extremes could have changed middle values
```
gitjul: is this comment still relevant?

app/javascript/elm/src/Main.elm
3v0k4: I guess this should answer the question https://www.highcharts.com/docs/chart-concepts/plot-bands-and-plot-lines

app/javascript/packs/elm.js
3v0k4: yep

3v0k4:
Thanks for the CR @gitjul

PR #179, opened 2019-05-03
Fix markers not showing on the map

gitjul:
Fixes marker image paths by linking to precompiled assets with the help of `rails-erb-loader`, according to this method: https://github.com/rails/webpacker/blob/master/docs/assets.md#import-from-sprockets-using-helpers.

### DIFF
--- a/app/javascript/angular/code/services/_draw_session.js.erb
+++ b/app/javascript/angular/code/services/_draw_session.js.erb
@@ -1,5 +1,12 @@
 import _ from "underscore";
-import { removeMarker } from "./google/_map";
+import { removeMarker } from "./google/_map.js.erb";
+
+<% helpers = ActionController::Base.helpers %>
+
+const locationMarker1Path = "<%= helpers.image_path('location_marker1.svg') %>";
+const locationMarker2Path = "<%= helpers.image_path('location_marker2.svg') %>";
+const locationMarker3Path = "<%= helpers.image_path('location_marker3.svg') %>";
+const locationMarker4Path = "<%= helpers.image_path('location_marker4.svg') %>";
 
 export const drawSession = (sensors, map, heat, note, empty) => {
   var DrawSession = function() {};
@@ -94,7 +101,7 @@ const createMeasurementMarker = (measurement, idx, heat, map, suffix) => {
     position: { lat: measurement.latitude, lng: measurement.longitude },
     title: roundedValue.toString() + suffix,
     zIndex: idx,
-    icon: "/assets/location_marker" + level + ".svg"
+    icon: eval(`locationMarker${level}Path`)
   });
 
   return marker;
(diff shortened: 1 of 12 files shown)


app/javascript/angular/code/services/_draw_session.js.erb
```diff
@@ -1,5 +1,12 @@
 import _ from "underscore";
-import { removeMarker } from "./google/_map";
+import { removeMarker } from "./google/_map.js.erb";
+
+<% helpers = ActionController::Base.helpers %>
+
+const locationMarker1Path = "<%= helpers.image_path('location_marker1.svg') %>";
+const locationMarker2Path = "<%= helpers.image_path('location_marker2.svg') %>";
+const locationMarker3Path = "<%= helpers.image_path('location_marker3.svg') %>";
+const locationMarker4Path = "<%= helpers.image_path('location_marker4.svg') %>";
```
gitjul: I wanted to extract a nice `imagePath` service to be used in nicely extracted functions but it's proven to complicate things too much. so I decided not to hate this.

app/javascript/angular/code/services/google/graph_highlight.js.erb
```diff
@@ -10,13 +13,15 @@ export const hide = () => {
 };
 
 export const show = points => {
+  
+
```
AkkarinForest: I guess u don't need two lines here.

app/javascript/angular/code/services/_draw_session.js.erb
AkkarinForest: Complicate things how? Maybe we can investigate if there is a simpler solution to extract something here?

3v0k4:
Did not know about this, cool. In the long term I'd consider moving away from the rails asset pipeline and use something like https://stackoverflow.com/questions/34257800/how-to-set-up-an-inline-svg-with-webpack (it works similarly as when importing the images and passing them through flags).

Rationale is that the map part of the app is entirely in JS and Webpack. For that reason I'd leave them in charge of everything w/o delegating to Rails

gitjul:
@3v0k4 that's right, I thought about doing it similarly to flags, but the thing is that I actually need the path to the file here, not the file itself. the marker accepts an `icon: "path_to_image.svg"` parameter. do you know how I could achieve that?

app/javascript/angular/code/services/_draw_session.js.erb
gitjul: I don't want to bother at this moment but might try doing it again sometime later when I touch the code again, and explain it better. might be that I was trying to do too much at once, but I extracted sth like this: 

```
const imagePath = imageName => {
  <% helpers = ActionController::Base.helpers %>  
  return "<%= helpers.image_path(imageName) %>";
};

export default imagePath;
```
and it wasn't enough just to use it in the particular files, things got broken. took too long so I wen't back to this simplistic version. I don't it's worth spending too much time on.

3v0k4:
Actually since we are using the file-loader (the one used for the flags) we can use that for svgs too it seems (https://stackoverflow.com/questions/34257800/how-to-set-up-an-inline-svg-with-webpack#34257837)

PR #180, opened 2019-05-06
Persistent sessions list

AkkarinForest:
Now the number of loaded sessions is saved in the params, so links with more than 50 work.
The map position before selecting a session is saved in the params, so links with preselected session allow a user to deselect it and see the original list (works even when more than 50 was loaded).

### DIFF
--- a/app/controllers/api/measurement_sessions_controller.rb
+++ b/app/controllers/api/measurement_sessions_controller.rb
@@ -34,7 +34,7 @@ module Api
       data[:time_from] = Time.strptime(data[:time_from].to_s, '%s')
       data[:time_to] = Time.strptime(data[:time_to].to_s, '%s')
 
-      respond_with MobileSession.filtered_json(data, page, page_size)
+      respond_with MobileSession.filtered_json(data, page_size, params[:offset])
     end
 
     def show_multiple
@@ -84,10 +84,6 @@ module Api
     end
 
     private
-
-    def page
-      @page ||= params[:page] || 0
-    end
 
     def page_size
       @page_size ||= params[:page_size] || 50
--- a/app/controllers/api/measurement_sessions_controller.rb
+++ b/app/controllers/api/measurement_sessions_controller.rb
@@ -18,7 +18,7 @@
 
 module Api
   class MeasurementSessionsController < BaseController
-    INT_Q_ATTRS = [:time_from, :time_to, :day_from, :day_to]
+    INT_Q_ATTRS = [:time_from, :time_to, :day_from, :day_to, :limit, :offset]
 
     # TokenAuthenticatable was removed from Devise in 3.1
     # https://gist.github.com/josevalim/fb706b1e933ef01e4fb6
@@ -34,7 +34,7 @@ module Api
       data[:time_from] = Time.strptime(data[:time_from].to_s, '%s')
       data[:time_to] = Time.strptime(data[:time_to].to_s, '%s')
 
-      respond_with MobileSession.filtered_json(data, page_size, params[:offset])
+      respond_with MobileSession.filtered_json(data, data[:limit], data[:offset])
     end
 
     def show_multiple
@@ -84,10 +84,6 @@ module Api
     end
 
     private
-
-    def page_size
-      @page_size ||= params[:page_size] || 50
-    end
 
     def decoded_query_data(query)
       if query.is_a?(String)
--- a/app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
+++ b/app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
@@ -85,6 +85,7 @@ export const FixedSessionsMapCtrl = (
       elmApp.ports.selectSensorId.subscribe(sensorId => {
         params.update({ selectedSessionIds: [] });
         params.update({ data: { sensorId } });
+        params.update({ fetchedSessionsCount: 0 });
         sensors.fetchHeatLevels();
         $scope.sessions.fetch();
       });
@@ -95,11 +96,13 @@ export const FixedSessionsMapCtrl = (
 
       elmApp.ports.toggleIndoor.subscribe(isIndoor => {
         params.update({ data: { isIndoor: isIndoor } });
+        params.update({ fetchedSessionsCount: 0 });
         $scope.sessions.fetch();
       });
 
       elmApp.ports.toggleStreaming.subscribe(isStreaming => {
         params.update({ data: { isStreaming } });
+        params.update({ fetchedSessionsCount: 0 });
         resetTimeRangeFilter();
         $scope.sessions.fetch();
       });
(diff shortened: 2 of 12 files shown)


app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -43,44 +43,17 @@ export const SessionsListCtrl = (
...
-    true
-  );
-
-  $scope.$watch(
-    "sessionFetchCondition()",
-    (newValue, oldValue) => {
-      console.log("watch - sessionFetchCondition()", newValue, oldValue);
-      if (sessions.hasSelectedSessions()) return;
-      $scope.page = 0;
-      sessions.fetch($scope.page);
```
AkkarinForest: I removed this watch cause it was impossible to track what's going on. There were two watches:
1. on line 66 - which reloads sessions when map moves
2. this one - which reloads sessions when something in the params data changes (so when some of the filters change)
The problem is that when the page is open for the first time both watches where triggered and probable sessions where fetched twice - anyway it was impossible to track what's going on.

I think this watch is not necessary, since we are calling sessions.fetch directly from mobile_sessions_map_crl and fixed_sessions_map_crl when something changes with the filters.

But I'm leaving this longish explanation here, cause it's possible that I missed some use case for this watch - so if you ever notice that the session list is not reloaded when it should - this might be one of the reasons.

app/javascript/angular/code/services/_fixed_sessions.js
```diff
@@ -254,24 +265,26 @@ export const fixedSessions = (
         });
       }
 
-      _(params).extend({ page: page });
-
       drawSession.clear(this.sessions);
 
-      if (page === 0) {
-        this.sessions = [];
+      if (params.get("selectedSessionIds").length === 1) {
```
3v0k4: This is a great one!

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -153,8 +126,9 @@ export const SessionsListCtrl = (
       });
 
       elmApp.ports.loadMoreSessions.subscribe(() => {
-        $scope.updateSessionsPage();
-        $scope.$apply();
+        sessions.fetch({
+          fetchedSessionsCount: params.get("fetchedSessionsCount")
```
3v0k4: Could we do something like `sessions.length` in the sessions module and not use this at all?

app/javascript/angular/code/services/_fixed_sessions.js
```diff
@@ -23,10 +23,15 @@ export const fixedSessions = (
     this.scope.params = params;
   };
 
-  let prevMapPosition = {
-    bounds: map.getBounds(),
-    zoom: map.getZoom()
-  };
+  let prevMapPosition = {};
+  if (params.get("selectedSessionIds").length === 1) {
+    prevMapPosition = params.get("prevMapPosition");
```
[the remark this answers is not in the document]
AkkarinForest: Yes, cause it's only set when session is selected.

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -21,10 +21,10 @@ export const SessionsListCtrl = (
     $scope.storage = storage;
     $scope.$window = $window;
     $scope.sensors = sensors;
-    $scope.page = 0;
     $scope.markerSelected = markerSelected;
     $window.sessions = sessions = $scope.sessions;
     $scope.sessionsForList = [];
+    $scope.firstLoad = true;
```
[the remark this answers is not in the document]
AkkarinForest: I don't see any simple way. At least not when we still have a lot of watches and angular stuff...

app/controllers/api/measurement_sessions_controller.rb
```diff
@@ -18,7 +18,7 @@
 
 module Api
   class MeasurementSessionsController < BaseController
-    INT_Q_ATTRS = [:time_from, :time_to, :day_from, :day_to]
+    INT_Q_ATTRS = [:time_from, :time_to, :day_from, :day_to, :limit, :offset]
```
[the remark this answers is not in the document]
AkkarinForest: Yes! Good catch, thanks :)

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -153,8 +126,9 @@ export const SessionsListCtrl = (
       });
 
       elmApp.ports.loadMoreSessions.subscribe(() => {
-        $scope.updateSessionsPage();
-        $scope.$apply();
+        sessions.fetch({
+          fetchedSessionsCount: sessions.sessions.length
```
gitjul: is there a way to avoid this unintuitive nesting?

app/javascript/angular/code/services/_mobile_sessions.js
```diff
@@ -24,10 +24,15 @@ export const mobileSessions = (
...
-  };
+  let prevMapPosition = {};
+  if (params.get("selectedSessionIds").length === 1) {
+    prevMapPosition = params.get("prevMapPosition");
+  } else {
+    prevMapPosition = {
+      bounds: map.getBounds(),
+      zoom: map.getZoom()
+    };
+  }
```
gitjul: this is the same as in `fixed_sessions`, right? what do you think of extracting it?

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -133,7 +133,7 @@ export const FixedSessionsMapCtrl = (
       const onTimeRangeChanged = (timeFrom, timeTo) => {
         elmApp.ports.timeRangeSelected.send({ timeFrom, timeTo });
         params.update({ data: { timeFrom, timeTo } });
-        sessions.fetch();
+        $scope.sessions.fetch();
```
gitjul: oh, so this was broken before, was it?

AkkarinForest: I think so, but there was a watch that was fetching the sessions anyway.

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -43,44 +43,17 @@ export const SessionsListCtrl = (
...
-
-  $scope.$watch(
-    "sessionFetchCondition()",
-    (newValue, oldValue) => {
-      console.log("watch - sessionFetchCondition()", newValue, oldValue);
-      if (sessions.hasSelectedSessions()) return;
-      $scope.page = 0;
-      sessions.fetch($scope.page);
+      if ($scope.firstLoad || hasChangedProgrammatically) {
+        sessions.fetch({ amount: params.get("fetchedSessionsCount") });
```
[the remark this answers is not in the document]
AkkarinForest: In line 25 I `$scope.sessions` is assigned to `sessions` - don't know why...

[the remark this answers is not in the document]
gitjul: I see. it looks weird, `$window.sessions = sessions = $scope.sessions;`. but not gonna try to understand the reason now

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
gitjul: (I saw the structure now and I get it)

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
gitjul: thx for finding this then :)

PR #181, opened 2019-05-07
Sessions count

AkkarinForest:
I implemented the logic for numbers in "showing x of y results". This also allowed to improve the logic responsible for displaying "Load More" button.

I checked the performance and the count of available sessions doesn't seem to slow anything.

### DIFF
--- a/app/controllers/api/measurement_sessions_controller.rb
+++ b/app/controllers/api/measurement_sessions_controller.rb
@@ -34,7 +34,8 @@ module Api
       data[:time_from] = Time.strptime(data[:time_from].to_s, '%s')
       data[:time_to] = Time.strptime(data[:time_to].to_s, '%s')
 
-      respond_with MobileSession.filtered_json(data, data[:limit], data[:offset])
+      respond_with sessions: MobileSession.filtered_json(data, data[:limit], data[:offset]),
+                   availableSessionsCount: MobileSession.filter(data).count
     end
 
     def show_multiple
--- a/app/controllers/api/measurement_sessions_controller.rb
+++ b/app/controllers/api/measurement_sessions_controller.rb
@@ -35,7 +35,7 @@ module Api
       data[:time_to] = Time.strptime(data[:time_to].to_s, '%s')
 
       respond_with sessions: MobileSession.filtered_json(data, data[:limit], data[:offset]),
-                   availableSessionsCount: MobileSession.filter(data).count
+                   fetchableSessionsCount: MobileSession.filter(data).count
     end
 
     def show_multiple
--- a/app/controllers/api/realtime/sessions_controller.rb
+++ b/app/controllers/api/realtime/sessions_controller.rb
@@ -37,7 +37,8 @@ module Api
         data[:time_from] = Time.strptime(data[:time_from].to_s, '%s')
         data[:time_to] = Time.strptime(data[:time_to].to_s, '%s')
 
-        respond_with FixedSession.filtered_json(data, data[:limit], data[:offset])
+        respond_with sessions: FixedSession.filtered_json(data, data[:limit], data[:offset]),
+                     availableSessionsCount: MobileSession.filter(data).count
       end
 
       def index_streaming
--- a/app/controllers/api/realtime/sessions_controller.rb
+++ b/app/controllers/api/realtime/sessions_controller.rb
@@ -38,7 +38,7 @@ module Api
         data[:time_to] = Time.strptime(data[:time_to].to_s, '%s')
 
         respond_with sessions: FixedSession.filtered_json(data, data[:limit], data[:offset]),
-                     availableSessionsCount: MobileSession.filter(data).count
+                     fetchableSessionsCount: MobileSession.filter(data).count
       end
 
       def index_streaming
--- a/app/javascript/angular/code/services/sessions_downloader.js
+++ b/app/javascript/angular/code/services/sessions_downloader.js
@@ -15,8 +15,8 @@ angular.module("aircasting").factory("sessionsDownloader", [
       errorCallback
     ) {
       var successCallback = function(data) {
-        preprocessData(data, sessions, params);
-        refreshSessionsCallback();
+        preprocessData(data.sessions, sessions, params);
+        refreshSessionsCallback(data.availableSessionsCount);
       };
       fetchPage(url, reqData, successCallback, errorCallback);
     };
--- a/app/javascript/angular/code/services/sessions_downloader.js
+++ b/app/javascript/angular/code/services/sessions_downloader.js
@@ -16,7 +16,7 @@ angular.module("aircasting").factory("sessionsDownloader", [
     ) {
       var successCallback = function(data) {
         preprocessData(data.sessions, sessions, params);
-        refreshSessionsCallback(data.availableSessionsCount);
+        refreshSessionsCallback(data.fetchableSessionsCount);
       };
       fetchPage(url, reqData, successCallback, errorCallback);
     };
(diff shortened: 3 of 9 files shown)


app/controllers/api/measurement_sessions_controller.rb
```diff
@@ -34,7 +34,8 @@ def index
       data[:time_from] = Time.strptime(data[:time_from].to_s, '%s')
       data[:time_to] = Time.strptime(data[:time_to].to_s, '%s')
 
-      respond_with MobileSession.filtered_json(data, data[:limit], data[:offset])
+      respond_with sessions: MobileSession.filtered_json(data, data[:limit], data[:offset]),
+                   availableSessionsCount: MobileSession.filter(data).count
```
3v0k4: In case you've used `amount` in the previous PR I'd suggest doing the same here

app/javascript/elm/tests/MainTests.elm
```diff
@@ -849,23 +864,45 @@ updateTests =
...
                 model
                     |> update (UpdateSessions newSessions)
                     |> Tuple.first
                     |> .sessions
                     |> Expect.equal [ newSession ]
+        , fuzz int "UpdateSessions decodes the encoded value and updates model.availableSessionsCount" <|
+            \count ->
+                let
+                    model =
+                        { defaultModel | availableSessionsCount = 0, selectedSession = NotAsked }
```
3v0k4: Is `selectedSession` needed?

app/javascript/elm/src/Main.elm
```diff
@@ -714,9 +730,9 @@ viewShortType length index shortType =
         ]
 
 
-viewLoadMore : Int -> Html Msg
-viewLoadMore sessionCount =
-    if sessionCount /= 0 && modBy 50 sessionCount == 0 then
+viewLoadMore : Int -> Int -> Html Msg
+viewLoadMore availableSessionsCount sessionCount =
+    if sessionCount < availableSessionsCount then
```
3v0k4: What a relief to see this gone!

app/javascript/elm/src/Main.elm
```diff
@@ -66,6 +67,7 @@ defaultModel =
     { page = Mobile
     , key = Nothing
     , sessions = []
+    , availableSessionsCount = 0
```
3v0k4: If I was reading this out of context "available" would be confusing. I'd suggest something like "fetchable" unless you have a better name. Maybe @gitjul has a suggestion?

app/javascript/elm/src/Main.elm
```diff
@@ -285,10 +287,16 @@ update msg model =
 
         UpdateSessions value ->
             let
+                decoder =
+                    Decode.map2 FetchedSessions
```
3v0k4: What if you do something like `Decode.map2 Tuple.pair` and then

```
                (sessions, amount) =
                    Decode.decodeValue (Decode.list Data.Session.decoder) value	                    Decode.decodeValue decoder value
                        |> Result.withDefault ([], 0)
```
That way you don't need to name the type.

Rationale: I feel that the `FetchedSession` type is a bit forced in this case

app/javascript/angular/code/services/sessions_downloader.js
```diff
@@ -15,8 +15,8 @@ angular.module("aircasting").factory("sessionsDownloader", [
       errorCallback
     ) {
       var successCallback = function(data) {
-        preprocessData(data, sessions, params);
-        refreshSessionsCallback();
+        preprocessData(data.sessions, sessions, params);
```
3v0k4: This change is correct? it seems strange done by itself w/o changing `preprocessData`

app/javascript/elm/src/Main.elm
gitjul: 🙌

app/controllers/api/realtime/sessions_controller.rb
```diff
@@ -37,7 +37,8 @@ def index
         data[:time_from] = Time.strptime(data[:time_from].to_s, '%s')
         data[:time_to] = Time.strptime(data[:time_to].to_s, '%s')
 
-        respond_with FixedSession.filtered_json(data, data[:limit], data[:offset])
+        respond_with sessions: FixedSession.filtered_json(data, data[:limit], data[:offset]),
+                     availableSessionsCount: MobileSession.filter(data).count
```
gitjul: I'd consider calling this simply `amount` or `count` or `quantity` or `total` in this context it should be clear, we're hitting the /sessions endpoint and getting back `{ sessions: ..., total: ... }`. but `{ sessions: ..., sessionAmount: ... }` is fine too I guess if you want to be very clear.

app/javascript/elm/src/Main.elm
gitjul: I was thinking of going simpler and using the word `total`, but `fetchable` is the most obvious and explicit so I'd just go with that.

app/javascript/angular/code/services/sessions_downloader.js
AkkarinForest: Yes, cause the structure fo `data` changed. Before it just contained a list of sessions and now it's like this:
```
{
        sessions: [],
        availableSessionsCount: 0
      }
```

app/javascript/elm/src/Main.elm
AkkarinForest: I agree, thanks! I was looking for a shorter way to do this :)

app/controllers/api/realtime/sessions_controller.rb
AkkarinForest: _* I find that confusing. If it says `filterUtils` - that sounds like there is one filter that has many utils. If it says `filtersUtils` - that sounds like many filters that can have multiple utils. That's how I've been always using it. Is that really grammatically incorrect? It seems like it's needed ssses, cause it's changing the meaning._

app/controllers/api/measurement_sessions_controller.rb
AkkarinForest: This is a different value than in the previous PR. If it just says `amount` then it wouldn't be clear that it's not the number of fetched sessions, but rather the number of available sessions (cause we are fetching only 50).
And as a side note, I really don't like using the word `amount` for countable items 😉

app/controllers/api/realtime/sessions_controller.rb
czorek: I agree that `FiltersUtils` sounds overkill, and I would also use it without the additional `s`, but in case of counting, I think I'm alright with that, and I use the form `sessionsCount`. This is because most of the time I know there are multiple sessions or whatever, and the name reflects that. It might also be in the context of an object that only deals with multiple instances of something. As an aside, I see a method like `getChildrenCount` in Android SDK, so I was wondering if that's also grammatically incorrect in this context, or not?

app/javascript/elm/tests/MainTests.elm
AkkarinForest: Not really, cause it's in the default model anyway. But it's in the similar test case above (line 854) and I think you wrote it, so I assumed you had a reason for that 😄 
I'll just deleted them both if it's not needed anymore.

app/controllers/api/realtime/sessions_controller.rb
gitjul: >  Is that really grammatically incorrect? It seems like it's needed ssses, cause it's changing the meaning.
Yes, and when I see it I always imagine it in an old German professor's voice ;) "Filter utils" means utils for filters, utils related to filters. It's not a possessive adjective, so it doesn't need `'s` or `s'` depending on sing/plur. It doesn't really matter if there's one or more filters. Like "crocodile tears" and not "crocodile's tears" or "crocodiles' tears".

> As an aside, I see a method like getChildrenCount in Android SDK, so I was wondering if that's also grammatically incorrect in this context, or not?
Imo `getChildrenCount` and `getChildCount` are both correct and ok. what would be incorrect is `childrensCount` or `childsCount`. Besides that, I agree with your take on "count".

PR #184, opened 2019-05-08
Tooltips

3v0k4:
I've noticed that `filtersUtils.js` adds as options
```
      animateFill: false,
      interactive: true,
      trigger: "manual"
```
could @AkkarinForest or @gitjul double check if any is needed in this case?

Notice that `TooltipText` is an [opaque data type](https://medium.com/@ckoster22/advanced-types-in-elm-opaque-types-ec5ec3b84ed2). That means the type system won't confuse `String`s with `TooltipText`

Stuff is unstyled, there's a ticket for that (https://llp.kanbanery.com/projects/8040/board/tasks/2407513). @gitjul let me know if I can simplify your style task in this PR.

### DIFF
--- a/app/javascript/elm/src/Tooltip.elm
+++ b/app/javascript/elm/src/Tooltip.elm
@@ -1,0 +1,71 @@
+module Tooltip exposing
+    ( TooltipText
+    , crowdMap
+    , fixedTab
+    , locationFilter
+    , mobileTab
+    , parameterFilter
+    , profilesFilter
+    , sensorFilter
+    , streamingToggleFilter
+    , tagsFilter
+    , timeRangeFilter
+    , typeToggleFilter
+    , view
+    )
+
+import Html exposing (Html, span, text)
+import Html.Attributes exposing (attribute)
+
+
+type TooltipText
+    = TooltipText String
+
+
+view : TooltipText -> Html msg
+view (TooltipText tooltipText) =
+    span [ attribute "data-tippy-content" tooltipText ] [ text "?" ]
+
+
+mobileTab =
+    TooltipText "The mobile tab displays measurements from mobile sessions. When recording mobile sessions, measurements are created, timestamped, and geolocated once per second. Average values for the duration of the session are displayed inside the session map markers."
+
+
+fixedTab =
+    TooltipText "The fixed tab displays measurements from fixed sessions. When recording fixed sessions, measurements are created and timestamped once per minute and geocoordinates are fixed to a set location. Hourly average values are displayed inside the session map markers. These measurements are updated on the hour."
+
+
+parameterFilter =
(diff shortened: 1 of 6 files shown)


AkkarinForest:
`trigger: "manual"` - needed, cause we trigger this tooltip programmatically. If you don't specify this the default will be "mouseenter focus"
`interactive: true,` - needed, means it allows users to click without hiding. The default is false
`animateFill: true` - I guess that's just a stylistic choice.

I guess you don't need them for the tooltips with explanations.
Here are descriptions of all options: https://atomiks.github.io/tippyjs/all-options/

app/javascript/elm/src/Tooltip.elm
```diff
@@ -0,0 +1,71 @@
...
+
+crowdMap =
+    TooltipText "The CrowdMap averages together all the measurements from all the sessions listed on the sessions list and displays these averages as colored grid cells. The color of each grid cell corresponds to the average intensity of all the measurements recorded in that area. Click on a grid cell to view the underlying data. Increase the CrowdMap resolution to reduce the size of the CrowdMap grid cells. Reduce the CrowdMap resolution to increase the size of the CrowdMap grid cells."
+
+
+typeToggleFilter =
+    TooltipText "Outdoor sessions are recorded by instruments located outdoors. Indoor measurements are recorded by instruments located indoors. Indoor sessions do not include geocoordinates and are therefore not geolocated on the map."
+
+
+streamingToggleFilter =
```
AkkarinForest: `activeToggleFilter`? Cause I think we want to unify the naming.

app/javascript/elm/src/Tooltip.elm
```diff
@@ -0,0 +1,71 @@
...
+import Html.Attributes exposing (attribute)
+
+
+type TooltipText
+    = TooltipText String
+
+
+view : TooltipText -> Html msg
+view (TooltipText tooltipText) =
+    span [ attribute "data-tippy-content" tooltipText ] [ text "?" ]
```
AkkarinForest: Awesome! I really like that structure with the opaque type, I'll try to use them more :)

app/javascript/elm/src/Tooltip.elm
3v0k4: I'll do that separately but good point. And thanks for the reminder!

app/javascript/elm/src/Tooltip.elm
3v0k4: Yeah let's get deep into types!

3v0k4:
@AkkarinForest thanks for the CR!

gitjul:
as for `animateFill: true` I don't remember what it did exactly, I'd check if and what it changes when I style tooltips :) maybe we don't even need it for the copy link, will check that too when I do this

PR #185, opened 2019-05-08
Center align location markers and blue hover marker

gitjul:
Before:

Screenshot 2019-05-08 at 17 45 36

After:

Screenshot 2019-05-08 at 17 46 58

### DIFF
--- a/app/javascript/angular/code/services/_draw_session.js.erb
+++ b/app/javascript/angular/code/services/_draw_session.js.erb
@@ -101,7 +101,11 @@ const createMeasurementMarker = (measurement, idx, heat, map, suffix) => {
     position: { lat: measurement.latitude, lng: measurement.longitude },
     title: roundedValue.toString() + suffix,
     zIndex: idx,
-    icon: eval(`locationMarker${level}Path`)
+    icon: {
+      anchor: new google.maps.Point(6, 6),
+      size: new google.maps.Size(12, 12),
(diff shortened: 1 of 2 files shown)


app/javascript/angular/code/services/_draw_session.js.erb
```diff
@@ -101,7 +101,11 @@ const createMeasurementMarker = (measurement, idx, heat, map, suffix) => {
     position: { lat: measurement.latitude, lng: measurement.longitude },
     title: roundedValue.toString() + suffix,
     zIndex: idx,
-    icon: eval(`locationMarker${level}Path`)
+    icon: {
+      anchor: new google.maps.Point(6, 6),
+      size: new google.maps.Size(12, 12),
+      url: eval(`locationMarker${level}Path`)
```
3v0k4: Can we avoid using eval? Wouldn't the internal string be enough?

gitjul: the internal string is the variable name we're generating :) so we have to eval it to get the string it keeps

gitjul: cos it's assigned to `"<%= helpers.image_path('location_marker4.svg') %>"` etc., I think I'd still need eval to use it

3v0k4: Not sure about that. But it's a minor thing (and it was already there) ;)

PR #187, opened 2019-05-09
Export selected session

3v0k4:
You can review commit by commit.

Add button in the graph area to export the single session. Notice that I use an [extensible record type](https://medium.com/@ckoster22/advanced-types-in-elm-extensible-records-67e9d804030d) cause I could pass both a `Session` or a `SelectedSession` to `exportPath`.

Also, in the second commit I've removed the two view tests related to the export link and moved them to app/javascript/elm/tests/ApiTests.elm. 
I believe the important stuff to be tested is the link generation which is still covered. We lose the coverage on the fact that
- the button is not there when a session is selected; even if this was broken I don't see this as a big problem
- the button is there when there are no selections; this was validated when creating the feature and removing it takes a lot of distractions (by the author of the change, by the reviewers, by the manual testing)

### DIFF
--- a/app/javascript/elm/src/Data/SelectedSession.elm
+++ b/app/javascript/elm/src/Data/SelectedSession.elm
@@ -8,12 +8,13 @@ module Data.SelectedSession exposing
     , view
     )
 
+import Api
 import Data.HeatMapThresholds as HeatMapThresholds exposing (HeatMapThresholds)
 import Data.Page exposing (Page(..))
 import Data.Session
 import Data.Times as Times
-import Html exposing (Html, div, p, span, text)
-import Html.Attributes exposing (class)
+import Html exposing (Html, a, div, p, span, text)
+import Html.Attributes exposing (class, href, target)
 import Http
 import Json.Decode as Decode exposing (Decoder(..))
 import Json.Decode.Pipeline exposing (custom, required)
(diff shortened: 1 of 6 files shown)


app/javascript/elm/src/Data/SelectedSession.elm
```diff
@@ -134,4 +135,5 @@ view session heatMapThresholds =
                 ]
             ]
         , span [ class "single-session-date" ] [ text <| Times.format session.startTime session.endTime ]
+        , a [ target "_blank", href <| Api.exportLink [ session ] ] [ text "export sessions" ]
```
AkkarinForest: should that be "export session" for single session view?

app/javascript/elm/tests/MainTests.elm
```diff
@@ -66,24 +44,14 @@ popups =
...
-
 parameterSensorFilter : Test
 parameterSensorFilter =
     describe "Parameter filter tests: "
         [ test "parameter filter shows parameter name based on selectedSensorId" <|
             \_ ->
                 { defaultModel
                     | selectedSensorId = "parameter-sensor (unit)"
-                    , sensors = sensors
+                    , sensors = defaultSensors
```
AkkarinForest: Maybe we can have a `defaultModel` that already contains `defaultSensors` in TestUtils.elm. Would make these test cases less noisy.

app/javascript/elm/tests/MainTests.elm
```diff
@@ -707,7 +598,7 @@ viewTests =
             \title username sensorName ->
                 let
                     selectedSession_ =
-                        { selectedSession
+                        { defaultSelectedSession
```
AkkarinForest: Maybe you can also change `selectedSession_ ` to `selectedSession` now?

PR #188, opened 2019-05-09
Search as I move the map

AkkarinForest:
Search as I move checkbox, it's turned off by default. When the map is moved on zoomed it changes into "search again" button. When the sessions list is refreshed it changes back to the checkbox.
Needs styling: https://llp.kanbanery.com/projects/8040/board/tasks/2407695

### DIFF
--- a/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
+++ b/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
@@ -51,19 +51,18 @@ export const SessionsListCtrl = (
       if (hasChangedProgrammatically === undefined) return;
 
       if (firstLoad) {
-        sessions.fetch({ amount: params.get("fetchedSessionsCount") });
+        sessions.fetch({ amount: params.paramsData["fetchedSessionsCount"] });
         firstLoad = false;
         return;
       }
 
       if (!params.get("data").isSearchOn) {
-        console.warn("map was moved");
         elmApp.ports.mapMoved.send(null);
         return;
       }
 
       if (hasChangedProgrammatically) {
-        sessions.fetch({ amount: params.get("fetchedSessionsCount") });
+        sessions.fetch({ amount: params.paramsData["fetchedSessionsCount"] });
         return;
       }
       sessions.fetch();
--- a/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
+++ b/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
@@ -48,12 +48,19 @@ export const SessionsListCtrl = (
     ({ hasChangedProgrammatically }) => {
       console.log("watch - params.get('map')");
       if (sessions.hasSelectedSessions()) return;
-      if ($scope.firstLoad || hasChangedProgrammatically) {
+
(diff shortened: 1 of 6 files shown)


app/javascript/elm/src/Main.elm
```diff
@@ -59,6 +59,8 @@ type alias Model =
     , resetIcon : String
     , heatMapThresholds : WebData HeatMapThresholds
     , isStreaming : Bool
+    , isSearchOn : Bool
```
3v0k4: Have you considered using the full feature name? That means `isSearchAsIMoveOn`. It's a bit confusing with just `isSearchOn`

app/javascript/elm/tests/MainTests.elm
```diff
@@ -984,4 +992,11 @@ updateTests =
                     |> update (UpdateHeatMapMaximum int)
                     |> Tuple.first
                     |> Expect.equal expected
+        , fuzz bool "ToggleIsSearchOn changes the state of model.isSearchOn" <|
```
3v0k4: Have you considered not testing at all? Also, I see that you haven't tested `update MapMoved`. Any reasons for that?

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -48,12 +48,24 @@ export const SessionsListCtrl = (
     ({ hasChangedProgrammatically }) => {
       console.log("watch - params.get('map')");
       if (sessions.hasSelectedSessions()) return;
-      if ($scope.firstLoad || hasChangedProgrammatically) {
-        sessions.fetch({ amount: params.get("fetchedSessionsCount") });
-      } else {
-        sessions.fetch();
+      if (hasChangedProgrammatically === undefined) return;
```
3v0k4: Let's have a chat about this. Not sure I follow what's going on

app/javascript/elm/tests/MainTests.elm
AkkarinForest: It just feels wrong - not testing at all. I know we discussed overtesting, but I feel uncomfortable not adding any tests when I add new logic. Probably because a few weeks ago when we were adding staff in angular you always suggested that I should do more testing 😄

I understand why it seems like the tests are unnecessary right now. But do you remember how it was joining aircasitng when frontend didn't have any testing at all? 
And if we don't add this test cases now, then no one is gonna add them any time soon, so if the project will grow in a few years it might get to a similar situation as before?

TLDR - I'm experimenting and trying to find a balance 😉

app/javascript/elm/src/Main.elm
gitjul: or `isSearchAsIMove`

app/javascript/elm/tests/MainTests.elm
3v0k4: I think the Angular case is a totally different one (Ruby and Rails too for that matter). TLDR: no type system and no managed runtime like in Elm.

I'm fine with having different opinions on this. As long as we can back them up with valid arguments ;) Let's just test mindfully and not by trying to cover everything possible

PR #190, opened 2019-05-09
Fix autocomplete when pressing enter

3v0k4:
Before this if the user filled "aa" and with the keyboard they selected
"aaBB" the app would have selected "aa".

### DIFF
--- a/app/javascript/angular/code/filtersUtils.js
+++ b/app/javascript/angular/code/filtersUtils.js
@@ -59,23 +59,15 @@ export const setupTimeRangeFilter = (callback, timeFrom, timeTo) => {
 
 export const setupAutocomplete = (callback, id, path) => {
   if (document.getElementById(id)) {
-    $("#" + id)
-      .bind("keydown", function(event) {
-        if (event.keyCode === $.ui.keyCode.ENTER) {
-          $(this)
-            .data("autocomplete")
-            .close(event);
-        }
-      })
-      .autocomplete({
-        source: function(request, response) {
-          const data = { q: request.term, limit: 10 };
-          $.getJSON(path, data, response);
-        },
-        select: function(event, ui) {
-          callback(ui.item.value);
-        }
-      });
+    $(`#${id}`).autocomplete({
+      source: function(request, response) {
+        const data = { q: request.term, limit: 10 };
+        $.getJSON(path, data, response);
+      },
+      select: function(event, ui) {
+        callback(ui.item.value);
+      }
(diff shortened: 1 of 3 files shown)


app/javascript/angular/code/filtersUtils.js
```diff
@@ -59,23 +59,15 @@ export const setupTimeRangeFilter = (callback, timeFrom, timeTo) => {
 
 export const setupAutocomplete = (callback, id, path) => {
   if (document.getElementById(id)) {
-    $("#" + id)
-      .bind("keydown", function(event) {
```
3v0k4: This was the core of the problem

app/javascript/elm/src/LabelsInput.elm
```diff
@@ -51,11 +51,16 @@ update msg model toCmd =
             ( withCandidate str model, Cmd.none )
 
         Add str ->
-            let
-                newModel =
-                    add str model
-            in
-            ( newModel, toCmd (asList newModel) )
+            case str of
```
3v0k4: By removing the code above the input sends empty string so we need to guard that

3v0k4: Could not find a better way :(

app/javascript/angular/code/filtersUtils.js
gitjul: cool, so the library handles that on it's own

app/javascript/elm/src/LabelsInput.elm
AkkarinForest: Yeah, I remember having that problem as well. I guess we could consider creating a custom type and having `List Label`.

app/javascript/angular/code/filtersUtils.js
3v0k4: It seems so

app/javascript/elm/src/LabelsInput.elm
3v0k4: Something like a nonempty string?

3v0k4:
I thought about it but then a glitch in loading the autocomplete list would totally prevent users from inputting anything. Could you bring it to the call if you feel strong about that @gitjul please?

gitjul:
yeah. I wouldn't assume it would prevent users from doing stuff :) enter is easy to press. saw the clearing behaviour in another app, I think it's better in terms of visualising applied filters (no non-existent stuff appears)

PR #191, opened 2019-05-09
ui: style single session action buttons, refactor styles

gitjul:
Besides styling the action button section in single session this PR contains a biggish refactor and some small improvements.

- refactored buttons
- extracted and cleaned up many components
- introduced BEM-style naming in various places
- added missing separator above toggle more parameters/sensors button in
  filter popup and some space around it
- I hope nothing broke :) but if so it will come up and I'll fix it

- [ ] TODO: need to add back missing underline on hover over buttons in "more" list
  (which don't look like buttons) ---> I'll do it when I tweak the popup in general, there are some changes to do there


app/javascript/elm/src/Main.elm
```diff
@@ -565,8 +565,8 @@ view model =
         , main_
             []
             [ div [ class "maps-page-container" ]
-                [ div [ class "map-filters" ]
-                    [ viewSessionTypes model
+                [ div [ class "filters" ]
+                    [ viewSessionTypeNav model
```
3v0k4: The `view` function is huge and a bit messy. That's my bad when unifying the Elm apps into 1. Whenever you can please make it better like in this case. I guess a priority would be to extract a lot of functions out

gitjul:
> There's something in BEM that makes it look really tidy ;)

I agree, it totally really grew on me. makes things easier to find and more modular. not always easy to do, it kinda evolves the more I refactor, and don't expect things to be perfectly categorized :) it's easier when you do it when creating components (here would be view functions, it would go nicely with extracting parts of the `view`), you can try doing that when you create a brand new one if you want :)

PR #192, opened 2019-05-10
Search as I move fix

AkkarinForest:
Map markers and crowd map need to refresh on zooming when "search as I move" is turned off.

### DIFF
--- a/app/javascript/angular/code/services/_fixed_sessions.js
+++ b/app/javascript/angular/code/services/_fixed_sessions.js
@@ -22,6 +22,7 @@ export const fixedSessions = (
     this.scope = $rootScope.$new();
     this.scope.params = params;
     this.fetchableSessionsCount = 0;
+    this.type = "FixedSessions";
   };
 
   let prevMapPosition = {};
@@ -91,7 +92,9 @@ export const fixedSessions = (
       if ($window.location.pathname !== constants.fixedMapRoute) return;
 
       this.drawSessionsInLocation();
-      this.fetchableSessionsCount = fetchableSessionsCount;
+      if (fetchableSessionsCount) {
+        this.fetchableSessionsCount = fetchableSessionsCount;
+      }
       sessionsUtils.onSessionsFetch(this);
     },
 
--- a/app/javascript/angular/code/services/_sessions_utils.js
+++ b/app/javascript/angular/code/services/_sessions_utils.js
@@ -97,6 +97,15 @@ export const sessionsUtils = (
     createSessionData(session, data);
     session.loaded = true;
     callback();
+  },
+  refreshMapView: function(sessions) {
+    if (sessions.type === "MobileSessions") {
+      sessions.onSessionsFetchWithCrowdMapLayerUpdate();
+    } else if (sessions.type === "FixedSessions") {
+      sessions.onSessionsFetch();
+    } else {
+      console.warn("Incorrect sessions type");
+    }
   }
 });
 
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -262,7 +262,7 @@ update msg model =
                 selectedSensorId =
                     Sensor.idForParameterOrLabel value model.selectedSensorId model.sensors
             in
-            ( { model | selectedSensorId = selectedSensorId, selectedSession = NotAsked }
+            ( { model | selectedSensorId = selectedSensorId }
             , Cmd.batch
                 [ Ports.selectSensorId selectedSensorId
                 , fetchHeatMapThresholds model.sensors selectedSensorId
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -56,7 +56,7 @@ type alias Model =
     , resetIcon : String
     , heatMapThresholds : WebData HeatMapThresholds
     , isStreaming : Bool
-    , isSearchOn : Bool
+    , isSearchAsIMoveOn : Bool
     , wasMapMoved : Bool
     }
 
@@ -85,7 +85,7 @@ defaultModel =
     , linkIcon = ""
     , resetIcon = ""
     , heatMapThresholds = NotAsked
-    , isSearchOn = False
+    , isSearchAsIMoveOn = False
     , wasMapMoved = False
     }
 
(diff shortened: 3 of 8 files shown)


app/javascript/elm/src/Main.elm
```diff
@@ -612,8 +612,8 @@ view model =
 
 
 viewSearchAsIMove : Bool -> Bool -> Html Msg
-viewSearchAsIMove wasMapMoved isSearchOn =
-    div []
+viewSearchAsIMove wasMapMoved isSearchAsIMoveOn =
+    div [ Html.Attributes.style "position" "absolute", Html.Attributes.style "z-index" "99999" ]
```
AkkarinForest: This as temporary. @gitjul I thought it's better if you can improve and move this to proper css file when styling,  cause I didn't want to mess with the css structure 😉

gitjul: Perfect:) feel free to leave comments for me in the code if you prefer

app/javascript/angular/code/services/_fixed_sessions.js
```diff
@@ -91,7 +92,9 @@ export const fixedSessions = (
       if ($window.location.pathname !== constants.fixedMapRoute) return;
 
       this.drawSessionsInLocation();
-      this.fetchableSessionsCount = fetchableSessionsCount;
+      if (fetchableSessionsCount) {
+        this.fetchableSessionsCount = fetchableSessionsCount;
```
gitjul: or `fetchableSessionsCount || 0` if you prefer (I like the shortness but don't have strong preferences)

app/javascript/angular/code/services/_sessions_utils.js
```diff
@@ -97,6 +97,15 @@ export const sessionsUtils = (
     createSessionData(session, data);
     session.loaded = true;
     callback();
+  },
+  refreshMapView: function(sessions) {
```
gitjul: how about using an arrow function?

app/javascript/angular/code/services/_sessions_utils.js
```diff
@@ -97,6 +97,15 @@ export const sessionsUtils = (
...
+    if (sessions.type === "MobileSessions") {
```
gitjul: wdyt about using a `case` statement?

app/javascript/angular/code/services/_sessions_utils.js
```diff
@@ -97,6 +97,15 @@ export const sessionsUtils = (
...
+      sessions.onSessionsFetchWithCrowdMapLayerUpdate();
+    } else if (sessions.type === "FixedSessions") {
+      sessions.onSessionsFetch();
+    } else {
+      console.warn("Incorrect sessions type");
```
gitjul: when can this happen?

app/javascript/elm/tests/MainTests.elm
```diff
@@ -887,11 +887,11 @@ updateTests =
                     |> update (UpdateHeatMapMaximum int)
                     |> Tuple.first
                     |> Expect.equal expected
-        , fuzz bool "ToggleIsSearchOn changes the state of model.isSearchOn" <|
+        , fuzz bool "ToggleIsSearchOn changes the state of model.isSearchAsIMoveOn" <|
             \onOrOff ->
-                { defaultModel | isSearchOn = onOrOff }
+                { defaultModel | isSearchAsIMoveOn = onOrOff }
```
gitjul: not sure why but `onOrOff` warms my heart 😍 but above you have `isSearchAsIMoveOn` in a similar test, I'd do this or that in both places

app/javascript/angular/code/services/_fixed_sessions.js
AkkarinForest: That wouldn't be the same effect? I don't want to set it to 0, I want to preserve the original value that's in `this.fetchableSessionsCount`

app/javascript/angular/code/services/_sessions_utils.js
AkkarinForest: Well, this should never happen. But I don't feel comfortable with:
```
    if (sessions.type === "MobileSessions") {
      sessions.onSessionsFetchWithCrowdMapLayerUpdate();
    } else {
      sessions.onSessionsFetch();
    } 
```
when it's clear that the second case should only happen for `(sessions.type === "FixedSessions")`

app/javascript/angular/code/services/_fixed_sessions.js
gitjul: ah, ok, I just saw the 0 default value above. so I guess it can be changed somewhere in between, ok

app/javascript/angular/code/services/_sessions_utils.js
AkkarinForest: That's a good point. I guess I'll change the whole file then.

app/javascript/angular/code/services/_sessions_utils.js
gitjul: I don't mind checking for the other type, but if there's no other option, don't you think it's misleading cos it suggests there is?

app/javascript/angular/code/services/_sessions_utils.js
gitjul: Rysiek mentioned doing that later too, so nice that you wanna take care of it right away:)

app/javascript/angular/code/services/_sessions_utils.js
AkkarinForest: But if there is no default branch, that could lead to silent errors if eg the type is misspelled or null. And it's so hard to track what's going on in this frontend, I don't want to be searching for a cause of some unexpected behavior in the future.

How would you write this?

app/javascript/angular/code/services/_sessions_utils.js
AkkarinForest: I don't mind either way, but I noticed it's never used in the code base and I'm not sure why. We have `else if`'s everywhere.

app/javascript/angular/code/services/_sessions_utils.js
AkkarinForest: Turns out it's not that easy and breaks some test, so I'm gonna stick with non-arrow functions 🙁

3v0k4:
👁

PR #193, opened 2019-05-10
Refactor mobile sessions endpoint

3v0k4:
- Use form object
- New route
- Integration test

### DIFF
--- a/app/controllers/api/measurement_sessions_controller.rb
+++ b/app/controllers/api/measurement_sessions_controller.rb
@@ -18,25 +18,12 @@
 
 module Api
   class MeasurementSessionsController < BaseController
-    INT_Q_ATTRS = [:time_from, :time_to, :day_from, :day_to, :limit, :offset]
-
     # TokenAuthenticatable was removed from Devise in 3.1
     # https://gist.github.com/josevalim/fb706b1e933ef01e4fb6
     before_filter :authenticate_user_from_token!, :only => :create
     before_filter :authenticate_user!, :only => :create
 
     respond_to :json
-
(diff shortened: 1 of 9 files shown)


app/services/api/to_mobile_sessions_array.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def limit
+    data[:limit]
+  end
+
+  def offset
+    data[:offset]
+  end
+end
```
AkkarinForest: This file looks very similar to `to_dormant_sessions_array`, maybe you can extract some duplication?

app/controllers/api/measurement_sessions_controller.rb
```diff
@@ -27,17 +27,6 @@ class MeasurementSessionsController < BaseController
 
     respond_to :json
 
-    def index
-      data = decoded_query_data(params[:q])
-      INT_Q_ATTRS.each { |key| data[key] = data[key].to_i if data.key?(key) }
```
AkkarinForest: You can also delete the `INT_Q_ATTRS ` array on line 21.

app/services/api/to_mobile_sessions_array.rb
3v0k4: Prolly the pagination is similar. I believe that the duplication is not that bad. Do you have any proposals?

app/controllers/api/measurement_sessions_controller.rb
3v0k4: Good catch!

PR #194, opened 2019-05-13
Highlight corresponding marker

AkkarinForest:
Draws a pointer on the corresponding marker when hovering on the session tile

### DIFF
--- a/app/assets/stylesheets/modules/markers.css.scss
+++ b/app/assets/stylesheets/modules/markers.css.scss
@@ -144,3 +144,25 @@
   height: 0px;
   width: 100px;
 }
+
+#highlightMarkerLayer img {
+  animation: pulse 0.7s infinite alternate;
+  -webkit-animation: pulse 0.7s infinite alternate;
+  transform-origin: center;
+  -webkit-transform-origin: center;
+}
+
+/* Animaitons */
+keyframes pulse {
+  to {
+    transform: scale(0.23);
+    -webkit-transform: scale(0.23);
+  }
+}
+
+@-webkit-keyframes pulse {
+  to {
+    transform: scale(0.23);
+    -webkit-transform: scale(0.23);
+  }
+}
--- a/app/assets/stylesheets/modules/markers.css.scss
+++ b/app/assets/stylesheets/modules/markers.css.scss
@@ -152,7 +152,7 @@
   -webkit-transform-origin: center;
 }
 
-/* Animaitons */
+/* Animations */
 keyframes pulse {
   to {
     transform: scale(0.23);
--- a/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
+++ b/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
@@ -1,5 +1,5 @@
 import _ from "underscore";
-import { formatSessionForList } from "../values/session";
+import { formatSessionForList, startingLatLngLiteral } from "../values/session";
 
 export const SessionsListCtrl = (
   $scope,
@@ -11,7 +11,8 @@ export const SessionsListCtrl = (
   drawSession,
   markerSelected,
   updateCrowdMapLayer,
-  sessionsUtils
+  sessionsUtils,
+  map
 ) => {
   let sessions;
   let firstLoad = true;
@@ -170,6 +171,19 @@ export const SessionsListCtrl = (
 
       elmApp.ports.fetchSessions.subscribe(() => {
         sessions.fetch();
+      });
+
+      elmApp.ports.highlightSessionMarker.subscribe(id => {
+        const session = sessionsUtils.find(sessions, id);
+
+        if (session === undefined) {
+          $scope.highlightedSessionMarker.setMap(null);
+          return;
+        }
+
+        $scope.highlightedSessionMarker = map.drawMarker({
+          position: startingLatLngLiteral(session, sensors.selectedSensorName())
+        });
       });
     });
   }
--- a/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
+++ b/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
@@ -1,5 +1,5 @@
 import _ from "underscore";
-import { formatSessionForList, startingLatLngLiteral } from "../values/session";
+import { formatSessionForList } from "../values/session";
 
 export const SessionsListCtrl = (
   $scope,
@@ -86,8 +86,16 @@ export const SessionsListCtrl = (
   };
 
   $scope.newSessionsForList = function() {
-    return $scope.sessions.get().map(formatSessionForList);
+    return $scope.sessions
+      .get()
+      .map($scope.selectedStream)
+      .map(formatSessionForList);
   };
+
+  $scope.selectedStream = session => ({
+    ...session,
+    selectedStream: session.streams[sensors.selectedSensorName()]
+  });
 
   $scope.$watch(
     "newSessionsForList()",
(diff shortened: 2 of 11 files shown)


app/javascript/angular/code/services/google/_map.js.erb
```diff
@@ -220,7 +221,27 @@ export const map = (
...
+        icon: {
+          anchor: new google.maps.Point(30, 31),
+          size: new google.maps.Size(60, 60),
+          scaledSize: new google.maps.Size(60, 60),
+          url: eval(`locationMarkerPath`)
+        }
+      });
+
+      let myoverlay = new google.maps.OverlayView();
+      myoverlay.draw = function() {
```
3v0k4: Have you considered naming it `overlay`?

app/javascript/angular/code/services/google/_map.js.erb
```diff
@@ -220,7 +221,27 @@ export const map = (
...
 
-    clusterMarkers: function(onClick) {      
+    drawHighlightMarker: function(position) {
+      const highlightMarker = this.drawMarker({
+        position: position,
+        icon: {
+          anchor: new google.maps.Point(30, 31),
+          size: new google.maps.Size(60, 60),
+          scaledSize: new google.maps.Size(60, 60),
+          url: eval(`locationMarkerPath`)
```
3v0k4: Does this work with just the string w/o `eval`?

app/javascript/angular/code/services/google/_map.js.erb
```diff
@@ -220,7 +221,27 @@ export const map = (
...
+        }
+      });
+
+      let myoverlay = new google.maps.OverlayView();
+      myoverlay.draw = function() {
+        this.getPanes().markerLayer.id = "markerLayer";
+      };
+      myoverlay.setMap(this.mapObj);
+
+      return highlightMarker
```
3v0k4: Are you using prettier? It seems strange that this does not have `;`

3v0k4: I guess CI should fail too if that needs the semicolon. Can you check what's going on please?

app/javascript/angular/code/services/google/_map.js.erb
```diff
@@ -220,7 +221,27 @@ export const map = (
...
+          anchor: new google.maps.Point(30, 31),
+          size: new google.maps.Size(60, 60),
+          scaledSize: new google.maps.Size(60, 60),
+          url: eval(`locationMarkerPath`)
+        }
+      });
+
+      let myoverlay = new google.maps.OverlayView();
+      myoverlay.draw = function() {
+        this.getPanes().markerLayer.id = "markerLayer";
```
3v0k4: Have you considered a more specific id?

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -171,6 +172,19 @@ export const SessionsListCtrl = (
       elmApp.ports.fetchSessions.subscribe(() => {
         sessions.fetch();
       });
+
+      elmApp.ports.highlightSessionMarker.subscribe(id => {
```
3v0k4: Could we pass in coords from Elm and avoid the dependency on `sessionsUtils`?

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -171,6 +172,19 @@ export const SessionsListCtrl = (
...
+        const session = sessionsUtils.find(sessions, id);
+
+        if (session === undefined) {
+          $scope.highlightedSessionMarker.setMap(null);
```
3v0k4: What if we used a variable like `firstLoad` instead of the scope for the marker?

app/javascript/elm/src/Main.elm
```diff
@@ -354,6 +355,7 @@ update msg model =
                     , Cmd.batch
                         [ Ports.toggleSession { deselected = Nothing, selected = Just id }
                         , SelectedSession.fetch model.sensors model.selectedSensorId model.page id (RemoteData.fromResult >> GotSession)
+                        , Ports.highlightSessionMarker Nothing
```
3v0k4: Could you explain why this and the next are needed? Shouldn't they disappear onMouseLeave no matter what?

app/javascript/angular/code/services/google/_map.js.erb
AkkarinForest: Prettier doesn't catch `js.erb` files.

app/javascript/elm/src/Main.elm
AkkarinForest: When a user clicks on the tile to select the session the tile disappears but the `onMouseLeave` event is not triggered by this.

app/javascript/angular/code/services/google/_map.js.erb
3v0k4: Aha good point. This no worries then, https://llp.kanbanery.com/projects/8040/board/tasks/2407976 will solve all problems

app/assets/stylesheets/modules/markers.css.scss
```diff
@@ -144,3 +144,25 @@
...
 }
+
+#highlightMarkerLayer img {
+  animation: pulse 0.7s infinite alternate;
+  -webkit-animation: pulse 0.7s infinite alternate;
+  transform-origin: center;
+  -webkit-transform-origin: center;
+}
+
+/* Animaitons */
```
gitjul: *animations :)

app/javascript/angular/code/services/_sessions_utils.js
```diff
@@ -97,6 +97,15 @@ export const sessionsUtils = (
...
     session.loaded = true;
     callback();
+  },
+  refreshMapView: function(sessions) {
+    if (sessions.type === "MobileSessions") {
+      sessions.onSessionsFetchWithCrowdMapLayerUpdate();
+    } else if (sessions.type === "FixedSessions") {
+      sessions.onSessionsFetch();
+    } else {
+      console.warn("Incorrect sessions type");
```
gitjul: `session type` or `sessions'` type or `type of sessions`

app/javascript/angular/code/values/session.js
```diff
@@ -34,6 +34,13 @@ export const startingLatLng = (session, selectedSensor) => {
   };
 };
 
+export const startingLatLngLiteral = (session, selectedSensor) => {
+  return {
+    lat: startingLat(session, selectedSensor),
+    lng: startingLng(session, selectedSensor)
+  };
+};
+
```
gitjul: if the function only `return`s something you can write it like this:

```
export const startingLatLngLiteral = (session, selectedSensor) => ({
  lat: startingLat(session, selectedSensor),
  lng: startingLng(session, selectedSensor)
});
```
(or however you prefer to put the {}, honestly I don't even remember now what I usually do :D)

app/javascript/elm/src/Main.elm
```diff
@@ -612,8 +618,8 @@ view model =
 
 
 viewSearchAsIMove : Bool -> Bool -> Html Msg
-viewSearchAsIMove wasMapMoved isSearchOn =
-    div []
+viewSearchAsIMove wasMapMoved isSearchAsIMoveOn =
+    div [ Html.Attributes.style "position" "absolute", Html.Attributes.style "z-index" "99999" ]
```
gitjul: this already changed on experimental I guess there will be a conflict here, the new version is 
`div [ class "search-control" ]`

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -84,9 +86,17 @@ export const SessionsListCtrl = (
...
 
   $scope.newSessionsForList = function() {
-    return $scope.sessions.get().map(formatSessionForList);
+    return $scope.sessions
+      .get()
+      .map($scope.selectedStream)
+      .map(formatSessionForList);
   };
 
+  $scope.selectedStream = session => ({
```
3v0k4: Have you considered not having this on the `$scope`? Could be a vanilla arrow function

PR #196, opened 2019-05-13
Style "search as I move" and "search again" button

gitjul:
We're not replacing the default map controls now so I made the search as I move look similar to the controls and made the "Search again" button stand out by applying a dark blue. This is loosely based on the designs.

Looks like this right now:

Screenshot 2019-05-14 at 15 44 50

Screenshot 2019-05-14 at 15 44 42


3v0k4:
Cool!

PR #199, opened 2019-05-15
Disabled state for inputs

gitjul:
In order to target labels and set them to 40% opacity like the inputs as imagined by Pina the structure was changed so that labels are after inputs I used flexbox + changed order of elements + `column-reverse`. It's not the default way of doing things, usually labels are above inputs in html, but I don't think it should cause accessibility issues (labels are connected to inputs with `for`). And it's a neat trick if I may say so :P 

I also refactored some styles, mainly extracted tag styles and changed class names.

I'm not sure if the difference is clearly visible, but we'll test.

Screenshot 2019-05-15 at 13 54 14

### DIFF
--- a/app/assets/stylesheets/modules/session-type-nav.css.scss
+++ b/app/assets/stylesheets/modules/session-type-nav.css.scss
@@ -11,6 +11,7 @@
   font-weight: $font-stack-bold;
   font-size: $small-font;
   padding: $margin-default;
+  position: relative;
   text-align: left;
   width: 50%;
   z-index: 1;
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/modules/session-type-nav.css.scss
```diff
@@ -24,4 +25,9 @@
   &:focus {
     outline: none;
   }
+
+  [data-tippy-content] {
```
gitjul: this will be moved when tips are styled, which will be soon

PR #201, opened 2019-05-15
Upgrade Webpacker

3v0k4:
I've found some problems importing the svgs with file-loader and the tests. Hopefully this will make my life easier making that happen.

Followed https://github.com/rails/webpacker/blob/master/docs/v4-upgrade.md


Gemfile
```diff
@@ -31,7 +31,7 @@ gem 'uglifier'
 gem 'sass-rails'
 gem 'coffee-rails'
 gem 'activesupport-json_encoder'
-gem 'webpacker', git: 'https://github.com/rails/webpacker.git', branch: '3-x-stable'
+gem 'webpacker', '~> 4.0.2'
```
gitjul: 🙌

gitjul:
thanks for taking care of this!

do you have a good reference you're using when configuring webpack, babel etc.? I haven't learned my way around stuff like that yet and usually have trouble finding out how to do what I need

3v0k4:
@gitjul unfortunately not. I know more or less how things are structured but I end up googling a lot every time

PR #202, opened 2019-05-15
UI: tooltips

gitjul:
- style tooltip icons
- use tippy "light-border" theme

Screenshot 2019-05-16 at 15 08 37


app/javascript/elm/src/TimeRange.elm
```diff
@@ -1,7 +1,7 @@
 module TimeRange exposing (TimeRange(..), defaultTimeRange, update, view)
 
-import Html exposing (Html, button, div, h4, input, label, text)
-import Html.Attributes exposing (attribute, class, disabled, for, id, name, placeholder, type_)
+import Html exposing (Html, button, div, h4, img, input, label, text)
+import Html.Attributes exposing (alt, attribute, class, disabled, for, id, name, placeholder, src, type_)
```
AkkarinForest: Do you need these in this file?

app/javascript/elm/tests/LabelsInputTests.elm
```diff
@@ -47,7 +47,7 @@ all =
                     labels =
                         LabelsInput.fromList [ label ]
                 in
-                LabelsInput.view labels "description" "input-id" "placeholder" False Tooltip.profilesFilter
+                LabelsInput.view labels "description" "input-id" "placeholder" False Tooltip.profilesFilter "tooltip-icon"
```
AkkarinForest: Why some are `tooltip-icon` and some `tooltip-icon.svg`? I guess it doesn't change anything, but it might be confusing.

app/javascript/packs/elm.js
```diff
@@ -2,6 +2,7 @@ import { Elm } from "../elm/src/Main.elm";
 import logoNav from "../../assets/images/aircasting-logo-nav.svg";
 import linkIcon from "../../assets/images/link-icon.svg";
 import resetIcon from "../../assets/images/reset-icon.svg";
```
AkkarinForest: Is there no way to load the image in elm directly? It would probably simplify a lot. And we should figure it out at some point if we will move more logic to elm and need to load there more images.

app/javascript/elm/src/Main.elm
```diff
@@ -142,6 +145,7 @@ init flags url key =
         , logoNav = flags.logoNav
         , linkIcon = flags.linkIcon
         , resetIcon = flags.resetIcon
+        , tooltipIcon = flags.tooltipIcon
```
3v0k4: One thing we could do is

```
type Path = Path String

type alias Model =
  { ...
  , tooltipIcon : Path
  }
```

And here `tooltipIcon = Path flags.tooltipIcon`. That way we can't pass a string when the function needs a `Path`. Then you would need a function like

```
path : Path -> String
path (Path str) = str
```

app/javascript/elm/tests/LabelsInputTests.elm
gitjul: by chance, thx for pointing out

app/javascript/elm/src/TimeRange.elm
gitjul: no! thx:)

app/javascript/elm/src/Main.elm
gitjul: good idea, and it will be easily distinguishable from other strings among function attribute types. I didn't like adding another String cos it looked random and it isn't. thx for showing it, I'll refactor but don't know if right away

gitjul:
@3v0k4 I need help figuring out how to fix js tests. I found how you'd go around doing it if we used `jest`: https://jestjs.io/docs/en/webpack#mocking-css-modules
but I can't find anything like that for `blue-tape`, and I'm not a webpack configuration genius...

app/javascript/packs/elm.js
gitjul: I guess this was answered offline

gitjul:
maybe a fix would look sth like this https://www.codementor.io/daveschinkel13/unexpected-token-errors-related-to-css-in-mocha-cdin9300j

3v0k4:
I don't have time to look into that. To work around a similar problem with importing SVG I've done https://github.com/HabitatMap/AirCasting/blob/experimental/app/javascript/assets.js. Maybe that helps? Otherwise, ping me tomorrow!

gitjul:
@3v0k4 yeah tomorrow, I want to leave soon too

gitjul:
note to self, this way seems cool: https://medium.com/@a_eife/unit-test-code-that-uses-css-modules-ef5b49efc707

app/javascript/elm/src/Main.elm
gitjul: @3v0k4 https://github.com/HabitatMap/AirCasting/pull/202/commits/bd12153191ed06fd13118e76b12b0c80da7ac5fb let me know if you have any comments

PR #203, opened 2019-05-16
ui(session card): truncate title and add ellipsis

3v0k4:
Screen Shot 2019-05-16 at 11 42 51

@gitjul seems like the width is not needed. In case you think we should add it please feel free to add commits to this PR cause I'd be clueless on how much that should be


gitjul:
@3v0k4 cool! looks good, nice that width isn't needed, must use a width from a parent

PR #204, opened 2019-05-16
Timerange yy instead of yyyy

3v0k4:
Just two digits for the timerange input

Screen Shot 2019-05-16 at 12 02 09 1

### DIFF
--- a/app/javascript/angular/code/filtersUtils.js
+++ b/app/javascript/angular/code/filtersUtils.js
@@ -26,25 +26,27 @@ export const oneHourAgo = () =>
     .subtract(1, "hour")
     .format("X");
 
+export const daterangepickerConfig = (timeFrom, timeTo) => ({
+  linkedCalendars: false,
+  timePicker: true,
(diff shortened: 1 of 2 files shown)


app/javascript/angular/code/filtersUtils.js
```diff
@@ -26,25 +26,27 @@ export const oneHourAgo = () =>
...
+  startDate: moment
+    .unix(timeFrom)
+    .utc()
+    .format("MM/DD/YY HH:mm"),
+  endDate: moment
+    .unix(timeTo)
+    .utc()
+    .format("MM/DD/YY HH:mm"),
+  locale: {
+    format: "MM/DD/YY HH:mm"
```
AkkarinForest: maybe we can extract `const format = "MM/DD/YY HH:mm"`?

3v0k4: Good point. But I'll leave it for later ;)

PR #205, opened 2019-05-16
Display statistics for the selected graph range

AkkarinForest:
Getting the values for fixed is easier, cause we only download the measurements that are displayed on the graph.
For mobile I had to do some fancy filtering.

Also, it has to be done into places:
- everytime someone selects a range button, scrolls the graph, selects a graph region - the afterSetExtremes is triggered and there new stats are calculated
- unfortunately, this is not triggered when first loading the graph, so there I had to figure out a different way to filter the measurements, and for mobile it's a bit longer, but it's the best I could figure out.

### DIFF
--- a/app/javascript/angular/code/services/graph.js
+++ b/app/javascript/angular/code/services/graph.js
@@ -11,7 +11,12 @@ let measurementsByTime = {};
 let chart = null;
 const RENDER_TO_ID = "graph";
 
-export const fetchAndDrawFixed = ({ sensor, heat, times, streamId }) => {
+export const fetchAndDrawFixed = showStatsCallback => ({
+  sensor,
+  heat,
+  times,
+  streamId
+}) => {
   // render empty graph with loading message
   drawFixed({
     measurements: [],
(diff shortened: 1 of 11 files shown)


app/javascript/elm/src/Data/SelectedSession.elm
```diff
@@ -122,23 +118,40 @@ view session heatMapThresholds linkIcon toMsg =
...
-                ]
-            , div [ class "single-min-max-container" ]
-                [ div [ class "single-session-color", class <| Data.Session.classByValue (Just session.max) heatMapThresholds ] []
-                , span [] [ text "max. " ]
-                , span [ class "single-session-max" ] [ text <| String.fromFloat session.max ]
-                ]
-            ]
+        , case session.selectedRangeMeasurements of
+            [] ->
+                div [] []
```
AkkarinForest: @gitjul How can I style this so that it's the same size as the div on line 136?

app/javascript/angular/code/services/graph.js
```diff
@@ -200,3 +238,26 @@ const draw = ({
     .getElementById(RENDER_TO_ID)
     .addEventListener("mouseleave", graphHighlight.hide);
 };
+
+const getValues = data => data.map(m => m.value);
+
+const filterMeasurements = measurements => {
+  measurements = Object.values(measurements);
+  const diff = mobileButtons[0][mobileButtons[1]];
```
AkkarinForest: that's how the default button for the selected time range is defined

app/javascript/elm/src/Data/SelectedSession.elm
```diff
@@ -122,23 +129,40 @@ view session heatMapThresholds linkIcon toMsg =
...
-                ]
-            , div [ class "single-min-max-container" ]
-                [ div [ class "single-session-color", class <| Data.Session.classByValue (Just session.max) heatMapThresholds ] []
-                , span [] [ text "max. " ]
-                , span [ class "single-session-max" ] [ text <| String.fromFloat session.max ]
-                ]
-            ]
+        , case session.selectedMeasurements of
+            [] ->
+                div [ class "single-session-placeholder" ] []
```
gitjul: what's this for?

AkkarinForest: I needed an element that I can style, so it's the same size as the div from the other case branch. Otherwise, the neighboring elements were changing positions.

app/javascript/elm/src/Data/SelectedSession.elm
gitjul: is this still an issue?

AkkarinForest: No, I styled this (but not responsive).

PR #206, opened 2019-05-16
Use one tooltip instance

AkkarinForest:
There was a bug when a user opened a tooltip window and clicked again on the button to open a tooltip window. That second click would try to create a second tooltip instance using the same id as the first instance. That would return a null instead of creating an instance and the SetContent would throw an error when executed on null.

### DIFF
--- a/app/javascript/angular/code/filtersUtils.js
+++ b/app/javascript/angular/code/filtersUtils.js
@@ -79,15 +79,17 @@ export const setupClipboard = () => {
   new Clipboard("#copy-link-button");
 };
 
-export const tooltipInstance = tooltipId =>
+export const tooltipInstance = (tooltipId, oldTooltip) =>
   tippy(`#${tooltipId}`, {
     animateFill: false,
     interactive: true,
     trigger: "manual"
-  })[0];
+  })[0] || oldTooltip;
+
+let tooltip;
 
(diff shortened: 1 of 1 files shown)


app/javascript/angular/code/filtersUtils.js
```diff
@@ -79,15 +79,17 @@ export const setupClipboard = () => {
...
-export const tooltipInstance = tooltipId =>
+export const tooltipInstance = (tooltipId, oldTooltip) =>
   tippy(`#${tooltipId}`, {
     animateFill: false,
     interactive: true,
     trigger: "manual"
-  })[0];
+  })[0] || oldTooltip;
+
+let tooltip;
```
3v0k4: I don't like this `tooltip` hanging around. An alternative is to use the [module pattern](https://coryrylan.com/blog/javascript-module-pattern-basics) like

```
const tooltipInstance = (() => {
  let instance;

  return tooltipId => {
    if (!instance) {
      instance = tippy(`#${tooltipId}`, {
        animateFill: false,
        interactive: true,
        trigger: "manual"
      })[0];
    }
    return instance;
  }
})();
```

AkkarinForest: Awesome!

PR #207, opened 2019-05-17
Search as I move behavior

AkkarinForest:


app/javascript/elm/src/Main.elm
```diff
@@ -598,7 +598,12 @@ view model =
 
                           else
                             text ""
-                        , viewSearchAsIMove model.wasMapMoved model.isSearchAsIMoveOn
+                        , case model.selectedSession of
```
3v0k4: It would be awesome to start extracting more functions in the future. Just using this as a pretext.

PR #210, opened 2019-05-17
Don't fetch unnecessary measurements

AkkarinForest:
For long fixed sessions that was causing very long loading times. And these measurements are not used for anything. 
We are doing a separate fetch on the angular side to fetch measurements for mobile session necessary to draw a session on the map.

### DIFF
--- a/app/javascript/angular/code/services/graph.js
+++ b/app/javascript/angular/code/services/graph.js
@@ -108,6 +108,7 @@ const afterSetExtremes = ({ streamId, times, showStatsCallback }) => e => {
         measurements,
         times
       });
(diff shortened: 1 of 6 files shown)


app/javascript/angular/code/services/graph.js
```diff
@@ -108,6 +108,7 @@ const afterSetExtremes = ({ streamId, times, showStatsCallback }) => e => {
         measurements,
         times
       });
+      measurementsByTime = dataByTime;
```
AkkarinForest: this is something I removed in some of my previous changes and I just realized it might be used somewhere so I'm putting it back :D

PR #211, opened 2019-05-17
fix(pulsing marker)

3v0k4:
* First commit moves scss for map to webpack
* Second commit fixes the pulsing marker problem

We can share scss between JS and Rails. In fact, `app/assets/stylesheets/application.css` is used by Rails (see `app/views/layouts/application.html.haml`) and `app/assets/stylesheets/main.scss` is used by JS (see `app/javascript/packs/elm.js`). Prolly we should move files around and rename them. But this should prove that we don't lose flexibility but allow us to keep responsibility separated (ie Rails handling Rails stuff via asset pipeline, JS handling JS stuff via Webpack).

Some stuff that is now in main.scss (JS) should be moved to `application.css` because it's needed in Rails pages

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -3,6 +3,8 @@
  * and any sub-directories. You're free to add application-wide styles to this file and they'll appear at
  * the top of the compiled file, but it's generally better to create a new file per style scope.
  *= require_self
- *= require main
- *= require_tree ./vendor
 */
+
+body {
+  background: pink;
+}
--- a/app/assets/stylesheets/config/jquery-ui-assets.css.scss
+++ b/app/assets/stylesheets/config/jquery-ui-assets.css.scss
@@ -1,33 +1,34 @@
 .ui-icon {
   width: 16px;
   height: 16px;
-  background-image: image-url("ui-icons_222222_256x240.png");
+  background-image: url("../../images/ui-icons_222222_256x240.png");
 }
 .ui-widget-content .ui-icon {
-  background-image: image-url("ui-icons_222222_256x240.png");
+  background-image: url("../../images/ui-icons_222222_256x240.png");
 }
 .ui-widget-header .ui-icon {
-  background-image: image-url("ui-icons_222222_256x240.png");
+  background-image: url("../../images/ui-icons_222222_256x240.png");
 }
 .ui-state-default .ui-icon {
-  background-image: image-url("ui-icons_888888_256x240.png");
+  background-image: url("../../images/ui-icons_888888_256x240.png");
 }
 .ui-state-hover .ui-icon,
 .ui-state-focus .ui-icon {
-  background-image: image-url("ui-icons_454545_256x240.png");
+  background-image: url("../../images/ui-icons_454545_256x240.png");
 }
 .ui-state-active .ui-icon {
-  background-image: image-url("ui-icons_454545_256x240.png");
+  background-image: url("../../images/ui-icons_454545_256x240.png");
 }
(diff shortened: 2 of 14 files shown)


app/assets/stylesheets/application.css
```diff
@@ -3,6 +3,8 @@
  * and any sub-directories. You're free to add application-wide styles to this file and they'll appear at
  * the top of the compiled file, but it's generally better to create a new file per style scope.
  *= require_self
- *= require main
- *= require_tree ./vendor
 */
+
+body {
+  background: pink;
```
3v0k4: Because why not

app/assets/stylesheets/modules/markers.css.scss
```diff
@@ -145,7 +145,7 @@
   width: 100px;
 }
 
-#highlightMarkerLayer img {
+img[src*="pulsing_location_marker"] {
```
3v0k4: Looks for substring. For example it matches 'sssssssspulsing_location_markersssssss'

gitjul: I think "pulsating" is the right word

app/assets/stylesheets/application.css
gitjul: but for what? to test or sth?

gitjul:
👍 

how about renaming the `main.scss` file to `app/assets/stylesheets/elm-application.scss`?

gitjul:
btw I think this will solve my problem with the custom tippy theme not loading on experimental

app/assets/stylesheets/config/jquery-ui-assets.css.scss
```diff
@@ -1,33 +1,34 @@
 .ui-icon {
   width: 16px;
   height: 16px;
-  background-image: image-url("ui-icons_222222_256x240.png");
+  background-image: url("../../images/ui-icons_222222_256x240.png");
```
gitjul: so for styles used by JS the path looks like this but for styles used by Rails it's still image-url? or this works too?

app/assets/stylesheets/application.css
3v0k4: Just left there as a todo. We need to style all that stuff

app/assets/stylesheets/modules/markers.css.scss
3v0k4: I'm using the word that Michael used in the ticket. But I'm fine changing it if you prefer

app/assets/stylesheets/config/jquery-ui-assets.css.scss
3v0k4: This works only with Webpack (https://www.npmjs.com/package/resolve-url-loader). That's why we should prolly move files around and renome

3v0k4:
I'm merging as is. Let's have a chat after the call to talk about moving files / renaming

app/assets/stylesheets/application.css
gitjul: ok. we need to share some of the styles.

app/assets/stylesheets/config/jquery-ui-assets.css.scss
gitjul: ah. so we can't share styles :(

3v0k4: Depends on the case ;)

3v0k4:
@gitjul feel free to move, rename css files depending on what's more comfy to you. If anything is unclear let me know

PR #212, opened 2019-05-20
Graph tooltip position adjustments

AkkarinForest:


app/javascript/angular/code/services/buildGraphOptions.js
```diff
@@ -155,11 +155,18 @@ const buildTooltip = ({
...
-      this.chart.plotWidth - labelWidth // extreme left
+      Math.max(0, point.plotX - (labelWidth - pointerWidth) / 2), // (extreme left, middle)
+      this.chart.plotWidth - labelWidth // extreme right
     );
-    const tooltipY = point.plotY - labelHeight - 10;
+
+    const pointerHeight = 10;
+    let tooltipY = point.plotY - labelHeight - pointerHeight;
+    if (tooltipY < 0) {
+      // changes tooltip position at the top of the graph
```
gitjul: do you mean that it sets the position at the top of the graph?

AkkarinForest: No, I mean that when the tooltip is close to the top of the graph the position is adjusted to prevent it overlapping with the line on the graph

gitjul: cool, could you maybe include the above in the comment?

PR #214, opened 2019-05-20
Refactor: simplify selected session logic

AkkarinForest:
I removed some convoluted logic as part of our process of getting rid of angular :)

### DIFF
--- a/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
+++ b/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
@@ -26,11 +26,6 @@ export const SessionsListCtrl = (
     $window.sessions = sessions = $scope.sessions;
     $scope.sessionsForList = [];
 
-    // prolly this can be removed
-    if (_(params.get("selectedSessionIds", [])).isEmpty()) {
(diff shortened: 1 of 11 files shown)


app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -39,15 +34,15 @@ export const SessionsListCtrl = (
...
-      sessions.hasSelectedSessions()
+      sessionsUtils.isSessionSelected()
     );
   };
 
   $scope.$watch(
     "params.get('map')",
     ({ hasChangedProgrammatically }) => {
       console.log("watch - params.get('map')");
-      if (sessions.hasSelectedSessions()) return;
```
AkkarinForest: The fact that this was working before is MAGIC! `hasSelectedSessions ` was calling a chain of at least 5 other methods and at the end is a function that does 
```_(self.sessions || []).detect(function(session) {
      return session.id === id;
    });
```
But when a page with selected session is first loaded (or refreshed) the `self.sessions` is empty so `sessions.hasSelectedSessions()` returned false event though session was selected 😮

PR #215, opened 2019-05-21
Horizontal scroll for sessions list

3v0k4:
This just works for the mouse wheel. If users have trackpads or such things they should already be able to scroll horizontally.

It uses https://gist.github.com/pablen/c07afa6a69291d771699b0e8c91fe547

### DIFF
--- a/app/javascript/createObserver.js
+++ b/app/javascript/createObserver.js
@@ -1,0 +1,63 @@
+/**
+ * Initializes a DOM mutation observer that will run a hook when
+ * a DOM node matching a selector is mounted or unmounted.
+ *
+ *   const myObs = createObserver({
+ *     selector: "[data-ace-editor]",
+ *     onMount: node => {},
+ *     onUnmount: node => {}
+ *   });
+ */
+
+const DEFAULT_ROOT_ELEMENT = document;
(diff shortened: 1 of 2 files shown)


app/javascript/createObserver.js
```diff
@@ -0,0 +1,63 @@
...
+          const matchingElements = getMatchingElementsFromTree(
+            addedNode,
+            selector
+          );
+          if (matchingElements.length < 1) return;
+          matchingElements.forEach(node => onMount(node));
+        });
+      }
+      // Handle removed nodes
+      if (onUnmount) {
```
AkkarinForest: Maybe we don't need to copy the entire file? eg looks like we are not using this if.

3v0k4: We are not but it could be useful, especially if we use more of that

3v0k4:
Good catch. Haven't noticed that "wheel" is triggered on scroll too. It seems like there actually no good way to do that w/o a lot of hacking. I'll check back with others what to do here. Thanks!

3v0k4:
Checking `deltaX` should have solved the problem mentioned by @AkkarinForest

gitjul:
👀 

thx for taking care of this <3

PR #216, opened 2019-05-21
Deselects session when using filters

AkkarinForest:


app/javascript/elm/src/Main.elm
```diff
@@ -474,6 +506,18 @@ update msg model =
             ( { model | selectedSession = SelectedSession.updateRange model.selectedSession measurements }, Cmd.none )
 
 
+deselectSession : { a | selectedSession : WebData SelectedSession } -> ( { a | selectedSession : WebData SelectedSession }, Cmd Msg )
```
3v0k4: The signature is a bit noisy. Consider either passing model or naming the type with an alias. If you go with the second consider renaming the argument to something different than `model`

app/javascript/elm/tests/MainTests.elm
```diff
@@ -138,7 +138,7 @@ locationFilter =
                 { defaultModel | location = location }
                     |> update SubmitLocation
                     |> Tuple.second
-                    |> Expect.equal (Ports.findLocation location)
```
3v0k4: Personally I'd consider removing all those tests. Not sure what value they add. Also, when they fail the error message is not clear.

PR #219, opened 2019-05-22
Clean up

AkkarinForest:
I removed methods that were suggesting the possibility of selecting multiple sessions and replaced them with some simple methods in sessionsUtils.
It's not really related to any ticket, but I hope it will make it easier to understand what's going on.


app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -22,7 +21,6 @@ export const FixedSessionsMapCtrl = (
     $scope.params = params;
     $scope.sensors = sensors;
     $scope.sessions = fixedSessions;
-    $scope.singleSession = singleFixedSession;
```
AkkarinForest: looks dead

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -26,28 +26,23 @@ export const SessionsListCtrl = (
...
+      !(sessionsUtils.selectedSessionId() === sessionId) &&
+      sessionsUtils.isSessionSelected()
     );
   };
 
   $scope.$watch(
     "params.get('map')",
     ({ hasChangedProgrammatically }) => {
       console.log("watch - params.get('map')");
-      if (sessions.hasSelectedSessions()) return;
```
AkkarinForest: The fact that this was working before is MAGIC! hasSelectedSessions was calling a chain of at least 5 other methods and at the end is a function that does

  find: function(self, id) {
    return _(self.sessions || []).detect(function(session) {
      return session.id === id;
    });
  },
But when a page with selected session is first loaded (or refreshed) the self.sessions is empty so sessions.hasSelectedSessions() returned false event though session was selected 😮

app/javascript/angular/code/calculateBounds.js
```diff
@@ -1,28 +1,16 @@
...
+    return {
+      north: stream.max_latitude,
+      east: stream.max_longitude,
+      south: stream.min_latitude,
+      west: stream.min_longitude
+    };
+  }
 
-  return { north, east, south, west };
+  return { north: -Infinity, east: -Infinity, south: Infinity, west: Infinity };
```
AkkarinForest: not sure when would that be useful, but that's how it was working before: ` Math.min.apply(null, maxLat)` returns  Infinity if `maxLat == []`.

app/javascript/angular/code/services/_fixed_sessions.js
```diff
@@ -95,7 +67,9 @@ export const fixedSessions = (
       if (fetchableSessionsCount) {
         this.fetchableSessionsCount = fetchableSessionsCount;
       }
-      sessionsUtils.onSessionsFetch(this);
+      if (sessionsUtils.isSessionSelected()) {
+        this.reSelectSession(sessionsUtils.selectedSessionId());
+      }
```
AkkarinForest: that's basically what it was doing before, but I prefer if it's here directly, makes it easier to understand and gives us a fighting chance to refactor this in the future 😄

app/javascript/angular/code/services/_fixed_sessions.js
```diff
@@ -236,17 +213,14 @@ export const fixedSessions = (
 
       const data = params.get("data");
 
-      // _.values suggests that `params.get('selectedSessionIds')` could be an obj, is it true?
```
AkkarinForest: oh yes, it could, cause `params.get()` might return `{}`: 
```
      get: function(name, defaultValue) {
        return this.paramsData[name] || defaultValue || {};
      },
```
I would suggest we stop using `params.get` and just create direct methods for things we need, then someday we can just move them from params to some vanilla js.

app/javascript/angular/code/services/_fixed_sessions.js
```diff
@@ -236,17 +213,14 @@ export const fixedSessions = (
...
-      const sessionIds = _.values(params.get("selectedSessionIds") || []);
-
       if (!data.timeFrom || !data.timeTo) return;
 
       var reqData = {
         time_from: data.timeFrom,
         time_to: data.timeTo,
         tags: data.tags,
         usernames: data.usernames,
-        session_ids: sessionIds
```
AkkarinForest: maybe we should also change that endpoint, to only expect one id, but I'll leave this for the future

app/javascript/angular/code/services/_mobile_sessions.js
```diff
@@ -145,7 +120,7 @@ export const mobileSessions = (
       // this must happen before everything else, otherwise the session is drawn on the map and removed right away
       drawSession.clear(this.sessions);
       setTimeout(() => {
-        const callback = (session, allSelected) => data => {
```
AkkarinForest: this argument isn't used here

app/javascript/angular/code/services/_mobile_sessions.js
```diff
@@ -245,7 +219,7 @@ export const mobileSessions = (
           cache: true,
           params: { sensor_name: sensorName }
         })
-        .success(callback(session, allSelected));
```
AkkarinForest: this is a callback to calculate bounds which I adjusted to only expect one id now.

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -26,28 +26,24 @@ export const SessionsListCtrl = (
     $window.sessions = sessions = $scope.sessions;
     $scope.sessionsForList = [];
 
-    // prolly this can be removed
-    if (_(params.get("selectedSessionIds", [])).isEmpty()) {
-      params.update({ selectedSessionIds: [] });
-    }
-
-    sessions.reSelectAllSessions();
+    if (sessionsUtils.isSessionSelected())
```
3v0k4: Is this `if` needed? Cause it seems it wasn't there before

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
3v0k4: The magic of js

app/javascript/angular/code/controllers/note_ctrl.js
```diff
@@ -1,5 +1,5 @@
-function NoteCtrl($scope, note, singleMobileSession) {
-  var session = singleMobileSession.get();
+function NoteCtrl($scope, note, mobileSessions, sessionsUtils) {
+  var session = sessionsUtils.selectedSession(mobileSessions);
```
3v0k4: I guess this is getting `mobileSessions` cause we have notes only for mobile sessions, correct?

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
AkkarinForest: Yes.
Before there was a method `reSelectAllSessions` which had a loop on all selected sessions. So if there was no session selected it didn't execute the code in the loop. Now there are only two possibilities - there is a selected session or not. So it doesn't make sense to loop over a list of selected session - easier to just have this `if`.

app/javascript/angular/code/controllers/note_ctrl.js
AkkarinForest: Yes, our mobile app expert confirmed we don't have notes for fixed sessions.

PR #222, opened 2019-05-24
Deleting notes + params validation

AkkarinForest:


app/models/api/user_sessions.rb
```diff
@@ -0,0 +1,80 @@
...
+            required(:latitude).filled(:int?)
+            required(:longitude).filled(:int?)
+            required(:date).filled(:str?)
+            required(:text).filled(:str?)
+          end
+        end
+        validate(valid_streams: :streams) do |stream|
+          stream.all? do |sensor_name, stream_data|
+            StreamSchema.call(stream_data).success?
+          end
```
AkkarinForest: that's the only way I discovered to validate hash values without knowing the keys (here keys can be anything, cause they are sensor_names).
Apparently, this feature is coming in the 1.0 version of dry-validation: 
https://github.com/dry-rb/dry-schema/issues/37

app/models/api/user_sessions.rb
```diff
@@ -0,0 +1,80 @@
...
+      attribute :longitude, Types::Strict::Float
+      attribute :type, Types::Strict::String
+      attribute :notes, Types::Strict::Array do
+        attribute :number, Types::Strict::Integer
+        attribute :latitude, Types::Strict::Integer
+        attribute :longitude, Types::Strict::Integer
+        attribute :date, Types::Strict::String
+        attribute :text, Types::Strict::String
+      end
+      attribute :streams, Types::Strict::Hash
```
AkkarinForest: Have no idea how to validate the internal of this hash, so this will have to be enough for now.

app/controllers/api/user_sessions_controller.rb
```diff
@@ -25,11 +7,14 @@ class Api::UserSessionsController < Api::BaseController
   respond_to :json
 
   def sync
-    data = JSON.parse(params[:data])
-    data = deep_symbolize(data)
+    form = Api::JsonForm.new(json: "{ \"data\": #{params[:data]} }", schema: Api::UserSessions::Schema, struct: Api::UserSessions::Struct)
```
3v0k4: This is a bit hacky. If it's needed again let's think of an abstraction

app/models/api/user_sessions.rb
3v0k4: Good job!

spec/controllers/api/user_sessions_controller_spec.rb
```diff
@@ -25,19 +7,62 @@
   before { allow(controller).to receive(:current_user) { user } }
 
   describe "#sync" do
-    let(:data) { [{ "something" => "value", :notes => [{ "key" => "value" }] }].to_json }
+    it "returns session for upload when it's not present in the db" do
```
3v0k4: I'd go for `database` instead of `db

spec/controllers/api/user_sessions_controller_spec.rb
```diff
@@ -25,19 +7,62 @@
...
+      post :sync, :format => :json, :data => session_data(uuid: "abc")
+
+      expected = { "download" => [], "upload" => ["abc"], "deleted" => [] }
+
```
3v0k4: I guess we can remove this line

spec/controllers/api/user_sessions_controller_spec.rb
```diff
@@ -25,19 +7,62 @@
...
+      expect(user.sessions.count).to eq(0)
+      expect(json_response).to eq(expected)
+    end
 
-    before do
-      expect(user).to receive(:sync).
-        with([{:something => "value", :notes => [{ :key => "value" }] }]).
-        and_return(["Result"])
+    it "returns session for download when present it's in the db, but not in the mobile app" do
```
3v0k4: Maybe when `present in the db` instead of `present it's in the db`?

spec/controllers/api/user_sessions_controller_spec.rb
```diff
@@ -93,4 +118,86 @@
...
+          \"threshold_medium\":70,
+          \"threshold_very_high\":100,
+          \"threshold_very_low\":20,
+          \"unit_name\":\"decibels\"}
+        },
+      \"tag_list\":\"#{attributes.fetch(:tag_list, "")}\",
+      \"title\":\"#{attributes.fetch(:title, "title")}\",
+      \"type\":\"MobileSession\",
+      \"uuid\":\"#{attributes.fetch(:uuid, "uuid")}\"}
+    ]"
```
3v0k4: I'd suggest something like
```
[{..}, ...].to_json
```
That way you should be able to avoid escaping and quotes

app/controllers/api/user_sessions_controller.rb
AkkarinForest: Any tips on how this abstraction might look? 🙏

3v0k4: None. Let's think about that when the problem appears again

PR #223, opened 2019-05-24
Links generated in mobile app redirect to correct url

AkkarinForest:


app/controllers/measurement_sessions_controller.rb
```diff
@@ -20,25 +20,29 @@ class MeasurementSessionsController < ApplicationController
...
+      usernames: session.user.username,
       sensorId: stream.sensor_id
     }
 
-    redirect_to map_path(
-      :anchor => "/map_sessions?map=#{map.to_json}&selectedSessionIds=#{selected_session_ids.to_json}&data=#{data.to_json}"
-    )
+    anchor = "?data=#{data.to_json}&selectedSessionIds=#{selected_session_ids.to_json}&map=#{map.to_json}"
+
+    if (session.type == "FixedSession")
```
3v0k4: If route helpers are available in models we could exploit polymorphism instead of using a class check here (https://refactoring.guru/replace-conditional-with-polymorphism)

app/controllers/measurement_sessions_controller.rb
```diff
@@ -20,25 +20,29 @@ class MeasurementSessionsController < ApplicationController
...
+      redirect_to fixed_map_url(:anchor => anchor)
+    else
+      redirect_to map_url(:anchor => anchor)
```
3v0k4: We don't have a `mobile_map_url`?

AkkarinForest: We do. But `map_url ` seems consistent with `else` branch. If I were to use `mobile_map_url ` I should probably put it into `else if (session.type == "Session")` and then have something else for the default...

PR #224, opened 2019-05-27
Style CrowdMap controls

gitjul:
Requirements:

>- change the crowdmap switch to a toggle that looks and works like the "type" and "streaming" toggles on the fixed tab.  the left side reads "off" and the right side reads "on".  the crowdmap is "off" by default.
>
>- enlarge the dot on the slider.  not sure how much bigger but it needs to be a good size target for mouse clicking.
>
>- the text reading "CrowdMap resolution" should read "grid cell size"
>
> - put a "+" on the far right side of the slider line and a "-" on the far left side of the slider line.  moving the slider towards the "+" will enlarge the crowdmap grid cells.  moving the slider towards the "-" will shrink the crowdmap grid cells.

> let's do 1-40 and eliminate 41 as an option.
> 40 becomes the largest grid cell size, which maps to the current 11.
> 1 becomes the smallest grid cell size, which maps to the current 50. 

Looks like this now:

Screenshot 2019-05-28 at 20 13 16


app/javascript/elm/src/Main.elm
```diff
@@ -988,38 +987,34 @@ viewCrowdMapOptions isCrowdMapOn crowdMapResolution selectedSession tooltipIcon
...
 
 
 viewCrowdMapSlider : String -> Html Msg
 viewCrowdMapSlider resolution =
     div [ id "crowd-map-slider" ]
-        [ p []
-            [ text "Resolution" ]
-        , div []
-            [ input
+        [ label [] [ text <| "grid cell size: " ++ (String.fromInt <| 51 - (Maybe.withDefault 20 <| String.toInt resolution)) ]
```
AkkarinForest: I'm not sure what would happen, but why a different default on line 1010?

app/javascript/elm/src/Main.elm
```diff
@@ -988,38 +987,34 @@ viewCrowdMapOptions isCrowdMapOn crowdMapResolution selectedSession tooltipIcon
...
                 , value resolution
                 , max "50"
-                , min "10"
+                , min "11"
                 , type_ "range"
                 ]
                 []
-            , span []
-                [ text resolution ]
+            , span [ class "plus" ] [ text "+" ]
```
AkkarinForest: Would be nice if these were buttons? When I opened the page I instinctively tried to click them 😉

app/javascript/elm/src/Main.elm
```diff
@@ -988,38 +987,34 @@ viewCrowdMapOptions isCrowdMapOn crowdMapResolution selectedSession tooltipIcon
...
     div [ class "filters__toggle-group" ]
-        [ input
-            [ id "checkbox-crowd-map"
-            , type_ "checkbox"
-            , checked isCrowdMapOn
-            , Events.onClick ToggleCrowdMap
-            ]
-            []
-        , label [ for "checkbox-crowd-map" ] [ text "Crowd Map" ]
+        [ label [] [ text "CrowdMap:" ]
```
AkkarinForest: All the other labels are lower case. Is this an own name (nazwa własna)? If not I guess it should be consistent with the rest.

app/javascript/elm/src/Main.elm
gitjul: yeah, I know :) for now they're just indicators

app/javascript/elm/src/Main.elm
gitjul: this was explicitly requested by Michael

app/javascript/elm/src/Main.elm
gitjul: this will 99% never show but I changed the random number to 26 which maps to the default 25 resolution. thx!

app/javascript/elm/src/Main.elm
gitjul: thx for being a test user and confirming it :D I added a ticket for that now

PR #226, opened 2019-05-28
Tweak mobile sessions perf

3v0k4:
By avoiding access to the measurements in database the query should run faster. Unfortunately, it seems like the cache on production already does a lot and this change does not change much in perf.

### DIFF
--- a/app/models/session.rb
+++ b/app/models/session.rb
@@ -208,10 +208,6 @@ class Session < ActiveRecord::Base
       end
     end
 
-    if type == "MobileSession"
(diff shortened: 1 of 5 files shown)


app/models/session.rb
```diff
@@ -208,10 +208,6 @@ def as_json(opts=nil)
       end
     end
 
-    if type == "MobileSession"
```
3v0k4: This would prolly still used. I'll double check with Radek if it creates problems with the mobile app

PR #227, opened 2019-05-28
Filter popups

gitjul:
Before:

Screenshot 2019-05-29 at 12 39 26
Screenshot 2019-05-29 at 12 39 37

Now:

Screenshot 2019-05-29 at 14 34 42
Screenshot 2019-05-29 at 14 34 53

### DIFF
--- a/app/assets/stylesheets/modules/filters-popups.css.scss
+++ b/app/assets/stylesheets/modules/filters-popups.css.scss
@@ -26,6 +26,8 @@ $popup-background-color: #f7f7f7;
 }
 
 .filter-popup__list--more {
+  @include flexbox();
+  @include flex-wrap(wrap);
   animation: show-up 1s ease;
 }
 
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/modules/filters-popups.css.scss
```diff
@@ -45,19 +53,31 @@
...
+    font-weight: $font-stack-bold;
+  }
+
+  &:nth-child(2n) {
+    margin-left: 3%;
+  }
 }
 
-.filter-popup-button--more {
+.filter-popup-secondary-button {
```
gitjul: changed the name bc it's not actually a modifier class but a separate one = isn't added in addition to the base class but instead of it

3v0k4: Thanks for explaining!

app/assets/stylesheets/modules/overlay.css.scss
```diff
@@ -4,7 +4,11 @@
   top: 0;
   min-width: 100%;
   height: 100%;
-  z-index: 3;
+  z-index: 5;
   background-color: white;
   opacity: 0.6;
+
+  &.overlay--indoor {
+    z-index: 3; // session cards and info are on top of overlay
```
3v0k4: I see that we have some others `z-index`es. Would it make sense to extract them as vars?

3v0k4:
👁

app/assets/stylesheets/modules/overlay.css.scss
gitjul: good point, I'll keep that in mind

PR #228, opened 2019-05-29
Graph measurement dot on top

3v0k4:
By being in the same pane of starting markers but created later, it is on top.

Screen Shot 2019-05-29 at 13 52 57

One refactor we should consider is to have custom functions when building custom markers. Instead of
```js
        const customMarker = map.drawCustomMarker({
          object: { latLng },
          colorClass: "default",
          callback: callback(Session.id(session)),
          type: "marker"
        });

// or

        const marker = map.drawCustomMarker({
          object: { latLng },
          content: content,
          colorClass: heatLevel,
          callback: callback(Session.id(session)),
          type: "data-marker"
        });
```

Something like `buildStartingMarker` or `buildMeasurementMarker` that only accepts vars as input and internally already knows `type`, `colorClass` where it's not dynamic (eg for a data marker `type` is always `"data-marker"`

### DIFF
--- a/app/assets/stylesheets/modules/markers.css.scss
+++ b/app/assets/stylesheets/modules/markers.css.scss
@@ -9,6 +9,10 @@
   color: transparent;
   cursor: pointer;
 
+  &.measurement {
+    background-color: pink;
+  }
+
   &.default {
     background-color: $grey;
   }
--- a/app/javascript/angular/code/services/google/_map.js
+++ b/app/javascript/angular/code/services/google/_map.js
@@ -311,11 +311,15 @@ export const removeMarker = function(marker) {
   marker.setMap(null);
 };
 
-export const drawMarker = function({ position, title, zIndex, icon }) {
-  var newMarker = new google.maps.Marker({ position, title, zIndex, icon });
+export const drawCustomMarker = ({ position }) => {
+  const customMarker = buildCustomMarker({
+    object: { latLng: new google.maps.LatLng(position) },
+    colorClass: "measurement",
+    type: "marker"
+  });
 
-  newMarker.setMap(window.__map);
-  window.__markers.push(newMarker);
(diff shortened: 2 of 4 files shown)


app/javascript/angular/code/services/google/_map.js
```diff
@@ -311,11 +311,15 @@ export const removeMarker = function(marker) {
   marker.setMap(null);
 };
 
-export const drawMarker = function({ position, title, zIndex, icon }) {
-  var newMarker = new google.maps.Marker({ position, title, zIndex, icon });
+export const drawCustomMarker = ({ position }) => {
```
AkkarinForest: There already is a method `drawCustomMarker ` in this file, the one you edited on line 196. If that's on purpose then ignore my comment, just wanted to make sure that you noticed that

app/assets/stylesheets/modules/markers.css.scss
```diff
@@ -9,6 +9,10 @@
   color: transparent;
   cursor: pointer;
 
+  &.measurement {
+    background-color: pink;
```
AkkarinForest: Is it supposed to be pink? 🤔

gitjul: should be `$blue`

gitjul: also `measurement` is not intuitive for me, I'm thinking sth like `trace-marker`, sth that says that this indicates the exact measurement in the session, maybe `current-measurement`, I don't know. any ideas?

gitjul:
that was simpler than I thought o.O nice! 👏 
and thanks for taking care of this one <3

app/assets/stylesheets/modules/markers.css.scss
3v0k4: Discuss on call

PR #229, opened 2019-05-30
Sessions list perf

3v0k4:
This makes the query go much faster. Manual testing on clone showed an improvement of ~83% (from 6 to 1 seconds). The awesome thing is that whoever uses `Session.filter` is now getting that improvement for free.

I've just noticed today that `Session.from_location` was calling `map`. Thus, it was firing an additional query. By using ActiveRecord syntax the additional query is gone and the filtering on `in_rectangle` becomes part of the unique big query generated by `Session.filter`.

I've manually checked that the two query returns the same things.

### DIFF
--- a/app/models/session.rb
+++ b/app/models/session.rb
@@ -77,7 +77,7 @@ class Session < ActiveRecord::Base
     sessions = sessions.where(is_indoor: data[:is_indoor]) unless data[:is_indoor].nil?
 
     if data[:east] && data[:west] && data[:north] && data[:south]
-      sessions = sessions.from_location(data)
+      joins(:streams).merge(Stream.in_rectangle(data))
     end
 
     sensor_name = data[:sensor_name]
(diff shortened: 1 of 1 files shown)


app/models/session.rb
```diff
@@ -77,7 +77,7 @@ def self.filter(data={})
     sessions = sessions.where(is_indoor: data[:is_indoor]) unless data[:is_indoor].nil?
 
     if data[:east] && data[:west] && data[:north] && data[:south]
-      sessions = sessions.from_location(data)
+      joins(:streams).merge(Stream.in_rectangle(data))
```
AkkarinForest: should this be `sessions = sessions.joins(:streams).merge(Stream.in_rectangle(data))` ?

3v0k4: Yep, thanks :) Too excited to push

PR #230, opened 2019-05-30
Filter empty streams

AkkarinForest:
Using `measurements` to check if a stream is empty was too slow. I think using the counter is the best option. I created a rake task that will fix the counters that are out of sync. 
The only disadvantage to this approach is that I couldn't discover how they become incorrect so we can't predict if this will ever happen in the future.


spec/models/mobile_session_spec.rb
```diff
@@ -92,14 +92,14 @@
     before { MobileSession.destroy_all }
 
     it 'should exclude not contributed sessions' do
-      session1 = FactoryGirl.create(:mobile_session, :contribute => true)
-      session2 = FactoryGirl.create(:mobile_session, :contribute => false)
+      session1 = create_session_with_streams_and_measurements!()
```
3v0k4: No need for `()`

spec/test_utils.rb
```diff
@@ -0,0 +1,106 @@
+module TestUtils
+  def create_session_with_streams_and_measurements!(attributes = {})
+    session = create_session!(
+      id: attributes.fetch(:id, rand(1000000)),
+      contribute: attributes.fetch(:contribute, true),
+      start_time_local: attributes.fetch(:start_time_local,  DateTime.current),
```
3v0k4: Double space `  ` here

spec/test_utils.rb
```diff
@@ -0,0 +1,106 @@
...
+      end_time_local: attributes.fetch(:end_time_local,  DateTime.current),
+      user: attributes.fetch(:user, create_user!()),
```
3v0k4: No need for `()`

spec/test_utils.rb
```diff
@@ -0,0 +1,106 @@
+module TestUtils
+  def create_session_with_streams_and_measurements!(attributes = {})
```
3v0k4: Could you please check if you can avoid passing some attributes to the create_* methods? For example `id` in session I guess can be skipped

lib/tasks/sync_measurements_counter.rake
```diff
@@ -0,0 +1,6 @@
+desc "Sync measurements count on streams"
+task :sync_measurements_counter => [:environment] do
+  Stream.all.each do |stream|
```
3v0k4: Please use something like [`find_each`](https://apidock.com/rails/ActiveRecord/Batches/ClassMethods/find_each) or something similar that batches the loop. Otherwise, Ruby will load into memory all records at once

lib/tasks/sync_measurements_counter.rake
```diff
@@ -0,0 +1,6 @@
+desc "Sync measurements count on streams"
```
3v0k4: fix.rake is not a good name but I guess it could be a good place to keep all this "fix"ing tasks.

app/models/session.rb
```diff
@@ -94,6 +94,10 @@ def self.filter(data={})
       sessions = filter_by_time_range(sessions, data[:time_from], data[:time_to])
     end
 
+    if sessions
```
3v0k4: Why do we need the `if`?

AkkarinForest: Some tests were failing cause the query at the beginning of filter returned nil:
```
    sessions = order("sessions.created_at DESC")
    .where("contribute = true OR sessions.id in (?)", data[:session_ids])
    .joins(:user)
```
I susspect this can only happen it tests, but can't be sure so i added this `if`.

spec/test_utils.rb
AkkarinForest: Yes, it's used in many tests, and one of the passes session id.
But in general, there are some more tests that can be refactored now to use this file, but it requires some adjustments, so I didn't want to do all of them at once.

app/models/session.rb
3v0k4: I'd prefer if you took a bit of time and investigated not having the `if`. This method is already complex, let's try to avoid adding more on top

PR #231, opened 2019-05-30
Fix: no session list for dormant sessions

AkkarinForest:
For dormant sessions, the field was undefined and elm couldn't decode the date. 
I set it to null. We have `(Decode.field "average" <| Decode.nullable Decode.float)` on the elm side so this should be handled correctly now.


3v0k4:
So Elm doesn't consider `undefined` and `null` as a `Nothing` when decoding? Interesting!

PR #233, opened 2019-05-31
Tooltip chores

AkkarinForest:


app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
```diff
@@ -57,8 +57,6 @@ export const MobileSessionsMapCtrl = (
       location: "",
       tags: "",
       usernames: "",
-      gridResolution: 25,
-      crowdMap: false,
```
AkkarinForest: Changing these values doesn't change the behavior of the website, so I think they're obsolete. They are set in `elm.js`.

PR #235, opened 2019-05-31
Crowd map plus minus buttons

czorek:
Making the crowdmap resolution +/- buttons functional. There needs to be a transformation between 50-11 range in JS and 1-40 range in elm, so we are subtracting the resolution value from 51.

TODO:
- comment why 51 -
- boundary check (ie should not be possible to go below 1 and above 40)- DONE
- fix compile errors - introduce bound types, expose functions instead of a constructor - DONE
- extract function in update to update resolution in the model and call the port - DONE
- debounce buttons - DONE
- fix failing tests - DONE

### DIFF
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -240,9 +240,7 @@ type Msg
     | SetScrollPosition
     | NoOp
     | Timeout Int
-    | UpdateResolution
-    | MaybeIncreaseResolution
-    | MaybeDecreaseResolution
+    | MaybeUpdateResolution (BoundedInteger -> BoundedInteger)
 
 
 update : Msg -> Model -> ( Model, Cmd Msg )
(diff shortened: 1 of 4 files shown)


app/javascript/elm/src/Main.elm
```diff
@@ -527,6 +546,33 @@ update msg model =
...
+                ( model, Cmd.none )
+
+        MaybeIncreaseResolution ->
+            debounce BoundedInteger.addOne model
+
+        MaybeDecreaseResolution ->
+            debounce BoundedInteger.subOne model
+
+
+debounce : (BoundedInteger -> BoundedInteger) -> Model -> ( Model, Cmd Msg )
```
AkkarinForest: Maybe we don't have to pass the entire model

app/javascript/elm/src/Main.elm
```diff
@@ -226,6 +231,10 @@ type Msg
     | HighlightSessionMarker (Maybe Location)
     | GraphRangeSelected (List Float)
     | UpdateIsShowingTimeRangeFilter Bool
+    | Timeout Int
+    | UpdateResolution
+    | MaybeIncreaseResolution
+    | MaybeDecreaseResolution
```
AkkarinForest: Have you considered merging them into one? 
`| MaybeUpdateResolution (BoundedInteger -> BoundedInteger)`

``` 
MaybeUpdateResolution updateFunction ->
     debounce updateFunction model
```

then calling: `Events.onClick (MaybeUpdateResolution BoundedInteger.addOne)`

app/javascript/elm/src/Main.elm
```diff
@@ -527,6 +546,33 @@ update msg model =
             in
             ( { model | overlay = Overlay.update overlay model.overlay }, Cmd.none )
 
+        UpdateResolution ->
+            ( model, Ports.updateResolution (51 - BoundedInteger.getValue model.crowdMapResolution) )
+
+        Timeout int ->
+            if int == model.debouncingCounter then
+                update UpdateResolution { model | debouncingCounter = 0 }
```
AkkarinForest: Will it work if you just have here 
`( { model | debouncingCounter = 0 }, Ports.updateResolution (51 - BoundedInteger.getValue model.crowdMapResolution) )` ? 
And then you can remove `UpdateResolution` branch completely.

app/javascript/elm/tests/MainTests.elm
```diff
@@ -345,11 +346,11 @@ crowdMapArea =
                     |> Expect.equal (not onOffValue)
         , fuzz int "UpdateCrowdMapResolution changes the value of model.crowdMapResolution" <|
```
AkkarinForest: You could use `fuzz intRange 1 40` here. https://package.elm-lang.org/packages/elm-community/elm-test/latest/Fuzz

Then you can add a separate test to check the behaviour of UpdateCrowdMapResolution when the value is outside of bounds if you want :)

PR #237, opened 2019-05-31
Don't refetch sessions when toggling CrowdMap

AkkarinForest:
This solves the problem with preserving the session list length and also makes the switch faster. Anyone sees any edge cases when removing this fetch could cause an issue?


3v0k4:
Cool. Too hard for me to catch potential problems in that code ;)

PR #238, opened 2019-05-31
Fix sessions fetch when reloading page (or when copy / pasting link)

3v0k4:
Comments in the code explain why this fixes the problem. I've used `$broadcast` and `$on` that are Angular things. But I think that it's not adding more complexity in terms of removing Angular code later. In fact, if the logic for fetching sessions moves to Elm we would just call a port from the `addListenerOnce`.

Can review commit by commit. The first one is a refactor to use `addListener` in place of `listen`. The idea is to use the same vocabulary as GoogleMaps w/o introducing new words / verbs

### DIFF
--- a/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
+++ b/app/javascript/angular/code/controllers/_sessions_list_ctrl.js
@@ -13,7 +13,6 @@ export const SessionsListCtrl = (
   map
 ) => {
   let sessions;
-  let firstLoad = true;
   let highlightedSessionMarker = null;
   const elmApp = $window.__elmApp;
   const CANNOT_SELECT_MULTIPLE_SESSIONS = "You can't select multiple sessions";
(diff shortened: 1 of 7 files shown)


app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -100,6 +93,12 @@ export const SessionsListCtrl = (
     true
   );
 
+  $scope.$on("googleMapsReady", function() {
+    sessions.fetch({
```
3v0k4: This is called once when google maps is initialized

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -43,16 +42,10 @@ export const SessionsListCtrl = (
     "params.get('map')",
     ({ hasChangedProgrammatically }) => {
       console.log("watch - params.get('map')");
-      if (sessionsUtils.isSessionSelected() && !firstLoad) return;
+      if (sessionsUtils.isSessionSelected()) return;
       // when loading the page for the first time sometimes the watch is triggered twice, first time with hasChangedProgrammatically as undefined
       if (hasChangedProgrammatically === undefined) return;
 
-      if (firstLoad) {
-        sessions.fetch({ amount: params.paramsData["fetchedSessionsCount"] });
```
3v0k4: Inside `fetch` there's a call to `map.getBounds()` that returns the wrong object if GoogleMaps didn't have time to initialize. Up until now the debounce was adding 750 ms of delay that gave time to to GoogleMaps to initialize

app/javascript/angular/code/services/google/_map.js
```diff
@@ -25,6 +25,9 @@ export const map = (
       this.mapObj = googleMaps.init(element, options);
       this.markers = $window.__markers;
       this.listen("idle", this.saveViewport);
+      googleMaps.addListenerOnce(this.mapObj, "idle", () =>
+        $rootScope.$broadcast("googleMapsReady")
```
3v0k4: This is called on the first "idle". In practice when google maps is ready

3v0k4:
@AkkarinForest what a catch! What about with this new line `if (newValue === oldValue) return; // on angular $watch init` ?

AkkarinForest:
Cool, everything works :)

3v0k4:
Thanks for the multiple reviews on this @AkkarinForest! It helped a lot

PR #239, opened 2019-05-31
Fade map when time range picker is open

3v0k4:
^^^^^^^^

### DIFF
--- a/app/javascript/elm/src/Data/Overlay.elm
+++ b/app/javascript/elm/src/Data/Overlay.elm
@@ -1,0 +1,76 @@
+module Data.Overlay exposing (Model, Operation(..), Overlay(..), init, none, update, view)
+
+import Html exposing (Html, div, p, text)
+import Html.Attributes exposing (class, id)
+
+
+type Model
+    = Overlays (List Overlay)
+
+
+type Overlay
+    = HttpingOverlay
+    | IndoorOverlay
+    | PopupOverlay
+    | TimeFrameOverlay
+
+
+type Operation
+    = AddOverlay Overlay
+    | RemoveOverlay Overlay
+
+
+none : Model
+none =
+    Overlays []
+
+
+init : Bool -> Model
+init isIndoor =
+    if isIndoor then
+        Overlays [ IndoorOverlay ]
+
+    else
+        none
+
+
+update : Operation -> Model -> Model
+update operation model =
+    case ( operation, model ) of
+        ( AddOverlay overlay, Overlays [] ) ->
+            Overlays [ overlay ]
+
+        ( AddOverlay overlay, Overlays overlays ) ->
+            Overlays <| overlay :: overlays
+
+        ( RemoveOverlay overlay, Overlays [] ) ->
+            none
+
+        ( RemoveOverlay overlay, Overlays overlays ) ->
+            Overlays <| List.filter ((/=) overlay) overlays
+
+
+view : Model -> Html msg
+view (Overlays overlays) =
+    case List.head overlays of
+        Just overlay ->
+            case overlay of
+                HttpingOverlay ->
+                    div [ class "overlay", id "overlay--httping" ]
+                        [ div [ class "spinner" ] []
+                        ]
+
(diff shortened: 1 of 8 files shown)


app/javascript/elm/src/Main.elm
```diff
@@ -509,7 +530,32 @@ update msg model =
...
-            ( { model | isShowingTimeRangeFilter = isShown }, Cmd.none )
+            let
+                overlay =
+                    if isShown then
+                        Overlay
+
+                    else
+                        NoOverlay
+            in
+            ( { model | overlay = updateOverlay overlay model }, Cmd.none )
```
AkkarinForest: What do you think about having just one generic update case like so:
```
        ChangeOverlay overlayType onOrOff ->
            let
                overlay =
                    if onOrOff then
                        overlayType

                    else
                        NoOverlay
            in
            ( { model | overlay = overlay }, Cmd.none )

```

Then you can call it like so 

```
, Ports.updateIsHttping (ChangeOverlay OverlayWithSpinner)
, Ports.isShowingTimeRangeFilter (ChangeOverlay BasicOverlay)
```
or
```
        ClosePopup ->
            ( { model | popup = Popup.None, overlay = NoOverlay }, Cmd.none )

```

That would let us remove the duplication from update branches and make the logic reusable.

app/javascript/elm/src/Main.elm
```diff
@@ -349,7 +362,15 @@ update msg model =
...
-            ( { model | isHttping = isHttpingNow }, Cmd.none )
+            let
+                overlay =
+                    if isHttpingNow then
+                        HttpingOverlay
+
+                    else
+                        NoOverlay
+            in
+            ( { model | isHttping = isHttpingNow, overlay = updateOverlay overlay model }, Cmd.none )
```
AkkarinForest: I guess no need to keep `isHttping` in the model anymore?

3v0k4:
As discussed with @AkkarinForest offline we couldn't address his comments w/o a heavy refactoring. The latest state of the CR should have addressed them

app/javascript/elm/src/Data/Overlay.elm
```diff
@@ -0,0 +1,85 @@
+module Data.Overlay exposing (Model, Operation(..), Overlay(..), init, none, update, view)
+
+import Html exposing (Html, div, p, text)
+import Html.Attributes exposing (class, id)
+import List.Nonempty as Nonempty exposing (Nonempty(..))
+
+
+type Model
+    = Overlays (Nonempty Overlay)
+    | NoOverlay
```
AkkarinForest: So this is equivalent to `type Model = Overlays (List Overlay)` where `NoOverlay` covers the case of an empty list, right? Can you share what are the advantages of that?
Cause I can see that having a normal list would make the updated shorter - you would only need @ branches - for adding and filtering the list. 
And in the view, you can just have `case (head model) of`.

3v0k4: Good point. Removed nonempty but kept the opaque type. Thanks!

app/javascript/elm/src/Data/Overlay.elm
```diff
@@ -0,0 +1,76 @@
...
+        ( RemoveOverlay overlay, Overlays overlays ) ->
+            Overlays <| List.filter ((/=) overlay) overlays
+
+
+view : Model -> Html msg
+view (Overlays overlays) =
+    case List.head overlays of
+        Just overlay ->
+            case overlay of
+                HttpingOverlay ->
```
gitjul: I'd probably go with `Loading`

gitjul:
👍 👍 👏

app/javascript/elm/src/Data/Overlay.elm
3v0k4: I like your suggestion better but for now I'm leaving it the way it is cause the verb httping is used in multiple places and prefer consistency now. I'm not sure how I would rename the rest

3v0k4:
@gitjul thanks for the CR!

PR #241, opened 2019-05-31
Single session UI tweaks

gitjul:
Before:
before

After:
after

### DIFF
--- a/app/javascript/elm/src/Data/SelectedSession.elm
+++ b/app/javascript/elm/src/Data/SelectedSession.elm
@@ -147,6 +147,7 @@ view session heatMapThresholds linkIcon toMsg =
                         [ div [ class "single-session-avg-color", class <| Data.Session.classByValue (Just average) heatMapThresholds ] []
                         , span [] [ text "avg. " ]
                         , span [ class "single-session-avg" ] [ text <| String.fromInt <| round average ]
+                        , span [] [ text " µg/m³" ]
                         ]
                     , div [ class "session-numbers-container" ]
                         [ div [ class "single-min-max-container" ]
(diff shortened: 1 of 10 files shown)


app/javascript/elm/src/Data/SelectedSession.elm
```diff
@@ -147,6 +147,7 @@ view session heatMapThresholds linkIcon toMsg =
                         [ div [ class "single-session-avg-color", class <| Data.Session.classByValue (Just average) heatMapThresholds ] []
                         , span [] [ text "avg. " ]
                         , span [ class "single-session-avg" ] [ text <| String.fromInt <| round average ]
+                        , span [] [ text " µg/m³" ]
```
AkkarinForest: This can be different depending on the selected sensor. From Main.elm you could use: `Sensor.unitForSensorId model.selectedSensorId model.sensors` to get the current unit symbol. 
But since this is a selected session view, then maybe it's better to just add `unitSymbol` to `SelectedSession` type  . (I mean send it in a json, decode it and have it in the type for local usage.)

3v0k4: +1 to add to `SelectedSession`. @gitjul feel free to leave a ticket on the board if you prefer to focus on css

gitjul: @AkkarinForest @3v0k4 thx for reviewing and catching this 👍 yes, I'll leave it for you to focus on styles. should I leave this PR open for whomever takes it? or merge?

gitjul: https://llp.kanbanery.com/projects/8040/board/tasks/2413080

AkkarinForest: You can merge it and just create a ticket on top of todo.

gitjul: ok. it's on top of "Frontend" column but feel free to move

PR #243, opened 2019-06-04
Put filters into a div instead of a form.

AkkarinForest:
It solves a fascinating bug: submitting location was changing the state of "Redo search when map is
moved". This is because pressing enter on a form acts as pressing the
neares button of type submit (which is the default type). So pressing
enter while using filters was 'clicking' the refresh time range button,
which in turn was executing fetch sessions. And receiving new sessions
list changes "Redo Search in Map" back into "Redo search when map is
moved".

It's worth noting that the fetched sessions where incorrect. The command
to fetch session was triggered before the map had time to move to the
new location, so the sessions were for the old location.

### DIFF
--- a/app/assets/stylesheets/modules/filters.css.scss
+++ b/app/assets/stylesheets/modules/filters.css.scss
@@ -9,7 +9,7 @@
   width: $filters-width;
 }
 
-.filters__form {
+.filters__div {
   padding: 12px 12px $action-container-height 12px;
 }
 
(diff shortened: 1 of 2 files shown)


3v0k4:
Wow good one!

3v0k4:
The commit body is highly appreciated!

app/assets/stylesheets/modules/filters.css.scss
```diff
@@ -9,7 +9,7 @@
   width: $filters-width;
 }
 
-.filters__form {
+.filters__div {
```
gitjul: bad name, this should reflect the role of the element on the website and not the used html tag :) without looking at the neighboring structure I'd guess this should be `filters__container`. it can also stay `filters__form` if the perceived role is that of a form.

gitjul:
👏

PR #244, opened 2019-06-04
refactor: remove watch on selectedSessionIds

AkkarinForest:
Just some refactor I need to understand my next ticket :)

Also, what do you think of changing `selectedSessionIds  :: List Int` to `selectedSessionId :: Int` in the params at some point? That would change our API, but we already have other breaking changes, so maybe we should add this as well before the big release?


3v0k4:
Good job. As always it's difficult to anticipate what JS changes could cause. Maybe be sure to test for selecting / deselecting and reloading the page with selections.

I'm totally on board with the change you suggest. We will then need to handle `null`s so keep an eye on that.

PR #248, opened 2019-06-05
Update Rails to v5.0.x and Ruby to v2.6.3

3v0k4:
^^^^^


Gemfile
```diff
@@ -17,20 +17,18 @@ gem 'paperclip', '4.1.1'
...
 gem 'pry-rails'
 gem 'rack-cors', require: 'rack/cors'
-gem 'rails', '4.2.11.1'
+gem 'rails', '5.0.7.2'
 gem 'rake', '10.0.3'
 gem 'rb-gsl', '1.16.0.6'
 gem 'rubyzip', '>= 1.0.0'
 gem 'sidekiq'
 gem 'sidekiq-unique-jobs', '3.0.12'
-gem 'sinatra', require: false
```
3v0k4: It seems like it was installed when adding the Sidekiq web panel. But the Sidekiq docs don't require that

Gemfile
```diff
@@ -17,20 +17,18 @@ gem 'paperclip', '4.1.1'
...
 gem 'thin', '~> 1.6', '>= 1.6.4'
 gem 'test-unit', '~> 3.0'
 gem 'a9n'
 gem 'uglifier'
 gem 'sassc-rails'
 gem 'coffee-rails'
-gem 'activesupport-json_encoder'
```
3v0k4: Connected to the `BigDecimal` monkey patch in application.rb

app/views/devise/confirmations/new.html.erb
```diff
@@ -5,7 +5,7 @@
 
   <div class="field">
     <%= f.label :email %><br />
-    <%= f.email_field :email, autofocus: true %>
+    <%= f.email_field :email, autofocus: true, value: (resource.pending_reconfirmation? ? resource.unconfirmed_email : resource.email) %>
```
3v0k4: Regenerated with rails generate devise:install. Same for following files

app/models/user.rb
```diff
@@ -81,10 +81,10 @@ def sync(data)
       end
     end
 
-    # Apparently NOT IN doesn't work if uuids is empty
-    uuids = data.map { |x| x[:uuid] } + [""]
+
+    uuids = data.map { |x| x[:uuid] }
```
3v0k4: Checked with @czorek. This `uuid` should never be missing in any element of data

PR #249, opened 2019-06-06
Fixed sessions links

AkkarinForest:
I create a new endpoint cause the old one was using as_json. And it was impossible to change it and I need it to return `last_hour_average` now.

### DIFF
--- a/app/controllers/measurement_sessions_controller.rb
+++ b/app/controllers/measurement_sessions_controller.rb
@@ -26,10 +26,15 @@ class MeasurementSessionsController < ApplicationController
     sensorId = "Temperature-airbeam2-f (F)"
 
     if (session.type == "FixedSession")
-      data = { sensorId: sensorId, isIndoor: session.is_indoor, isStreaming: (session.last_measurement_at > Time.current - 1.hour) }
+      data =
+        { sensorId: sensorId,
+          isIndoor: session.is_indoor,
+          isStreaming: (session.last_measurement_at > Time.current - 1.hour),
+          usernames: session.user.username
+        }
       redirect_to fixed_map_path(:anchor => "?selectedSessionIds=#{selected_session_ids.to_json}&data=#{data.to_json}")
     else
-      data = { sensorId: sensorId }
+      data = { sensorId: sensorId, usernames: session.user.username, }
       redirect_to map_path(:anchor => "?selectedSessionIds=#{selected_session_ids.to_json}&data=#{data.to_json}")
     end
   end
--- a/app/controllers/measurement_sessions_controller.rb
+++ b/app/controllers/measurement_sessions_controller.rb
@@ -23,14 +23,14 @@ class MeasurementSessionsController < ApplicationController
     session = Session.find_by_url_token(params[:url_token]) or raise NotFound
 
     selected_session_ids = [session.id]
-    data = { sensorId: "Temperature-airbeam2-f (F)" }
-
-    anchor = "?selectedSessionIds=#{selected_session_ids.to_json}&data=#{data.to_json}"
+    sensorId = "Temperature-airbeam2-f (F)"
 
     if (session.type == "FixedSession")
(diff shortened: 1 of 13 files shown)


app/controllers/measurement_sessions_controller.rb
```diff
@@ -23,14 +23,19 @@ def show
...
-
-    anchor = "?selectedSessionIds=#{selected_session_ids.to_json}&data=#{data.to_json}"
+    sensorId = "Temperature-airbeam2-f (F)"
 
     if (session.type == "FixedSession")
-      redirect_to fixed_map_path(:anchor => anchor)
+      data =
+        { sensorId: sensorId,
+          isIndoor: session.is_indoor,
+          isStreaming: (session.last_measurement_at > Time.current - 1.hour),
```
3v0k4: Better to move this to `Session`. Something like `session.is_streaming`

app/controllers/measurement_sessions_controller.rb
```diff
@@ -23,14 +23,19 @@ def show
     session = Session.find_by_url_token(params[:url_token]) or raise NotFound
 
     selected_session_ids = [session.id]
-    data = { sensorId: "Temperature-airbeam2-f (F)" }
-
-    anchor = "?selectedSessionIds=#{selected_session_ids.to_json}&data=#{data.to_json}"
+    sensorId = "Temperature-airbeam2-f (F)"
```
3v0k4: Is this still a wip?

app/controllers/measurement_sessions_controller.rb
```diff
@@ -23,14 +23,19 @@ def show
...
 
     if (session.type == "FixedSession")
-      redirect_to fixed_map_path(:anchor => anchor)
+      data =
+        { sensorId: sensorId,
```
3v0k4: Remember to use snake case in Ruby: `sensor_id`

app/javascript/angular/code/services/_fixed_sessions.js
```diff
@@ -82,31 +78,39 @@ export const fixedSessions = (
...
+        this.drawMarkersWithoutLabel(sessionData);
+      }
+    },
     _selectSession: function(id, callback) {
       var sensorId = sensors.selectedId();
       var sensor = sensors.sensors[sensorId] || {};
       var sensorName = sensor.sensor_name;
       if (!sensorName) return;
       $http
-        .get("/api/realtime/sessions/" + id, {
```
3v0k4: Can we remove the old endpoint?

spec/controllers/api/fixed/sessions_controller_spec.rb
```diff
@@ -9,7 +9,7 @@
       end_time_local = DateTime.new(2001, 11, 4, 5, 6)
       sensor_name = "sensor-name"
       user = create_user!(username: username)
-      session = create_fixed_session!(user: user, title: title, start_time_local: start_time_local, end_time_local: end_time_local)
+      session = create_fixed_session!({ user: user, title: title, start_time_local: start_time_local, end_time_local: end_time_local })
```
3v0k4: In ruby you can leave the `{}` out. Internally `fnc(a: 1)` == `fnc({ a: 1 })`

app/controllers/measurement_sessions_controller.rb
AkkarinForest: These are parameters, they have to look like usual params we have in the urls.

app/javascript/angular/code/services/_fixed_sessions.js
AkkarinForest: Probably, I was planning to look into that in a separate PR.

app/controllers/measurement_sessions_controller.rb
3v0k4: The right side is a ruby var ;)

PR #250, opened 2019-06-06
Allow users of mobile app to select a stream

AkkarinForest:
Endpoint changes needed for the new mobile app version which will allow users to select a stream for the generated link.

### DIFF
--- a/app/controllers/measurement_sessions_controller.rb
+++ b/app/controllers/measurement_sessions_controller.rb
@@ -1,40 +1,42 @@
-# AirCasting - Share your Air!
-# Copyright (C) 2011-2012 HabitatMap, Inc.
-#
-# This program is free software: you can redistribute it and/or modify
-# it under the terms of the GNU Affero General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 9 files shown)


app/controllers/measurement_sessions_controller.rb
```diff
@@ -1,40 +1,42 @@
...
+
+    if result.success?
+      redirect_to result.value
+    else
+      render json: result.errors, status: :bad_request
+    end
+  end
+
+  def show_old
+    # suporrts legacy mobile apps relesed before 06.2019
```
3v0k4: Typo in supports ;)

app/services/api/to_link_anchor.rb
```diff
@@ -0,0 +1,43 @@
...
+    { sensorId: stream.sensor_id,
+      usernames: session.user.username,
+      heat: { highest: stream.threshold_very_high,
+              high: stream.threshold_high,
+              mid: stream.threshold_medium,
+              low: stream.threshold_low,
+              lowest: stream.threshold_very_low },
+    }
+
+    if (session.type == "FixedSession")
```
3v0k4: I remember suggesting this already so prolly it's a bad idea. We could have the path generation in both Mobile and Fixed session. That way this if is gone

app/services/api/to_link_anchor.rb
```diff
@@ -0,0 +1,43 @@
...
+      fix_data = data.merge(
+          isIndoor: session.is_indoor,
+          isStreaming: (session.last_measurement_at > Time.current - 1.hour)
```
3v0k4: Sorry for the double comment, we could have this as `session.is_streaming`

PR #251, opened 2019-06-07
Update to Rails 5.1.x

3v0k4:
^^^^^^


3v0k4:
Sorry, that wasn't in the upgrade guide and migrations on my machine didn't fail :) Fixed now

PR #253, opened 2019-06-07
Fetch sessions after moving the map to new location

AkkarinForest:
`$scope.sessions.fetch()` has to be passed in a callback to make sure it's executed after the map moves to the new location. 

I'm not sure I like that solution. It would be nice to have this in elm so we can have more control over the flow and just fetch session when needed, but I don't see that happening around google maps code 😕

### DIFF
--- a/app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
+++ b/app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
@@ -84,8 +84,13 @@ export const FixedSessionsMapCtrl = (
         $scope.sessions.fetch();
       });
 
-      elmApp.ports.findLocation.subscribe(location => {
-        FiltersUtils.findLocation(location, params, map);
+      elmApp.ports.goToLocation.subscribe(location => {
+        FiltersUtils.goToLocation({
+          location,
+          params,
+          map,
+          callback: () => $scope.sessions.fetch()
+        });
       });
 
       elmApp.ports.toggleIndoor.subscribe(isIndoor => {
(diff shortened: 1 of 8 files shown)


app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -85,7 +85,9 @@ export const FixedSessionsMapCtrl = (
       });
 
       elmApp.ports.findLocation.subscribe(location => {
-        FiltersUtils.findLocation(location, params, map);
+        FiltersUtils.findLocation(location, params, map, () => {
```
3v0k4: Maybe we could use an object as an argument? Example
```
findLoc({
  location:
  params
  ...
```
Gets difficult to track arguments when it's that many

app/javascript/angular/code/services/google/_map.js
```diff
@@ -68,14 +68,15 @@ export const map = (
       googleMaps.listenPanOrZoom(this.mapObj, callback);
     },
 
-    goToAddress: function(address) {
+    goToAddress: function(address, fetch_sessions) {
```
3v0k4: I'd suggest calling it `callback`. `fetchSession` is not a concern of google maps

3v0k4: Also, watch out for the snake_case in JS

app/javascript/javascript/filtersUtils.js
```diff
@@ -148,9 +148,9 @@ const updateTooltipContent = (link, tooltip) => {
   });
 };
 
-export const findLocation = (location, params, map) => {
+export const findLocation = (location, params, map, fetch_sessions) => {
```
3v0k4: Maybe we could rename the function? It's not really finding a location

PR #254, opened 2019-06-07
Aggregate columns not included in the GROUP BY clause

AkkarinForest:
We assume that for a given sensor_name, measurement_type and unit_symbol
all thresholds are the same, so it is ok to select a random value.

### DIFF
--- a/app/models/stream.rb
+++ b/app/models/stream.rb
@@ -148,8 +148,8 @@ class Stream < ApplicationRecord
   def self.sensors
     joins(:session).
       where("sessions.contribute" => true).
-      select("sensor_name, measurement_type, threshold_very_low, threshold_low, unit_symbol,
-           threshold_medium, threshold_high, threshold_very_high, count(*) as session_count").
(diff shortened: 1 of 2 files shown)


app/models/stream.rb
```diff
@@ -148,8 +148,8 @@ def build_measurements!(data = [], jid = nil)
   def self.sensors
     joins(:session).
       where("sessions.contribute" => true).
-      select("sensor_name, measurement_type, threshold_very_low, threshold_low, unit_symbol,
-           threshold_medium, threshold_high, threshold_very_high, count(*) as session_count").
+      select("sensor_name, measurement_type, MIN(threshold_very_low) as threshold_very_low, MIN(threshold_low) as threshold_low, unit_symbol,
+           MIN(threshold_medium) as threshold_medium, MAX(threshold_high) as threshold_high, MAX(threshold_very_high) as threshold_very_high, count(*) as session_count").
```
3v0k4: Cool! Is there a more Ruby way of writing this select? (i.e less strings more code)

PR #255, opened 2019-06-07
Remove old endpoint

AkkarinForest:
Since single session view doesn't need to download the session list in the background anymore, we can remove this endpoint.

### DIFF
--- a/app/models/session.rb
+++ b/app/models/session.rb
@@ -61,7 +61,7 @@ class Session < ApplicationRecord
 
   def self.filter_(data={})
     sessions = order("sessions.created_at DESC")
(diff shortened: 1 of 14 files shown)


app/models/session.rb
```diff
@@ -61,7 +61,7 @@ class Session < ApplicationRecord
 
   def self.filter_(data={})
     sessions = order("sessions.created_at DESC")
-    .where("contribute = true OR sessions.id in (?)", data[:session_ids])
+    .where("contribute = true")
```
3v0k4: You could (`contribute: true`)

PR #258, opened 2019-06-10
Order sensor list naturally

gitjul:
Apply NaturalOrdering to the list of sensors same way as was done for the parameter list.

### DIFF
--- a/app/javascript/elm/src/Sensor.elm
+++ b/app/javascript/elm/src/Sensor.elm
@@ -108,8 +108,7 @@ parameters sensors =
     let
         othersParameters =
             sensors
-                |> List.map .parameter
-                |> List.map String.trim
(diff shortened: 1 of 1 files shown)


app/javascript/elm/src/Sensor.elm
```diff
@@ -108,8 +108,7 @@ parameters sensors =
     let
         othersParameters =
             sensors
-                |> List.map .parameter
-                |> List.map String.trim
+                |> List.map (String.trim << .parameter)
```
3v0k4: Nice composition!

3v0k4:
Good one!

app/javascript/elm/src/Sensor.elm
gitjul: thx to @AkkarinForest :)

PR #259, opened 2019-06-11
Pretty ruby

3v0k4:
Add pretty for ruby. Please double check your editor automatically formats ruby files as it does for js and css. I'm around if I can help with that!

It would be great if we agreed on using the defaults. But I'm ok if you really really feel that we should change some config


AkkarinForest:
This is awesome! Thanks for taking care of that.

As for the defaults - it's a good idea to keep them. The only thing I would consider is increasing the number of characters per line (same for js). I feel like 80 is from ancient times when monitors were much smaller.

3v0k4:
@AkkarinForest On one hand I agree on the other I'd say that
- shorter lines less info to digest at once so maybe simpler code?
- I use to work with a lot of vertical splits and that makes my life a bit easier

PR #260, opened 2019-06-11
Fetch measurements for all streams for a given sensor name

AkkarinForest:
Sometimes a session has more than one stream with a given `sensor_name`. When `to_session_hash.rb` was just taking `session.streams.first` there was data missing from the graph for these sessions.

### DIFF
--- a/app/models/api/measurements.rb
+++ b/app/models/api/measurements.rb
@@ -8,7 +8,7 @@ module Api::Measurements
 
   Schema =
     Dry::Validation.Schema do
-      required(:stream_id).filled(:str?)
+      required(:stream_ids).filled(:str?)
       optional(:start_time).filled(:str?)
       optional(:end_time).filled(:str?)
     end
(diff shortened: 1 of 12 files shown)


app/models/api/measurements.rb
```diff
@@ -7,15 +7,15 @@ module Types
   end
 
   Schema = Dry::Validation.Schema do
-    required(:stream_id).filled(:str?)
+    required(:stream_ids)
```
3v0k4: Have you thought about adding the `filled` part?

app/models/api/measurements.rb
```diff
@@ -7,15 +7,15 @@ module Types
...
     optional(:start_time).filled(:str?)
     optional(:end_time).filled(:str?)
   end
 
   class Struct < Dry::Struct
     transform_keys(&:to_sym)
 
-    attribute :stream_id, Types::Coercible::Integer
+    attribute :stream_ids, Types::Array
```
3v0k4: Have you considered checking the content of the array?

app/services/api/to_measurements_array.rb
```diff
@@ -24,11 +24,11 @@ def page
     start_time = Time.at(form.to_h[:start_time] / 1000)
     end_time = Time.at(form.to_h[:end_time] / 1000)
 
-    Measurement.with_streams(form.to_h[:stream_id]).where(time: start_time..end_time)
+    Measurement.with_streams(form.to_h[:stream_ids].split(",")).where(time: start_time..end_time)
```
3v0k4: `with_streams` works with either array or single id?

AkkarinForest: Yes, active record `where(stream_id: stream_ids)` works like that.

PR #261, opened 2019-06-11
Display correct unit of measure for selected session

3v0k4:
Show unit of measure depending of what sensor is selected

Screen Shot 2019-06-11 at 14 19 04

### DIFF
--- a/app/javascript/elm/src/Data/SelectedSession.elm
+++ b/app/javascript/elm/src/Data/SelectedSession.elm
@@ -117,8 +117,8 @@ updateRange result measurements =
             result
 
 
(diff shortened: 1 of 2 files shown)


gitjul:
👍 

if I understood correctly, Alex's suggestion was to send the sensor unit in the json from the backend instead of doing `Sensor.unitForSensorId model.selectedSensorId model.sensors`, have you considered that?

3v0k4:
@gitjul I considered but haven't done at first. I've implemented what @AkkarinForest suggested here https://github.com/HabitatMap/AirCasting/pull/262. Which one do you prefer?

3v0k4:
Decided to merge https://github.com/HabitatMap/AirCasting/pull/262/files

PR #263, opened 2019-06-11
Turn off autocomplete for location filter

AkkarinForest:
I reverted my previous changes and removed the autocomplete.


3v0k4:
:eye"

PR #265, opened 2019-06-12
Check last measurement time individually for each stream

AkkarinForest:
Before this change, if the AirBeams were recording measurements with a different time zone some of them were not included in the last hour average and number of samples.
This created a bug where clusters marker had a different color than the average in the info window would suggest.

### DIFF
--- a/app/services/fixed_region_info.rb
+++ b/app/services/fixed_region_info.rb
@@ -1,11 +1,12 @@
 class FixedRegionInfo
   def call(data)
     streams_ids = streams_ids(data)
+    stats = calculate_stats(streams_ids)
 
-     {
-      average: calculate_average(streams_ids).to_f,
+    {
+      average: stats[:average],
       number_of_contributors: number_of_contributors(streams_ids),
-      number_of_samples: number_of_samples(streams_ids),
+      number_of_samples: stats[:count],
       number_of_instruments: streams_ids.count,
     }
   end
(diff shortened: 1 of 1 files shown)


app/services/fixed_region_info.rb
```diff
@@ -1,53 +1,53 @@
...
+  def number_of_samples(streams_ids)
+     streams_ids.reduce(0) do |acc, stream_id|
+      end_time = end_time(stream_id)
+
+      acc + Measurement.with_streams(stream_id).where(time: (end_time - 1.hour)..end_time).count
+    end
+  end
+
+  def end_time(stream_id)
+    Measurement.with_streams(stream_id).select("MAX(time) as last_time").first.last_time
```
3v0k4: Would this work too? https://apidock.com/rails/ActiveRecord/Calculations/maximum

app/services/fixed_region_info.rb
```diff
@@ -1,53 +1,53 @@
...
       count
   end
 
-   def number_of_samples(streams_ids, end_time)
-    Measurement.
-      with_streams(streams_ids).
-      where(time: (end_time - 1.hour)..end_time).
-      count
+  def number_of_samples(streams_ids)
+     streams_ids.reduce(0) do |acc, stream_id|
```
3v0k4: Can we implement this in a clean way with one query?

app/services/fixed_region_info.rb
```diff
@@ -1,53 +1,53 @@
 class FixedRegionInfo
   def call(data)
     streams_ids = streams_ids(data)
-    last_measurement_time = last_measurement_time(streams_ids)
 
      {
-      average: calculate_average(streams_ids, last_measurement_time).to_f,
+      average: calculate_average(streams_ids).to_f,
```
3v0k4: What do you think of dropping the `to_f` and use `fdiv` in `calculate_average`?

https://apidock.com/ruby/Numeric/fdiv

app/services/fixed_region_info.rb
3v0k4: Cause otherwise this fires one query per stream. How's the performance of this feature?

AkkarinForest: I wouldn't know how to do that.
Instead, I merged number_of_samples with calculate_average - so this limits the number of queries by half :)

AkkarinForest:
This file is covered by tests. 

Mobile is different - we pass the time by params, so I guess no change is needed there. The bug, in this case, was that we were fetching the time once from the database assuming that it would be the same for all streams.

PR #267, opened 2019-06-12
Style cluster and crowdmap info windows

gitjul:


3v0k4:
👁 We should start doing sessions together so you can teach me some CSS

gitjul:
@3v0k4 be careful what you wish for

PR #268, opened 2019-06-13
Crowd map fix group by query

3v0k4:
Measurement has `default_scope { order('time ASC') }` which influences
every query on the model. We should move away from default scopes cause
they create this sort of problems. In other words, queries should be
explicit, the order should be used where necessary not in a
`default_scope`. This is just patching the problem for now

From what I can see this is the only query having this problem.

### DIFF
--- a/lib/average_info.rb
+++ b/lib/average_info.rb
@@ -30,7 +30,7 @@ class AverageInfo
 
   def measurements
     @measurements ||=
-      Measurement.select(
+      Measurement.unscoped.select(
         'AVG(value) AS avg, ' +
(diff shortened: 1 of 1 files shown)


AkkarinForest:
`default_scope` was causing this, wow, nice catch!

+1 for removing the scope. I think it also slows the performance, so we should definitely remove it from queries where it's not needed. Just tried running `Measurement.unscoped.last` and `Measurement.last` - the former took 121ms and the latter still haven't finish running as I'm writing this message...

3v0k4:
Good point with performance!

PR #269, opened 2019-06-13
Notes pagination

AkkarinForest:
It looks like I created a new file, but it's parts are copied from `app/javascript/angular/code/services/google/note.js` so there might be still some places to refactor there.


app/javascript/angular/code/services/_draw_session.js
```diff
@@ -41,10 +42,7 @@ export const drawSession = (sensors, map, heat, note, empty) => {
         drawnObjects.markers.push(marker);
         points.push(measurement);
       });
-
-      (session.notes || []).forEach(function(noteItem, idx) {
-        drawnObjects.noteDrawings.push(note.drawNote(noteItem, idx));
-      });
+      drawnObjects.notes = drawNotes(session.notes || [], map);
```
AkkarinForest: I tried passing the callback instead of the whole `map`, but map uses this.something in too many places. Maybe in the next refactor :)

app/javascript/angular/code/services/_draw_session.js
```diff
@@ -38,14 +39,12 @@ export const drawSession = (sensors, map, heat, note, empty) => {
...
-        drawnObjects.markers.push(marker);
+        drawnObjects.push(marker);
         points.push(measurement);
       });
 
-      (session.notes || []).forEach(function(noteItem, idx) {
-        drawnObjects.noteDrawings.push(note.drawNote(noteItem, idx));
-      });
-      drawnObjects.lines.push(map.drawLine(points));
+      drawnObjects = drawnObjects.concat(drawNotes(session.notes || [], map));
```
AkkarinForest: I tried passing the callback instead of the whole map, but map uses this.something in too many places. Maybe in the next refactor :)

PR #270, opened 2019-06-13
UI: Single session tweaks

gitjul:
- remove excessive whitespace below zoom buttons
- change hover state of zoom button to blue - before when hovering on
  active element it became white which was counterintuitive because
clicking on it doesn't turn it off
- slimmer scrollbar
- pointer cursor on scroll thumb and arrows
- clip session names to two lines
- restructure html to improve the flex layout, align columns
- rename classes using bem-style
- align the button of thescrollbar with the bottom ofaction buttons
- fix credits position after changes
- move close button away from credits so that users don't mistakenly click the credits when they want to close the session and vice versa
- fix close button broken in Safari

One requested change wasn't possible to fix: making the whole area of zoom buttons clickable.
I submitted an issue with Highcharts: https://github.com/highcharts/highcharts/issues/10986

Before - one line name:

Screenshot 2019-06-13 at 15 09 09

Before - three line name:

Screenshot 2019-06-13 at 16 12 45

After - one line name:

Screenshot 2019-06-13 at 16 13 46

After - three line name (Chrome, webkit):

Screenshot 2019-06-13 at 16 10 28

After - threee line name (FF 67 - in 68 line-clip will work):

Screenshot 2019-06-13 at 16 10 38

### DIFF
--- a/app/assets/stylesheets/modules/action-buttons.css.scss
+++ b/app/assets/stylesheets/modules/action-buttons.css.scss
@@ -2,7 +2,6 @@
   @include flexbox();
   @include justify-content(space-between);
   @include align-items(center);
-  height: $action-container-height;
   width: 100%;
 }
 
--- a/app/assets/stylesheets/modules/buttons.css.scss
+++ b/app/assets/stylesheets/modules/buttons.css.scss
@@ -40,11 +40,3 @@
   background: transparent;
   border: none;
 }
-
-.close-button--session {
-  font-size: 32px;
-  line-height: 0;
-  position: absolute;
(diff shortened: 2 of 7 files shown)


gitjul:
todo separately: shrink zoom buttons

3v0k4:
👁

3v0k4:
I think we are using in the code both "single session" and "selected session" for the same case. What would you use if we needed to choose?

gitjul:
yeah, I've been getting confused by this. "single session" is what I'd choose.

PR #271, opened 2019-06-13
Average dot

AkkarinForest:
Added a little dot on the left of `avg`:

Screen Shot 2019-06-13 at 15 00 38

### DIFF
--- a/app/assets/stylesheets/modules/info-window.css.scss
+++ b/app/assets/stylesheets/modules/info-window.css.scss
@@ -48,3 +48,14 @@
   font-weight: $font-stack-bold;
   margin-top: 10px;
 }
+
+.info_window__avg-color {
+  min-width: 9px;
+  min-height: 9px;
+  max-width: 9px;
+  max-height: 9px;
+  border-radius: 12px;
+  margin-right: 6px;
+  margin-top: 5px;
+  float: left;
+}
--- a/app/javascript/angular/code/controllers/_info_window_ctrl.js
+++ b/app/javascript/angular/code/controllers/_info_window_ctrl.js
@@ -1,8 +1,11 @@
-export const InfoWindowCtrl = ($scope, sensors, infoWindow, map) => {
+export const InfoWindowCtrl = ($scope, sensors, infoWindow, map, heat) => {
   $scope.sensors = sensors;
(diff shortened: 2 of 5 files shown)


app/javascript/angular/code/controllers/_info_window_ctrl.js
```diff
@@ -1,8 +1,31 @@
...
   $scope.zoomIn = () => {
     map.zoomToSelectedCluster();
   };
+
+  $scope.class = () =>
+    "info_window__avg-color " +
+    classByValue(heat.getLevel(Math.round(infoWindow.data.average)));
+};
+
+const classByValue = value => {
```
3v0k4: What do you think of keeping this in the `heat` module?

3v0k4: There could be something like `heat.classByValue(infoWindow.data.average)`

AkkarinForest: Good point! thanks

app/assets/stylesheets/modules/info-window.css.scss
```diff
@@ -48,3 +48,14 @@
   font-weight: $font-stack-bold;
   margin-top: 10px;
 }
+
+.info_window__avg-color {
+  min-width: 9px;
+  min-height: 9px;
+  max-width: 9px;
+  max-height: 9px;
```
gitjul: why min and max instead of just width & height?

AkkarinForest: That's how it's done for a similar dot in `single-session.css.scss` in `.single-session__avg-color `. Do you think it's not needed here?

gitjul: ok, don't worry, I'll take a look. looks weird! maybe some old code

gitjul:
@AkkarinForest just noticed that it's not added in the crowd map info window

gitjul:
@AkkarinForest added it :)

app/assets/stylesheets/modules/info-window.css.scss
gitjul: ✅

PR #272, opened 2019-06-14
One method to clear the map

AkkarinForest:
I created one method in `mapsUtils.js` that will be responsible for clearing the map. It creates an array on the map object which stores the drawn markers. So every time `setMap` is used that marker is automatically added to the array. There is no need for the developer to remember about adding it.
The same with info windows.

I didn't figure out how to achieve the same effect with rectangles and session lines. So they still have to be added to the array every time they are created. 
Also, this clearing method doesn't make sense for trace markers (the blue that appears when hovering over the graph) - cause then we don't want to clear the entire map, only the blue dot. 
So these are problems for the next refactor, but at least this PR is the first step in centralizing the clearing map logic.


app/javascript/javascript/mapsUtils.js
```diff
@@ -4,3 +4,59 @@ export const lengthToPixels = (length, zoom) => {
 };
 
 export const pixelsToLength = (pixels, zoom) => pixels * Math.pow(2, -zoom);
+
+export const clearMap = () => {
+  if (process.env.NODE_ENV !== "test") {
```
3v0k4: What do you think of
```
if (process.env.NODE_ENV === "test") return;
...
```
that would make the code a bit more flat

app/javascript/javascript/mapsUtils.js
```diff
@@ -4,3 +4,59 @@ export const lengthToPixels = (length, zoom) => {
...
+
+  google.maps.Marker.prototype._setMap = google.maps.Marker.prototype.setMap;
+
+  google.maps.Marker.prototype.setMap = function(map) {
+    if (map) {
+      map.markers.push(this);
+    }
+    this._setMap(map);
+  };
+}
```
3v0k4: I'm not a big fan of modifying prototypes. I think it's great if this is a step towards centralising the logic. In the end I think we should wrap Google Maps instead of modifying it.

If we had all of the calls interacting with google maps in one place then we could keep track of things in there encapsulated.

AkkarinForest: I agree modifying is not ideal (although I found this approach as a recommended solution on StackOverflow.) On the other hand, using a wrap still, leave a possibility that someone unfamiliar with a project will use `marker.setMap(mapObj)` directly, and then have a bug with this marker not being cleared from the map.

AkkarinForest: And in this case, these bugs are hard to notice (like when one marker is on top of it's duplicate or if the bug only happens when switching filters in a specific order).
What I mean is that a solution where it's impossible to introduce a bug is better than a solution that relies on developers following good practices 😉 
Anyway, next time there is an opportunity to refactor that code we can do some brainstorming :)

PR #274, opened 2019-06-17
Preserve the scroll position

AkkarinForest:
When deselecting a session, the sessions list preserves the scroll position. 
This also works for links saved with a selected session. However, I had to use ports and rely on the mutation observer to make sure that the `setScrollPosition` is executed after the sessions list already loaded. I'm not sure if there's a simpler solution.

### DIFF
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -2,6 +2,7 @@ module Main exposing (Msg(..), defaultModel, update, view)
 
 import Api
 import Browser exposing (..)
+import Browser.Dom as Dom
 import Browser.Events
 import Browser.Navigation
 import Data.GraphData exposing (GraphData)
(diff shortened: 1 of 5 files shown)


app/javascript/elm/src/Main.elm
```diff
@@ -226,6 +232,9 @@ type Msg
     | HighlightSessionMarker (Maybe Location)
     | GraphRangeSelected (List Float)
     | UpdateIsShowingTimeRangeFilter Bool
+    | SaveScrollPosition Float
+    | SetScrollPosition
```
czorek: So one of those two branches above sets and saves the scroll position on the model, and the other on the sessions list viewport? Have you considered making `SetScrollPosition` name a bit more specific and explicit about that? Just my 2 cents, it looks good either way!

AkkarinForest: Thanks, that's a good point. I'll try to find better names :)

PR #288, opened 2019-06-24
Only show scroll on hover

AkkarinForest:
Otherwise, google maps attribution is permanently obscured.


app/assets/stylesheets/modules/sessions.css.scss
```diff
@@ -2,7 +2,12 @@
   @include flexbox();
   @include flex-direction(row);
   width: calc(100vw - #{$filters-width});
+  overflow-x: hidden;
+  padding-bottom: 15px;
```
gitjul: I'm fine with the hardcoding but I'd add a comment. better yet, I'd add a variable at the top which says sth about this value being the height of the attribution and using it here.

PR #289, opened 2019-06-24
Don't cover heatmap with indoor sessions overlay

gitjul:
Cleaned up z-indexes a bit gathering them in variables and hopefully naming them in a clear way.


gitjul:
Note: this unblocks heatmap controls for both single session and the list of sessions. In the second case manipulating the heatmap doesn't affect the session list.

PR #290, opened 2019-06-25
Set permanent sessions list height

AkkarinForest:


app/javascript/elm/src/Main.elm
```diff
@@ -932,7 +932,7 @@ viewSessions fetchableSessionsCount sessions heatMapThresholds =
         text ""
 
     else
-        div []
+        div [ class "sessions-list" ]
```
gitjul: this needs to be BEM-ified at some point, it's messy (I mean the whole adjacent code, not this class, just a note)

PR #292, opened 2019-06-25
UI: responsive navigation

gitjul:
Screenshot 2019-06-25 at 18 22 16

Screenshot 2019-06-25 at 18 22 23

Screenshot 2019-06-25 at 18 23 24


app/assets/javascripts/application.js
```diff
@@ -17,3 +17,10 @@
 
   You can contact the authors by email at <info@habitatmap.org>
 */
+
+// mobile site nav
+window.addEventListener("load", () => {
+  document.querySelector(".js-menu-button").addEventListener("click", e => {
+    document.querySelector("header").classList.toggle("menu-collapsed");
+  });
+});
```
gitjul: since there's nothing here yet thought I might as well just put this script here, but I'm open to suggestions where else to have this.

app/assets/stylesheets/modules/site-nav.css.scss
```diff
@@ -1,41 +1,100 @@
-.nav {
-  margin: 0;
-  min-width: 100%;
-  max-width: 100%;
-  height: $nav-height;
-  background-color: white;
+$breakpoint: 780px;
```
gitjul: I prefer to keep variables that are used only in one file at the top of that file, but that's a matter of preference. if you have strong feelings about this lmk

app/javascript/elm/src/Main.elm
```diff
@@ -714,41 +724,59 @@ viewDocument model =
...
-            , li []
-                [ a [ class "nav__link", href "/donate" ]
-                    [ text "Donate" ]
+        , nav
+            [ class "nav"
+            , id "menu"
+            , role "navigation"
+            ]
+            [ ul []
+                [ li []
```
gitjul: not perfect BEM structure, but didn't feel like refactoring at this point

app/javascript/elm/src/Main.elm
```diff
@@ -67,6 +68,7 @@ type alias Model =
     , overlay : Overlay.Model
     , scrollPosition : Float
     , debouncingCounter : Int
+    , isNavExpanded : Bool
```
3v0k4: I'd encourage as an experiment to create custom types also when a boolean would be enough. In this case
```
type Nav
  = Expanded
  | Collapsed
```

It's more code but it makes it more readable. Especially cause then stuff like
```
not model.isNavExpanded
```
becomes
```
model.nav == Collapsed
```
Also, this makes impossible to use `Nav` as a boolean for example when calling a function

gitjul: it doesn't seem to work like that :(

3v0k4: I'm available next week if you want to take a look at this together ;)

AkkarinForest: Can I join you? Cause I had a similar dilemma recently. I wanted to use the custom type but realized it will make an update branch longer. Instead of
```
ToggleSth ->
    (  { model | sth = not model.sth }, Cmd.none)
```
we would need:
```
ToggleSth ->
    let
        newSth = 
            case model.sth of
                 Value1 ->
                      Value2
                 Value2 ->
                      Value1

    (  { model | sth = newSth }, Cmd.none )
```
Or is there a shorter way? I thought about creating a custom equivalent of `not` function for my type?

3v0k4: Yeah. You could use something like https://package.elm-lang.org/packages/joneshf/elm-tagged/latest/ and then just using `Tagged.map` would be enough

gitjul: is that what you prefer?

3v0k4: It looks cool. Plus the implementation is prolly a few lines of code. Powerful pattern but really lightweight. If anyone wants to dig more, that library is using "phantom types" to "tag" different values of the same type

3v0k4: I'd be +1 if anybody added that

PR #293, opened 2019-06-25
Sitemap.xml file

AkkarinForest:
I created and added a `sitemap.xml`. 
I updated `robots.txt` with the content prepared by Sonia.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -38,7 +38,8 @@ set :linked_files,
       'config/database.yml',
       'config/secrets.yml',
       'newrelic.yml',
-      'config/configuration.yml'
+      'config/configuration.yml',
+      'public/robots.txt'
     )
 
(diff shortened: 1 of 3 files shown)


gitjul:
also not an expert :) I guess this will affect the search response and we'll see if the sitemap is nicely generated (i.e. specific links under the main one in google search results)

AkkarinForest:
Yes, it's this :)
But it didn't create sitelinks in google search results. The sitemap can help with that, but it doesn't guarantee anything. In the end, Google has the power to decide who is worthy of having them :D [More here](https://searchenginewatch.com/sew/how-to/2430604/how-do-i-get-sitelinks-to-appear-in-my-site-s-search-results)
Also, it might take some time before google notices that we added a sitemap, cause the probably don't crawl us every day 🤔

PR #276, opened 2019-06-19
Vanilla js lightbox library

AkkarinForest:
The library: https://github.com/imgix/luminous

PR #294, opened 2019-06-26
Vanilla lightbox

gitjul:
Replace the jquery lighbox plugin with a simpler vanilla lightbox which doesn't look broken.


gitjul:
going with https://github.com/HabitatMap/AirCasting/pull/276 after all, this was more complicated to use

PR #296, opened 2019-06-27
Use custom markers to draw selected session measurements

AkkarinForest:
Using HTML instead of SVG icons will make it easier to switch colours themes. 
Also, there was no reason to use different logic for drawing similar markers.

But I can't do the same for the clusters for fixed sessions. Cause we are using google maps clusterer and it only accepts url as icons. How do you feel about having HTML files instead of svg files and passing a url to them???


gitjul:
@AkkarinForest let's talk offline, I don't quite get the HTML file thing

3v0k4:
👁 Not sure I get what's going on here but the code looks simpler!

PR #297, opened 2019-06-27
Refactor: use Path type for AC logo

gitjul:


app/javascript/elm/src/Main.elm
```diff
@@ -724,20 +724,20 @@ viewDocument model =
...
     div [ id "elm-app" ]
-        [ viewNav model.logoNav model.menuIcon model.isNavExpanded
+        [ viewNav (Path.toString model.navLogo) model.menuIcon model.isNavExpanded
         , viewMain model
         ]
 
 
 viewNav : String -> Path -> Bool -> Html Msg
-viewNav logoNav menuIcon isNavExpanded =
+viewNav navLogo menuIcon isNavExpanded =
```
3v0k4: Have you considered passing the `Path` to the function? That way the unboxing happens right where needed

AkkarinForest:
👀

app/javascript/elm/src/Main.elm
gitjul: definitely!

PR #298, opened 2019-06-28
Optimize crowdmap

czorek:
https://llp.kanbanery.com/projects/8040/board/tasks/2412128

As it turns out, we can't get rid of two separate averages endpoints, because the web and mobile average data is generated differently - the web app needs to select measurements in viewport, belonging to specific sessions and filtered by time range, apart from the date range. Mobile data is only concerned with the viewport and date range.


lib/crowdmap_averages.rb
```diff
@@ -0,0 +1,105 @@
+require_dependency 'aircasting/username_param'
```
AkkarinForest: Do you know you can use `git mv oldfilename newfilename` to rename files. This way git can recognize the connection even when the renamed file was edited. Makes looking at PR and searching git history easier :)
https://help.github.com/en/articles/renaming-a-file-using-the-command-line

lib/crowdmap_averages.rb
```diff
@@ -0,0 +1,105 @@
...
+  class AveragesBase
+    Y_SIZES = (1..300).map { |i| 1.2**i * 0.000001 }
+
+    attr_reader :data
+
+    def initialize(data)
+      @data = data
+    end
+
+    def averages_from_measurements(selected_measurements)
```
AkkarinForest: Can these methods be private?

lib/crowdmap_averages.rb
```diff
@@ -0,0 +1,105 @@
...
+      data[:tags].to_s.split(/[\s,]/)
+    end
+  end
+
+  class ForWeb < AveragesBase
+    def as_json(options = nil)
+      web_averages
+    end
+
+    def web_averages
```
AkkarinForest: Can these methods be private?

lib/crowdmap_averages.rb
czorek: Nice tip, thanks!

lib/crowdmap_averages.rb
czorek: Sure, will change that

PR #300, opened 2019-07-02
Refactor Active/Dormant 

gitjul:
- add custom type in elm
- use active/dormant naming instead of streaming/non-streaming throughout the application

### DIFF
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -115,6 +115,7 @@ type alias Flags =
     , isCrowdMapOn : Bool
     , crowdMapResolution : Int
     , timeRange : Encode.Value
+    , isActive : Bool
     , isIndoor : Bool
     , selectedSessionId : Maybe Int
     , sensors : Encode.Value
@@ -171,7 +172,12 @@ init flags url key =
         , isSearchAsIMoveOn = flags.isSearchAsIMoveOn
         , overlay = Overlay.init flags.isIndoor
         , scrollPosition = flags.scrollPosition
-        , status = Active
+        , status =
+            if flags.isActive then
+                Active
+
+            else
+                Dormant
       }
     , Cmd.batch
         [ fetchSelectedSession sensors flags.selectedSessionId flags.selectedSensorId page
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -13,6 +13,7 @@ import Data.Page exposing (Page(..))
 import Data.Path as Path exposing (Path)
 import Data.SelectedSession as SelectedSession exposing (SelectedSession)
 import Data.Session exposing (..)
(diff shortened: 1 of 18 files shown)


config/routes.rb
```diff
@@ -52,7 +52,7 @@
     resources :sensors, only: %i[index]
 
     namespace :realtime do
-      get 'streaming_sessions' => 'sessions#index_streaming'
+      get 'streaming_sessions' => 'sessions#index_active'
```
gitjul: @AkkarinForest is that alright as for leaving a working fallback? or did you have sth else in mind when you mentioned an alias?

app/javascript/elm/src/Main.elm
```diff
@@ -172,6 +171,7 @@ init flags url key =
         , isSearchAsIMoveOn = flags.isSearchAsIMoveOn
         , overlay = Overlay.init flags.isIndoor
         , scrollPosition = flags.scrollPosition
+        , status = Active
```
AkkarinForest: can you check what happens when you save and open a link with fixed dormant sessions? Cause I think it's not gonna work if you don't send it in the flags.

app/javascript/packs/elm.js
```diff
@@ -95,7 +95,7 @@ document.addEventListener("DOMContentLoaded", () => {
         resetIconWhite,
         tooltipIcon,
         heatMapThresholdValues,
-        isStreaming: data.isStreaming,
+        isActive: data.isActive,
```
AkkarinForest: If you removed it from the flags on the elm side, then you can also delete this. (but as I wrote it might be needed on the elm side)

config/routes.rb
AkkarinForest: Yes, that's what I had in mind. 
What do you think about creating a separate section at the bottom of this file, something like`#legacy api`? then we can also put there `get 's/:url_token'`. 

And it might be a good idea to leave a comment with the date that we change the endpoint? This way if someone works on this code and sees a very old date they will be more willing to remove it.

app/javascript/elm/src/TimeRange.elm
```diff
@@ -64,7 +65,7 @@ view refreshTimeRange isDisabled tooltipIcon resetIcon =
             , placeholder "time frame"
             , type_ "text"
             , name "time-range"
-            , disabled isDisabled
+            , disabled (status == Active)
```
AkkarinForest: I'm not sure about this. Maybe TmieRange should know the session status, it should only be interested in knowing if it's disabled? Similar to how you did it with `viewToggleButton`.

config/routes.rb
gitjul: added a comment but keeping it in place for now

app/javascript/elm/src/Main.elm
gitjul: you're right, thx

PR #301, opened 2019-07-02
Colours switch

AkkarinForest:
Default theme:
Screenshot 2019-07-03 at 15 24 04

Blue theme:
Screenshot 2019-07-03 at 15 23 57


app/javascript/javascript/theme.js
```diff
@@ -0,0 +1,111 @@
...
+
+const params = () =>
+  window.location.hash
+    .slice(2)
+    .split("&")
+    .filter(x => x.length !== 0)
+    .map(x => x.split("="))
+    .map(([k, v]) => [k, decodeURIComponent(v)])
+    .map(([k, v]) => [k, JSON.parse(v)])
+    .reduce((acc, [k, v]) => ({ ...acc, [k]: v }), {});
```
AkkarinForest: I'm thinking of extracting some params helpers functions soon, so I'm leaving this crude duplication for now :D cause this PR is already too long

gitjul: short PRs ftw

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -50,7 +51,8 @@ export const FixedSessionsMapCtrl = (
         mid: 35,
         high: 55,
         highest: 150
-      }
+      },
+      customTheme: false
```
3v0k4: Is `theme` the word Michael uses for this feature?

app/javascript/angular/code/controllers/_mobile_sessions_map_ctrl.js
```diff
@@ -174,6 +176,12 @@ export const MobileSessionsMapCtrl = (
 
         FiltersUtils.fetchShortUrl(tooltipId, currentUrl);
       });
+
+      elmApp.ports.toggleTheme.subscribe(isCustomThemeOn => {
+        params.update({ customTheme: isCustomThemeOn });
+        $scope.$apply();
+        applyTheme();
+      });
```
3v0k4: Just as a random thought, could we merge `_mobile_sessions_map_ctrl.js ` and `_fixed_sessions_map_ctrl.js`, add some ifs and pass from Elm the `Page` the app is on? Not sure if that would add too many ifs but it would get rid of the double controller file

app/javascript/elm/src/Data/Session.elm
```diff
@@ -67,16 +67,16 @@ classByValue average heatMapThresholds =
                     "grey-bg"
 
                 Range2 ->
-                    "green-bg"
+                    "level1-bg"
```
3v0k4: Does it make sense that `Range` is `2` and `level` is `1`?

app/javascript/elm/src/Main.elm
```diff
@@ -69,9 +70,14 @@ type alias Model =
     , scrollPosition : Float
     , debouncingCounter : Int
     , isNavExpanded : Bool
+    , isCustomThemeOn : Tagged Theme Bool
```
3v0k4: YEEEEES! I LOVE THIS PACKAGE!

app/javascript/elm/src/Main.elm
```diff
@@ -723,7 +740,16 @@ viewDocument model =
 
 view : Model -> Html Msg
 view model =
-    div [ id "elm-app" ]
+    let
+        themeClass =
+            case Tagged.untag model.isCustomThemeOn of
```
3v0k4: Have you considered a custom type for the theme? It feels like forcing a bool here. Tagged is great as a package. At the same time I feel like this is not an on / off thing. It's more like a choice. As in
```elm
type Theme
  = One -- use better name here
  | Two
```
would communicate better than `isCustomThemeOn`

3v0k4: For example Michael could say: "let's switch themes". Now custom is the one that wasn't custom :D

app/javascript/javascript/theme.js
```diff
@@ -0,0 +1,72 @@
...
+
+export const applyTheme = () => {
+  if (window.__map) {
+    updateFixedClusters();
+    updateRectangles();
+  }
+};
+
+const updateFixedClusters = () => {
+  if (window.__map.clusterers[0]) {
```
3v0k4: Is it possible `window.__map` is undefined / null here?

app/javascript/javascript/theme.js
```diff
@@ -0,0 +1,72 @@
...
+      { url: assets.clusterTheme1Level1Path, height: 30, width: 30 },
+      { url: assets.clusterTheme1Level2Path, height: 30, width: 30 },
+      { url: assets.clusterTheme1Level3Path, height: 30, width: 30 },
+      { url: assets.clusterTheme1Level4Path, height: 30, width: 30 }
+    ];
+  }
+};
+
+const updateRectangles = () => {
+  if (window.__map.rectangles) {
```
3v0k4: Have you considered an early return to keep the code flat?

```js
const f = () => {
  if (condition) return;
  // do stuff
}
```

app/javascript/javascript/theme.js
```diff
@@ -0,0 +1,72 @@
...
+      rectangle.setOptions({
+        fillColor: rectangleColour(rectangle.data.value)
+      });
+    });
+  }
+};
+
+const rectanglesStyles = () => {
+  if (params().customTheme) {
+    return [null, "#81dbcb", "#4ebcd5", "#2a70b8", "#19237e"];
```
3v0k4: Would `""` work too in place of `null`?

app/javascript/javascript/theme.js
```diff
@@ -0,0 +1,72 @@
...
+  } else {
+    return [null, "#96d788", "#ffd960", "#fca443", "#e95f5f"];
+  }
+};
+
+export const rectangleColour = value => {
+  const levels = heatLevels();
+  const level = _(levels).detect(level => value < level);
```
3v0k4: Have you considered vanilla js here? `filter` should be ok for this case

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
gitjul: yes

app/javascript/elm/src/Main.elm
gitjul: let's use Default and Blue, is that ok @AkkarinForest ?

app/javascript/elm/src/Main.elm
gitjul: will refactor `isNavExpanded` same way later

PR #303, opened 2019-07-02
Direct toggle only

gitjul:
Changes behavior of the toggle buttons in a way that double clicking a toggle option doesn't toggle to the opposite selection.

### DIFF
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -216,7 +216,7 @@ type Msg
     | SubmitLocation
     | TagsLabels LabelsInput.Msg
     | ProfileLabels LabelsInput.Msg
-    | ToggleCrowdMap
+    | ToggleCrowdMap Bool
     | UpdateCrowdMapResolution Int
     | UpdateTimeRange Encode.Value
     | RefreshTimeRange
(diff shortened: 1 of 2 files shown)


app/javascript/elm/src/Main.elm
```diff
@@ -288,8 +288,12 @@ update msg model =
             in
             ( subModel2, Cmd.batch [ subCmd1, subCmd2 ] )
 
-        ToggleCrowdMap ->
-            ( { model | isCrowdMapOn = not model.isCrowdMapOn }, Ports.toggleCrowdMap (not model.isCrowdMapOn) )
+        ToggleCrowdMap isCrowdmapOn ->
+            if model.isCrowdMapOn == isCrowdmapOn then
```
gitjul: I wonder what you think about these @3v0k4 @AkkarinForest ? not sure if this is needed or the way it should be handled

gitjul: the reason I did this was that I didn't want to do anything if you click on an already selected option

3v0k4: One alternative would be to have a `NoOp` `Msg` and in the view using that when the button should be a noop. In general I go with whatever keeps the view logic the dumbest.

Another would be to have `TurnOn...` and `TurnOff...` messages for each. That would make the view really ignorant of domain logic. It would double up the branches though. But I think that's not too bad.

If you keep this logic the only thing maybe is the naming. `isCrowdMapOn` is a bit confusing to me. That means reading `model.isCrowdMapOn == isCrowdmapOn` I get lost a bit.

Also you could extract a function like
```elm
updateIf : Bool -> Model -> (Model -> Model) -> Cmd Msg -> ( Model, Cmd Msg)
updateIf condition model updateModel msg =
  if condition then
    ( updateModel model, msg )
  else
    ( model, Cmd.none)
```
Or maybe this is too much, don't know :)

gitjul: thx for the input!

PR #305, opened 2019-07-05
Active sessions endpoint

gitjul:
This PR contains a proper refactor of fixed active sessions - added a new endpoint in a similar fashion to the dormant active sessions endpoint and cleaned up obsolete stuff. It also renames streaming to active and non-streaming to dormant everywhere.

### DIFF
--- a/app/controllers/api/fixed/active/sessions_controller.rb
+++ b/app/controllers/api/fixed/active/sessions_controller.rb
@@ -1,6 +1,6 @@
 module Api
   module Fixed
-    module Dormant
+    module Active
       class SessionsController < BaseController
         respond_to :json
 
@@ -17,7 +17,7 @@ module Api
               struct: Api::FixedSessions::Struct
             )
 
-          result = Api::ToDormantSessionsArray.new(form: form).call
+          result = Api::ToActiveSessionsArray.new(form: form).call
 
           if result.success?
             render json: result.value, status: :ok
--- a/app/controllers/api/realtime/sessions_controller.rb
+++ b/app/controllers/api/realtime/sessions_controller.rb
@@ -21,25 +21,12 @@ module Api
     class SessionsController < BaseController
       require 'uri'
 
-      INT_Q_ATTRS = %i[time_from time_to day_from day_to limit offset]
-
       # TokenAuthenticatable was removed from Devise in 3.1
       # https://gist.github.com/josevalim/fb706b1e933ef01e4fb6
       before_action :authenticate_user_from_token!, only: :create
       before_action :authenticate_user!, only: :create
 
       respond_to :json
-
-      def index_active
-        data = decoded_query_data(params[:q])
-        INT_Q_ATTRS.each { |key| data[key] = data[key].to_i if data.key?(key) }
-
-        data[:time_from] = Time.strptime(data[:time_from].to_s, '%s')
-        data[:time_to] = Time.strptime(data[:time_to].to_s, '%s')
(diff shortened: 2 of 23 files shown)


app/controllers/api/realtime/sessions_controller.rb
```diff
@@ -30,17 +30,6 @@ class SessionsController < BaseController
 
       respond_to :json
 
-      def index_streaming
-        data = decoded_query_data(params[:q])
-        INT_Q_ATTRS.each { |key| data[key] = data[key].to_i if data.key?(key) }
```
AkkarinForest: I guess you can also remove INT_Q_ATTRS on line 24

app/controllers/api/fixed/active/sessions_controller.rb
```diff
@@ -0,0 +1,32 @@
+module Api
+  module Fixed
+    module Active
+      class SessionsController < BaseController
+        respond_to :json
+        INT_Q_ATTRS = %i[time_from time_to day_from day_to limit offset]
```
AkkarinForest: is this needed?

app/javascript/elm/src/Main.elm
```diff
@@ -183,6 +183,12 @@ init flags url key =
         , overlay = Overlay.init flags.isIndoor
         , scrollPosition = flags.scrollPosition
         , theme = Theme.toTheme flags.theme
+        , status =
+            if flags.isActive then
+                Active
+
+            else
+                Dormant
```
AkkarinForest: You could move this to the Status module in a similar fashion like theme works
```
toStatus status =
    case status of
        True ->
            Active

        False ->
            Dormant

```

app/javascript/elm/src/Main.elm
```diff
@@ -417,13 +423,21 @@ update msg model =
...
                 ( subModel, subCmd ) =
                     deselectSession model
+
+                newStatus =
+                    case model.status of
+                        Active ->
+                            Dormant
+
+                        Dormant ->
+                            Active
```
AkkarinForest: Same here, you can extract this into a function in a status module
```
toggle : Theme -> Theme
toggle theme =
    case theme of
        Default ->
            Blue

        Blue ->
            Default
```

app/javascript/elm/src/Main.elm
```diff
@@ -13,6 +13,7 @@ import Data.Page exposing (Page(..))
 import Data.Path as Path exposing (Path)
 import Data.SelectedSession as SelectedSession exposing (SelectedSession)
 import Data.Session exposing (..)
+import Data.Status exposing (Status(..))
```
AkkarinForest: In general, you could create helpers functions in a Status module in a way that you wouldn't even have to expose the internals of status - I mean the `(..)`
It's more code, but I'm a fan of these small methods. It feels right :)
You can even write very small thingies like `default = Active`
Check next comments and a Theme module if you want to see what I mean.

app/models/api/active_session.rb
```diff
@@ -1,7 +1,7 @@
 require 'dry-validation'
 require 'dry-struct'
 
-module Api::DormantSessions
+module Api::ActiveSessions
```
AkkarinForest: Is this file used?

gitjul: no:)

app/controllers/api/fixed/active/sessions_controller.rb
gitjul: no:)

PR #308, opened 2019-07-08
Direct toggle only

gitjul:
Changes behavior of the toggle buttons in a way that double clicking a toggle option doesn't toggle to the opposite selection.

This PR re-applies changes made in https://github.com/HabitatMap/AirCasting/pull/303 which had to be reverted.

### DIFF
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -222,7 +222,7 @@ type Msg
     | SubmitLocation
     | TagsLabels LabelsInput.Msg
     | ProfileLabels LabelsInput.Msg
-    | ToggleCrowdMap
+    | ToggleCrowdMap Bool
     | UpdateCrowdMapResolution Int
     | UpdateTimeRange Encode.Value
     | RefreshTimeRange
@@ -236,8 +236,8 @@ type Msg
     | UpdateSessions Encode.Value
     | LoadMoreSessions
     | UpdateIsHttping Bool
-    | ToggleIndoor
-    | ToggleStatus
+    | ToggleIndoor Bool
+    | ToggleStatus Status
     | DeselectSession
     | ToggleSessionSelectionFromAngular (Maybe Int)
     | ToggleSessionSelection Int
@@ -295,8 +295,12 @@ update msg model =
             in
             ( subModel2, Cmd.batch [ subCmd1, subCmd2 ] )
 
-        ToggleCrowdMap ->
-            ( { model | isCrowdMapOn = not model.isCrowdMapOn }, Ports.toggleCrowdMap (not model.isCrowdMapOn) )
+        ToggleCrowdMap newValue ->
+            if model.isCrowdMapOn == newValue then
+                ( model, Cmd.none )
+
+            else
+                ( { model | isCrowdMapOn = newValue }, Ports.toggleCrowdMap newValue )
 
         UpdateCrowdMapResolution resolution ->
             let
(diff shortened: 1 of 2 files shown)


app/javascript/elm/src/Main.elm
```diff
@@ -403,32 +407,36 @@ update msg model =
...
             let
                 ( subModel, subCmd ) =
                     deselectSession model
             in
-            if subModel.isIndoor then
-                ( { subModel | isIndoor = False, overlay = Overlay.update (RemoveOverlay IndoorOverlay) model.overlay }
-                , Cmd.batch [ Ports.toggleIndoor False, subCmd ]
-                )
+            if subModel.isIndoor == newValue then
+                ( subModel, Cmd.none )
```
AkkarinForest: If you are using submodel, then execute the subCmd as well, together they are responsible for deselecting a session. Right now there is a bug:
- go to fix active
- select a session
- click active again
- looks like session is deselected
- refresh the page
- it's still selected

app/javascript/elm/src/Main.elm
```diff
@@ -403,32 +407,36 @@ update msg model =
...
                     deselectSession model
-
-                newStatus =
-                    Status.toggle model.status
             in
-            ( { subModel | status = newStatus }
-            , Cmd.batch [ Ports.toggleActive (Status.toBool newStatus), subCmd ]
-            )
+            if subModel.status == newStatus then
+                ( subModel, Cmd.none )
```
AkkarinForest: subCmd here as well

app/javascript/elm/src/Main.elm
```diff
@@ -295,8 +295,12 @@ update msg model =
             in
             ( subModel2, Cmd.batch [ subCmd1, subCmd2 ] )
 
-        ToggleCrowdMap ->
-            ( { model | isCrowdMapOn = not model.isCrowdMapOn }, Ports.toggleCrowdMap (not model.isCrowdMapOn) )
+        ToggleCrowdMap newValue ->
+            if model.isCrowdMapOn == newValue then
+                ( model, Cmd.none )
```
AkkarinForest: What happens if you delete this branch of if else and always execute line 303 - does the behaviour changes? Same for the other toggles. Maybe it's not worth having it. I'm not sure, we can chat about it.

app/javascript/elm/src/Main.elm
```diff
@@ -403,32 +407,36 @@ update msg model =
...
+            else if newValue then
                 ( { subModel | isIndoor = True, profiles = LabelsInput.empty, overlay = Overlay.update (AddOverlay IndoorOverlay) model.overlay }
                 , Cmd.batch [ Ports.toggleIndoor True, Ports.updateProfiles [], subCmd ]
                 )
 
-        ToggleStatus ->
+            else
+                ( { subModel | isIndoor = False, overlay = Overlay.update (RemoveOverlay IndoorOverlay) model.overlay }
+                , Cmd.batch [ Ports.toggleIndoor False, subCmd ]
+                )
```
AkkarinForest: You could also use `case` instead of multiple if-else branches.

app/javascript/elm/src/Main.elm
gitjul: I wanted to cause no additional unnecessary action. yep, let's talk

app/javascript/elm/src/Main.elm
gitjul: fixed so that nothing is happening

app/javascript/elm/src/Main.elm
gitjul: fyi after talking offline went with returning the model instead

PR #310, opened 2019-07-10
Responsive map

gitjul:


app/javascript/elm/src/Main.elm
```diff
@@ -1145,16 +1155,26 @@ viewSensorFilter sensors selectedSensorId tooltipIcon =
...
         ]
 
 
+viewSensorPopup : Bool -> Popup.Popup -> Popup.PopupStatus -> Html Msg
+viewSensorPopup isPopupExtended popup popupStatus =
+    if Popup.isSensorPopupShown popupStatus then
+        Popup.view TogglePopupState SelectSensorId isPopupExtended popup
+
+    else
+        text ""
```
AkkarinForest: I don't know the contex, so maybe this change is needed for something else, but just looking at this commit it seems like there is no reason to have `viewSensorPopup ` and `viewParameterPopup `. This could be just:
```
viewWhateverPopup : Bool -> Popup.Popup -> Popup.PopupStatus -> Html Msg
viewWhateverPopup isPopupExtended popup popupStatus =
    if Popup.isWhateverShown popupStatus then
        Popup.view TogglePopupState SelectSensorId isPopupExtended popup

     else
        text ""
```

but then it doesn't make sense to have this `Popup.PopupStatus` since `Popup.view` already knows that it should show `text ""` if no popup is selected (see in popup module that `Popup.Popup` has two states and one of them is `None`)

app/javascript/elm/src/Popup.elm
```diff
@@ -16,6 +16,12 @@ type PopupPart
     | OtherPart
 
 
+type PopupStatus
+    = ParameterPopupShown
+    | SensorPopupShown
+    | PopupHidden
```
AkkarinForest: this seems to similar to `type Popup`. If indeed parameter and sensor popup are different then I would just have one type:
```
type Popup
    = ParameterList someArgs
    | SensorList someDifferentArgs
    | None
```
if they are the same then what we already have on line 9 should be enough:
```
type Popup
    = SomeList someArgs
    | None
```

app/javascript/elm/src/Main.elm
```diff
@@ -1078,9 +1079,8 @@ viewMobileFilters model =
 viewFixedFilters : Model -> Html Msg
 viewFixedFilters model =
     div [ class "filters-container" ]
-        [ viewParameterFilter model.sensors model.selectedSensorId model.tooltipIcon
-        , viewSensorFilter model.sensors model.selectedSensorId model.tooltipIcon
-        , viewLocationFilter model.location model.isIndoor model.tooltipIcon
```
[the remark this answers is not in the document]
gitjul: oops, no!

app/javascript/elm/src/Main.elm
gitjul: the thing is that I need to know not just if there's a popup, but which popup it is. hence all the changes.

app/javascript/elm/src/Popup.elm
gitjul: the former sounds like the way to go.

app/javascript/elm/src/Popup.elm
```diff
@@ -16,38 +18,38 @@ type PopupPart
...
-        None ->
-            text ""
+                  else
+                    togglePopupStateButton ("more " ++ itemType) toggle
+                ]
+
+
+viewEmailForm : Html msg -> Html msg
+viewEmailForm emailForm =
+    emailForm
```
AkkarinForest: I wonder if we need this function since it doesn't really do anything?

app/javascript/elm/src/Main.elm
```diff
@@ -331,14 +339,42 @@ update msg model =
         ShowCopyLinkTooltip tooltipId ->
             ( model, Ports.showCopyLinkTooltip tooltipId )
 
-        ShowPopup items itemType selectedItem ->
-            ( { model | popup = Popup.SelectFrom items itemType selectedItem, isPopupExtended = False, overlay = Overlay.update (AddOverlay PopupOverlay) model.overlay }, Cmd.none )
+        ShowListPopup popup ->
+            ( { model | popup = popup, isPopupListExpanded = False, overlay = Overlay.update (AddOverlay PopupOverlay) model.overlay }, Cmd.none )
+
+        ShowExportPopup ->
+            ( { model | popup = Popup.EmailForm }, Cmd.none )
```
AkkarinForest: There is this one small thing that has to do with the overlays - 
on session list view click on paramter - you have a list popup and a gray overlay over the map
- then click on export sessions - the list popup dissapears but the gray overlay remains.
Not a big thing, so not sure if it's worth changing, unless that's something easy. I think adding `overlay = Overlay.update (RemoveOverlay PopupOverlay) model.overlay` could work. (similar to what we have in `ClosePopup` branch)

app/javascript/elm/src/Popup.elm
AkkarinForest: I wonder if we need this function, cause it doesn't do anything?

app/javascript/elm/src/Main.elm
AkkarinForest: There is this one small thing that has to do with the overlays - 
on session list view click on parameter field - you have a list popup and a gray overlay over the map
- then click on export sessions - the list popup disappears but the gray overlay remains.
Not a big thing, so not sure if it's worth changing unless that's something easy. I think adding 
`overlay = Overlay.update (RemoveOverlay PopupOverlay) model.overlay` 
could work. (similar to what we have in `ClosePopup` branch).
There might be some other edge cases when switching between overlays, but I only encountered this one so far.

gitjul: nice catch

app/javascript/elm/src/Popup.elm
gitjul: 😅

PR #311, opened 2019-07-10
Syncing mobile app sessions

AkkarinForest:
We decided to separate updating a session that was edited in the mobile app form syncing the session list. I'm still leaving the old endpoint to support old versions of mobile apps.

We added versioning to sessions to make sure that when two users use the same account the sessions would be synced correctly. So now every time session is updated via `update_session` endpoint we bump the version. And every time a user hits `sync2` we check if they have the newest version and if not we add this to the `download` array.

### DIFF
--- a/app/models/api/user_session.rb
+++ b/app/models/api/user_session.rb
@@ -1,0 +1,52 @@
+require 'dry-validation'
+require 'dry-struct'
+
+module Api::UserSession
+  module Types
+    include Dry::Types.module
+  end
+
+  StreamSchema =
+    Dry::Validation.Schema do
+      required(:deleted).filled(:bool?)
+      required(:average_value).filled(:float?)
+      required(:measurement_type).filled(:str?)
+      required(:sensor_package_name).filled(:str?)
+      required(:sensor_name).filled(:str?)
+    end
+
+  Schema =
+    Dry::Validation.Schema do
+      configure do
+        def self.messages
+          super.merge(en: { errors: { valid_streams: 'streams is not valid' } })
+        end
+      end
+
+      required(:uuid).filled(:str?)
+      required(:tag_list)
+      required(:title)
+      required(:notes).each do
+        schema do
+          required(:number).filled(:int?)
+          required(:text).filled(:str?)
+        end
+      end
+      validate(valid_streams: :streams) do |stream|
+        stream.all? do |sensor_name, stream_data|
+          StreamSchema.call(stream_data).success?
+        end
+      end
+    end
+
+  class Struct < Dry::Struct
+    attribute :uuid, Types::Strict::String
+    attribute :title, Types::Strict::String
+    attribute :tag_list, Types::Strict::String
+    attribute :notes, Types::Strict::Array do
+      attribute :number, Types::Strict::Integer
+      attribute :text, Types::Strict::String
+    end
+    attribute :streams, Types::Strict::Hash
+  end
+end
--- a/app/models/api/user_sessions2.rb
+++ b/app/models/api/user_sessions2.rb
@@ -1,0 +1,25 @@
+require 'dry-validation'
+require 'dry-struct'
+
+module Api::UserSessions2
+  module Types
+    include Dry::Types.module
+  end
+
+  Schema =
+    Dry::Validation.Schema do
+      required(:data).each do
+        schema do
+          required(:uuid).filled(:str?)
+          required(:deleted).filled(:bool?)
+        end
+      end
+    end
+
+  class Struct < Dry::Struct
+    attribute :data, Types::Array do
+      attribute :uuid, Types::Strict::String
+      attribute :deleted, Types::Strict::Bool
+    end
+  end
+end
--- a/app/models/api/user_sessions2.rb
+++ b/app/models/api/user_sessions2.rb
@@ -12,6 +12,7 @@ module Api::UserSessions2
         schema do
           required(:uuid).filled(:str?)
           required(:deleted).filled(:bool?)
+          required(:version).filled(:int?)
         end
       end
     end
@@ -20,6 +21,7 @@ module Api::UserSessions2
     attribute :data, Types::Array do
       attribute :uuid, Types::Strict::String
       attribute :deleted, Types::Strict::Bool
+      attribute :version, Types::Strict::Integer
     end
   end
 end
--- a/app/services/api/to_user_sessions_hash2.rb
+++ b/app/services/api/to_user_sessions_hash2.rb
@@ -11,8 +11,8 @@ class Api::ToUserSessionsHash2
 
     Success.new(
       {
-        upload: new_in_mobile_app,
-        download: new_in_db + outdated,
+        upload: new_in_params,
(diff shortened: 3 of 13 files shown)


app/models/api/user_sessions2.rb
```diff
@@ -0,0 +1,27 @@
...
+  Schema =
+    Dry::Validation.Schema do
+      required(:data).each do
+        schema do
+          required(:uuid).filled(:str?)
+          required(:deleted).filled(:bool?)
+          required(:version).filled(:int?)
+        end
+      end
+    end
```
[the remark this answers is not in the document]
3v0k4: There's also one for JSON https://dry-rb.org/gems/dry-schema/json/

app/models/api/user_session.rb
```diff
@@ -0,0 +1,52 @@
...
+      required(:measurement_type).filled(:str?)
+      required(:sensor_package_name).filled(:str?)
+      required(:sensor_name).filled(:str?)
+    end
+
+  Schema =
+    Dry::Validation.Schema do
+      configure do
+        def self.messages
+          super.merge(en: { errors: { valid_streams: 'streams is not valid' } })
```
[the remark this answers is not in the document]
3v0k4: Does this drop the error that was triggered in the nested schema? Is there a way to preserve that?

app/services/api/to_user_sessions_hash2.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def call
+    return Failure.new(form.errors) if form.invalid?
+
+    delete_sessions(data.select(&:deleted))
+
+    Success.new(
+      {
+        upload: new_in_mobile_app,
```
3v0k4: Not sure I like the mobile app part in the name. The endpoint is open to any type of client

app/services/api/to_user_sessions_hash2.rb
```diff
@@ -0,0 +1,73 @@
...
+        download: new_in_db + outdated,
```
3v0k4: Have you considered `database` instead of `db`? I find it makes things more clear, helps with autocomplete, grepping and such stuff

app/services/api/to_user_sessions_hash2.rb
```diff
@@ -0,0 +1,73 @@
...
+  def data
+    form.to_h.data
+  end
+
+  def delete_sessions(sessions)
+    user.sessions.where(uuid: sessions.map(&:uuid)).destroy_all
+  end
+
+  def deleted
+    DeletedSession.where(user: user, uuid: data.map(&:uuid)).map(&:uuid)
```
3v0k4: Not sure it changes much but I maybe you could use `.pluck(:uuid)` instead of `.map` at the end

app/services/api/to_user_sessions_hash2.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def new_in_mobile_app
+    uuids_present_in_mobile_app - uuids_present_in_db - deleted
+  end
+
+  def uuids_present_in_mobile_app
+    uuids_present_in_mobile_app ||= present_in_mobile_app.map(&:uuid)
```
3v0k4: Why the `=`?

3v0k4: Not really sure what this does to be honest :D

app/services/api/to_user_sessions_hash2.rb
```diff
@@ -0,0 +1,73 @@
...
+    present_in_db.select do |session_in_db|
+      if uuids_present_in_mobile_app.exclude?(session_in_db.uuid)
+        false
+      else
+        session_in_db.version >
+          present_in_mobile_app.detect do |sessio_in_mobile_app|
+            sessio_in_mobile_app.uuid == session_in_db.uuid
+          end.version
+      end
+    end.map(&:uuid)
```
3v0k4: You could break it up a bit here

```
present_in_db
  .select { |s| uuids_present_in_mobile_app.exclude?(s.uuid) }
  ...
```

3v0k4: Less performant but prolly in this case it does not matter

3v0k4: Is detect the same as find? If so please consider using one (I like find)

app/services/api/to_user_sessions_hash2.rb
```diff
@@ -0,0 +1,73 @@
+class Api::ToUserSessionsHash2
```
3v0k4: Maybe you could think of some refactoring here. Don't have any ideas at the moment. But available to sit together if you want

3v0k4: You can always extract services if that helps. Prolly not, I feel like the problem here is that we are fighting against OOP :)

app/models/api/user_sessions2.rb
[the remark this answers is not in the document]
AkkarinForest: Awesome! I'll investigate that asap.

app/services/api/to_user_sessions_hash2.rb
AkkarinForest: Detect find the first occurrence. So it's equivalent to `find.first`.

app/services/api/to_user_sessions_hash2.rb
AkkarinForest: Actually, I forgot about @ at the beginning of `@uuids_present_in_mobile_app`
The idea is that the calculation on the right only happens once, on the first call to the method. The next time the value is reused. Of course, it doesn't work if it's a local variable 🤦‍♂

spec/controllers/api/user_sessions_controller_spec.rb
```diff
@@ -165,8 +165,185 @@
...
+
+      post :update_session,
+           params: {
+             data: {
+               uuid: session.uuid,
+               title: session.title,
+               tag_list: session.tag_list.to_s,
+               notes: [],
+               streams: {
+                 whatever: {
```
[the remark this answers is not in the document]
AkkarinForest: Sure, I put it as a placeholder and forgot about that :D The fact is that we are only processing the values from this streams hash, the keys are ignored in the schema validation.

app/models/api/user_sessions2.rb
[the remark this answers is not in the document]
3v0k4: Let me know if you find out anything new!

app/services/api/to_user_sessions_hash2.rb
3v0k4: Just checked, I think you mean `select().first`. `find` seems to be the same as `detect`

app/models/api/user_session.rb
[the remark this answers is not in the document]
AkkarinForest: There is an open issue about handling validation for dynamic keys: https://github.com/dry-rb/dry-schema/issues/37
I'll make a note to check if it's resolved and update our gems if it's ready.

PR #313, opened 2019-07-12
Return unauthorized for incorrect credentials

AkkarinForest:
When logging in the mobile app we want to return a json response with
status unauthorized.
When logging from the web, eg in admin panel we want to redirect.

Adopted from: https://stackoverflow.com/a/43336121


app/services/custom_failure.rb
```diff
@@ -0,0 +1,15 @@
+class CustomFailure < Devise::FailureApp
```
gitjul: maybe sth more specific? like AuthorizationError/Failure?

app/services/custom_failure.rb
```diff
@@ -0,0 +1,15 @@
...
+  def respond
+    if http_auth?
+      http_auth
```
gitjul: I like early returns (`return http_auth if http_auth?`) instead of more `if` branches but it's a matter of preference :)

AkkarinForest:
Thanks for the review. I'm gonna leave it as it is, so it resembles the docs. Might make it easier if someone will be editing in the future.
https://github.com/plataformatec/devise/wiki/How-To:-Redirect-to-a-specific-page-when-the-user-can-not-be-authenticated

PR #314, opened 2019-07-17
User settings

AkkarinForest:
This logic allows the mobile app to change user settings regarding the email alert about a session that stopped streaming.


spec/controllers/api/users_controller_spec.rb
```diff
@@ -63,4 +45,22 @@
       it { is_expected.to respond_with(:unprocessable_entity) }
     end
   end
+
+  describe '#settings' do
+    let!(:user) { create_user!(session_stopped_alert: false) }
```
gitjul: `session_stopped_alert` means that the user wants to be notified when the session is stopped, right? I'd consider finding a better name for this flag

gitjul: you know, so it's more obvious that this is about notifications being turned on or off and not e.g. that sth was sent or whatever

AkkarinForest: yeah, I agree it's not good, but I couldn't figure out anything else :(

PR #317, opened 2019-07-18
Email alerts

AkkarinForest:
Adds email service and email templates for alerts about sessions that stopped streaming.


app/views/layouts/mailer.html.erb
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <style>
+      /* Email styles need to be inline */
```
gitjul: unless there needs to be styling (check with Michael?) I'd remove this whole `<style>` section

app/views/user_mailer/session_stopped_email.html.erb
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <p> <%= @title %> stopped streaming at <%= @time %> </p>
```
gitjul: I'd remove the whitespace after `<p>` and before `</p>` to avoid rendering them in the email

config/environments/development.rb
```diff
@@ -35,6 +35,10 @@
 
   config.action_mailer.perform_caching = false
 
+  config.action_mailer.delivery_method = :smtp
```
gitjul: if there's styling needed we should use mailcatcher or some other service to be able to preview the emails in development

AkkarinForest: There is no styling needed. But I was using mailcatcher, these lines are the needed setup :)

PR #318, opened 2019-07-30
Use empty array as default when no data provided in params

AkkarinForest:
Solves this crash: https://app.honeybadger.io/projects/47179/faults/51247617


spec/controllers/api/user_sessions_controller_spec.rb
```diff
@@ -15,6 +15,12 @@
       expect(json_response).to eq(expected)
     end
 
+    it "doesn't crush when to data provided in the params" do
```
gitjul: crash :) and "to" -> "no", right?

PR #319, opened 2019-07-30
Update analytics

Sonje:
Ga.js is outdated so I'm upgrading to analytics.js. 
The analytics.js library is the latest JavaScript library for Google Analytics.


Sonje:
Instead of analytics.js I updated the code with gtag.js which was recommended by google. It should work properly now.

gitjul:
👀

PR #322, opened 2019-08-02
Adjust heat map levels

AkkarinForest:
Added a button that adjusts heatmap levels to fit the measurements that are currently displayed on the graph.
It's only visible on a single session view.

Then I changed what happens to the graph when heat map level change. Cause before we were reloading the entire graph. This had a side effect of resetting the graph zoom level to default (and also took a lot of time to fetch the measurement again). Now we only update the graph y-axis options.

### DIFF
--- a/app/javascript/elm/src/Data/GraphData.elm
+++ b/app/javascript/elm/src/Data/GraphData.elm
@@ -1,4 +1,4 @@
-module Data.GraphData exposing (GraphData)
+module Data.GraphData exposing (GraphData, GraphHeatData)
 
 
 type alias GraphData =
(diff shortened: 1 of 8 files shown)


app/javascript/elm/src/Data/GraphData.elm
```diff
@@ -1,24 +1,27 @@
...
-        { threshold1 : Int
-        , threshold5 : Int
-        , levels :
-            List
-                { from : Int
-                , to : Int
-                , className : String
-                }
-        }
+    , heat : GraphHeatData
```
gitjul: I'd probably name this `HeatGraphData` but I'm not sure

app/javascript/elm/tests/Data/HeatMapThresholdsTests.elm
```diff
@@ -42,4 +42,52 @@ suite =
...
+                    |> fitThresholds bounds
+                    |> toValues
+                    |> Expect.equal
+                        { threshold1 = 0
+                        , threshold2 = 1
+                        , threshold3 = 2
+                        , threshold4 = 3
+                        , threshold5 = 4
+                        }
+        , test "fitThresholds sets the uppper threshold always bigger than max bound and bottom threshold always smaller than min bound" <|
```
gitjul: one `p` too many in `uppper` :)

gitjul: is that a correct description? what if max == upper or bottom == min?

PR #325, opened 2019-08-07
Center blue marker

AkkarinForest:
Now the marker location is based on the cluster and not on the individual session it highlights so they are all in the same place. 

Still, no matter what I tried it always seems to be a bit off-center (but all of them exactly the same direction). But moving it by one pixel is to much. And it's not possible to move it by half a pixel :/ maybe it's good enough?


app/assets/images/pulsing_location_marker.svg
```diff
@@ -1,3 +1,3 @@
-<svg width="16" height="16" xmlns="http://www.w3.org/2000/svg">
-  <circle cx="8" cy="8" r="8" fill="#09a8f0"/>
-</svg>
\ No newline at end of file
+<svg width="50" height="50" xmlns="http://www.w3.org/2000/svg">
+  <circle transform="translate(19,19)" cx="6" cy="6" r="6" fill="#09a8f0"/>
```
AkkarinForest: translate moves the circle to the middle of the svg
formula (width / 2) - radius = (50 / 2) - 6) = 19

PR #332, opened 2019-08-14
Trigger autocomplete with empty input

AkkarinForest:
Also sorts the tags alphabetically, but I added a bit more logic there so tags that have numbers or some weird characters are at the end.

### DIFF
--- a/app/javascript/javascript/filtersUtils.js
+++ b/app/javascript/javascript/filtersUtils.js
@@ -74,21 +74,44 @@ export const setupTimeRangeFilter = (
     );
   }
 };
-export const setupAutocomplete = (callback, id, path, createParams) => {
-  if (document.getElementById(id)) {
-    $(`#${id}`).autocomplete({
+export const setupTagsAutocomplete = (callback, path, createParams) => {
+  if (document.getElementById("tags")) {
+    $(`#tags`)
+      .autocomplete({
+        source: function(request, response) {
+          const data = {
+            q: { input: request.term, ...createParams() }
+          };
+          $.getJSON(path, data, response);
+        },
+        select: function(event, ui) {
+          callback(ui.item.value);
+        },
+        minLength: 0
+      })
+      .focus(function() {
+        $(this).autocomplete("search");
+      });
+  } else {
+    window.setTimeout(setupAutocomplete(callback, path), 100);
+  }
+};
+
+export const setupProfileNamesAutocomplete = callback => {
+  if (document.getElementById("profile-names")) {
+    $(`#profile-names`).autocomplete({
       source: function(request, response) {
         const data = {
-          q: { input: request.term, ...createParams() }
+          q: { input: request.term }
         };
-        $.getJSON(path, data, response);
+        $.getJSON("api/autocomplete/usernames", data, response);
       },
       select: function(event, ui) {
         callback(ui.item.value);
       }
     });
   } else {
-    window.setTimeout(setupAutocomplete(callback, id, path), 100);
+    window.setTimeout(setupAutocomplete(callback), 100);
   }
 };
 
--- a/app/models/api/mobile_tags_params.rb
+++ b/app/models/api/mobile_tags_params.rb
@@ -8,7 +8,7 @@ module Api::MobileTagsParams
 
   Schema =
     Dry::Validation.Schema do
-      required(:input).filled(:str?)
+      required(:input)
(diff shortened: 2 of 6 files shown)


app/javascript/javascript/filtersUtils.js
```diff
@@ -74,21 +74,44 @@ export const setupTimeRangeFilter = (
...
+          };
+          $.getJSON(path, data, response);
+        },
+        select: function(event, ui) {
+          callback(ui.item.value);
+        },
+        minLength: 0
+      })
+      .focus(function() {
+        $(this).autocomplete("search");
```
AkkarinForest: this two line trigger autocomplete before you start typing

app/javascript/javascript/filtersUtils.js
```diff
@@ -74,21 +74,44 @@ export const setupTimeRangeFilter = (
...
+      });
+  } else {
+    window.setTimeout(setupAutocomplete(callback, path), 100);
+  }
+};
+
+export const setupProfileNamesAutocomplete = callback => {
+  if (document.getElementById("profile-names")) {
+    $(`#profile-names`).autocomplete({
```
gitjul: how about using `document.getElementById("profile-names")` here as well, possibly keeping it in a variable above the if?

app/models/api/mobile_tags_params.rb
```diff
@@ -8,7 +8,7 @@ module Types
 
   Schema =
     Dry::Validation.Schema do
-      required(:input).filled(:str?)
+      required(:input)
```
gitjul: is it still required though? what's the value when it's not filled with a string?

app/javascript/javascript/filtersUtils.js
```diff
@@ -74,21 +74,44 @@ export const setupTimeRangeFilter = (
...
-        $.getJSON(path, data, response);
+        $.getJSON("api/autocomplete/usernames", data, response);
       },
       select: function(event, ui) {
         callback(ui.item.value);
       }
     });
   } else {
-    window.setTimeout(setupAutocomplete(callback, id, path), 100);
+    window.setTimeout(setupAutocomplete(callback), 100);
```
gitjul: hmm where's the definition of this function? sorry if I'm blind

app/javascript/javascript/filtersUtils.js
gitjul: then we could get rid of jquery in one more place :)

app/javascript/javascript/filtersUtils.js
```diff
@@ -74,21 +74,44 @@ export const setupTimeRangeFilter = (
...
+export const setupProfileNamesAutocomplete = callback => {
+  if (document.getElementById("profile-names")) {
+    $(`#profile-names`).autocomplete({
       source: function(request, response) {
         const data = {
-          q: { input: request.term, ...createParams() }
+          q: { input: request.term }
         };
-        $.getJSON(path, data, response);
+        $.getJSON("api/autocomplete/usernames", data, response);
```
gitjul: would be great to switch to fetch (whatwg-fetch) which we're already using in other places, e.g. `elm.js`
https://github.com/github/fetch

this is a cool source when you suspect you don't need jquery somewhere: http://youmightnotneedjquery.com/

app/javascript/javascript/filtersUtils.js
AkkarinForest: Doesn't work. This library only works with elements found by jquery. Cause `Uncaught TypeError: document.getElementById(...).autocomplete is not a function`

app/models/api/mobile_tags_params.rb
AkkarinForest: It means you can pass `"input":""`, but the keyword input has to be in the params. It also means that in the code later I can do `form[:input]` without  checking if it exists.

app/javascript/javascript/filtersUtils.js
AkkarinForest: nowhere 😱  good catch!

app/javascript/javascript/filtersUtils.js
AkkarinForest: Again, not sure if it's possible with this jquery library. But we can try get rid of it totally :)

app/javascript/javascript/filtersUtils.js
gitjul: 😿

PR #338, opened 2019-08-20
Fix graph rendering for fixed sessions.

AkkarinForest:
After adding the responsive rules to graph options the graph stopped rendering on small screens. These options seem to set extremes and so they trigger `afterSetExtremes`, but at that time the `chart` variable is null or `{}` so it caused a crash.


gitjul:
👀 
can't say I know what's going on here exactly but I'm glad you fixed this :)

PR #341, opened 2019-08-20
Persist params

AkkarinForest:


app/javascript/javascript/params.js
```diff
@@ -5,5 +5,31 @@ export const getParams = () =>
     .filter(x => x.length !== 0)
     .map(x => x.split("="))
     .map(([k, v]) => [k, decodeURIComponent(v)])
-    .map(([k, v]) => [k, JSON.parse(v)])
+    .map(([k, v]) => [k, parseIfJSON(v)])
```
AkkarinForest: Because some of our params are like `data = {a:1, b:2}` but some are like `theme = default`. So in the latter case you can't really do `JSON.parse("default")`.

But hopefully, this is just a temporary solution, like we discussed maybe we will remove the `data` params

app/javascript/angular/code/controllers/_fixed_sessions_map_ctrl.js
```diff
@@ -53,8 +53,7 @@ export const FixedSessionsMapCtrl = (
         mid: 35,
         high: 55,
         highest: 150
-      },
-      theme: DEFAULT_THEME
```
AkkarinForest: These are added in the `data` part of params, it should have never been here.

app/javascript/elm/src/Main.elm
```diff
@@ -840,7 +843,7 @@ viewDocument model =
 view : Model -> Html Msg
 view model =
     div [ id "elm-app", class (Theme.toString model.theme) ]
-        [ viewNav model.navLogo model.filterIcon model.menuIcon model.areFiltersExpanded model.isNavExpanded
```
AkkarinForest: because it was not used

app/javascript/javascript/params.js
```diff
@@ -5,5 +5,41 @@ export const getParams = () =>
...
+const parseIfJSON = value => {
+  if (isJSON(value)) {
+    return JSON.parse(value);
+  } else {
+    return value;
+  }
+};
+
+export const isJSON = text => {
+  if (
```
AkkarinForest: I just copied this regex as a recommended way to check if something is a json :)

app/javascript/elm/src/Main.elm
```diff
@@ -1125,13 +1128,26 @@ viewSessionTypeNav : Model -> Html Msg
 viewSessionTypeNav model =
     ul [ class "session-type-nav" ]
         [ li [ classList [ ( "session-type-nav__item", True ), ( "selected", model.page == Mobile ) ] ]
-            [ a [ href "/mobile_map" ]
+            [ a
+                [ href
+                    (Url.Builder.absolute [ "mobile_map#" ]
+                        [ Url.Builder.string "keepFiltersExpanded" "true"
+                        , Url.Builder.string "theme" (Theme.toString model.theme)
+                        ]
```
gitjul: neat!

PR #342, opened 2019-08-20
Replace fit scale to stream measurements icon

gitjul:
This PR replaces the "Fit scale to stream measurements" icon with a better one designed by Pina and also fixes the heatmap icons' alignment in Chrome.

Screenshot 2019-08-20 at 18 04 32


AkkarinForest:
There is something weird going on. When I first select a session the arrows are very thin. After I refresh the page they look correct, like on your screenshot. Tested on clone.

Screen Shot 2019-08-21 at 10 30 12

gitjul:
@AkkarinForest yeah. did you try to refresh without cache? seems to fix it for me

AkkarinForest:
I tested in an incognito window. Every time I open a new window it's thin for the first time. After I refresh it's thicker.

gitjul:
@AkkarinForest hmm... can't really reproduce, right now I see it smaller for a second while the session loads - as if the class was added with a delay - but in no other case. let's see if it will happen after deploying.

gitjul:
@AkkarinForest the size issue is fixed :)

PR #344, opened 2019-08-21
Place autocomplete

gitjul:
This PR turns the location field into a Google Place Autocomplete input.


app/javascript/angular/code/services/google/_map.js
```diff
@@ -71,17 +84,24 @@ export const map = (
...
-      if (!address) return;
-
-      const callback = (results, status) => {
-        if (!googleMaps.wasGeocodingSuccessful(status)) return;
-
-        const latLngBounds = results[0].geometry.viewport;
-        this._fitBoundsWithoutPanOrZoomCallback(latLngBounds);
-      };
+    _goToAddress: function(place) {
+      if (!place.geometry) {
```
AkkarinForest: Not sure how this works, but when I type something invalid and press enter I see an error in the console.

gitjul:
> The only thing I noticed is that when you use tab to select from the dropdown then it switches the focus to the tooltip, seems like an unexpected side effect.

Hmm... You're right, this is what is happening. But do you actually select with tab? I use the up/down arrows, wouldn't think of using the tab.

> Also, what do you think about the little markers on the left side of the suggestions? The red one doesn't really fit our design and colors...

Wouldn't try hacking the regular implementation. Doesn't bug me personally, they're pretty delicate.

app/javascript/angular/code/services/google/_map.js
gitjul: That is what should happen when you type sth invalid :)

gitjul: I mean, I left the error to indicate that. It doesn't need to be written out but thought it might be helpful (but I guess it wasn't?).

AkkarinForest: No, I just have this instinct that warnings are okish, but an error indicates something that should never happen just happend. Not a big deal though.

gitjul: hmmm ah, ok, I know what you mean. I'll check that

gitjul: Screenshot 2019-08-26 at 18 09 51

so you're seeing an error besides the above text?

gitjul: aaaaah, the error (`Uncaught Error: Trying to send an unexpected type of value through port locationUpdated:) is showing in Chrome only...

gitjul: we should take care of this, I'll add it to the board for now cos I need to finish other things

AkkarinForest: Yes, I see this error. 
Sound like a good plan.

PR #346, opened 2019-08-26
UI: tweak mobile site navigation

gitjul:
- change expanded site nav background to dark blue and text color to white
- use alternative hamburger icon background images
- move hamburger icon to far left
- use filter icon as a background image


app/assets/stylesheets/modules/site-nav-maps.css.scss
```diff
@@ -0,0 +1,12 @@
...
+
+  .menu-collapsed & {
+    background-image: url("../../images/icons/menu-icon.svg");
+  }
+}
+
+.nav__menu-button--filter {
+  background: url("../../images/icons/filter-icon.svg") center / auto 17px
+    no-repeat;
+}
```
gitjul: these things are in separate files loaded just for maps OR static pages because of the different image paths

gitjul: like I said I didn't bother with organizing the stylesheets for maps and static in a neater way at this point, but will when more things like this occur. although with the redesign coming there's little sense to do this for things that will change

PR #347, opened 2019-08-27
Show mobile/fixed and parameter/sensor info in mobile header

gitjul:
Screenshot 2019-08-27 at 18 16 37

### DIFF
--- a/app/assets/stylesheets/config/utilities.css.scss
+++ b/app/assets/stylesheets/config/utilities.css.scss
@@ -1,0 +1,11 @@
+.u--hide-on-mobile {
+  @media screen and (max-width: $mobile) {
+    display: none;
+  }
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/config/utilities.css.scss
```diff
@@ -0,0 +1,11 @@
+.u--hide-on-mobile {
```
AkkarinForest: I don't think this is used anywhere? But if you just added it here to have the opposite of show-on-mobile then ignore me :)

gitjul: yep, I used it before, stopped but left so it's there to use. but I don't usually do that. actually, I'll remove :P

PR #348, opened 2019-08-28
Fetch sessions when location is updated

gitjul:


gitjul:
hard to believe that that is all I have to do :D

gitjul:
> I'm testing on clone and I think it triggers the search every time I move the map?

uhh, true, I knew it couldn't be THIS simple ;)

PR #355, opened 2019-09-05
Fix info windows

AkkarinForest:
When an info window is opened near the edge of the screen the map should pan to that location. When the info window is closed the map return to the original position.
Also, moved some stuff from angular `_map.js` to vanilla `mapUtils.js`. We should try to do a bit of that with every commit.


app/javascript/javascript/clearMap.js
```diff
@@ -0,0 +1,60 @@
+export const clearMap = () => {
```
AkkarinForest: just moved it from another file, nothing new here

app/javascript/javascript/mapsUtils.js
```diff
@@ -1,67 +1,40 @@
...
+export const getHasChangedProgrammatically = () => hasChangedProgrammatically;
+
 export const lengthToPixels = (length, zoom) => {
   const pixelSize = Math.pow(2, -zoom);
   return length / pixelSize;
 };
 
 export const pixelsToLength = (pixels, zoom) => pixels * Math.pow(2, -zoom);
 
-export const clearMap = () => {
```
AkkarinForest: just moved all of this red to a separate file

app/javascript/angular/code/controllers/_sessions_list_ctrl.js
```diff
@@ -26,11 +26,8 @@ export const SessionsListCtrl = (
       // when loading the page for the first time sometimes the watch is triggered twice, first time with hasChangedProgrammatically as undefined
       if (newValue.hasChangedProgrammatically === undefined) return;
 
-      if (newValue.hasChangedProgrammatically) {
-        //triggered when deselecting a session
-        sessions.fetch({ amount: params.paramsData["fetchedSessionsCount"] });
-        return;
-      }
+      //triggered when deselecting a session or panning to info window
+      if (newValue.hasChangedProgrammatically) return;
```
gitjul: I don't have a strong preference but I guess this could be an `else` or `case` combined with line 27

app/javascript/angular/code/services/google/info_window.js
```diff
@@ -57,3 +77,5 @@ angular.module("google").factory("infoWindow", [
     return new InfoWindow();
   }
 ]);
+
+const first_popup = popup => popup.content === undefined;
```
gitjul: `firstPopup` plz :)

app/javascript/angular/code/services/google/info_window.js
```diff
@@ -22,9 +28,14 @@ angular.module("google").factory("infoWindow", [
       },
 
       show: function(url, data, position, sessionType) {
+        if (first_popup(this.popup)) savePosition();
+
         this.popup.setContent("fetching...");
         this.popup.setPosition(position);
-        this.popup.open(map.get());
+
```
gitjul: would remove this line

PR #359, opened 2019-09-11
GTM snippet after the body tag starts

Sonje:
Google Tag Assistant still have problems with reading the noscript snippet in the body (/fixed_map and /mobile_map), because it's in div and not right after the body tag starts.

### DIFF
--- a/app/javascript/elm/src/Main.elm
+++ b/app/javascript/elm/src/Main.elm
@@ -871,17 +871,9 @@ setScrollPosition value =
 viewDocument : Model -> Browser.Document Msg
 viewDocument model =
     { title = "AirCasting"
-    , body = [ view model ]
+    , body = [ snippetGoogleTagManager 
+    ,  view model ]
     }
-
-
-view : Model -> Html Msg
-view model =
-    div [ id "elm-app", class (Theme.toString model.theme) ]
-        [ snippetGoogleTagManager
-        , viewNav model.navLogo model.isNavExpanded model.sensors model.selectedSensorId model.page
-        , viewMain model
-        ]
 
 
 snippetGoogleTagManager =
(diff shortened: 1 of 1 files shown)


app/javascript/elm/src/Main.elm
```diff
@@ -897,6 +889,14 @@ snippetGoogleTagManager =
         ]
 
 
+view : Model -> Html Msg
+view model =
+    div [ id "elm-app", class (Theme.toString model.theme) ]
+        [ viewNav model.navLogo model.isNavExpanded model.sensors model.selectedSensorId model.page
+        , viewMain model
+        ]
```
gitjul: it appears to me that this change (moving the `view` function down) wasn't necessary

app/javascript/elm/src/Main.elm
```diff
@@ -871,19 +871,11 @@ setScrollPosition value =
 viewDocument : Model -> Browser.Document Msg
 viewDocument model =
     { title = "AirCasting"
-    , body = [ view model ]
+    , body = [ snippetGoogleTagManager 
+    ,  view model ]
```
gitjul: this is what likely causes the build failure, try running `yarn run elm-format --validate app/javascript/elm/` (and I think without the `--validate` to simply format the files properly)

app/javascript/elm/src/Main.elm
```diff
@@ -871,19 +871,11 @@ setScrollPosition value =
 viewDocument : Model -> Browser.Document Msg
 viewDocument model =
     { title = "AirCasting"
-    , body = [ view model ]
+    , body = [ snippetGoogleTagManager
```
gitjul: imo it's ok to do it this way cos it's clearly visible that this element has to be the first thing rendered in the `body` so 👍

PR #377, opened 2019-12-17
Update colors and fonts in the filters tab

gitjul:
- updates filters tab
- cleans up css variables
- uses new blue across the app

Screenshot 2019-12-17 at 20 20 16
Screenshot 2019-12-17 at 20 20 35


3v0k4:
Nice :)

PR #378, opened 2019-12-18
Fix sidekiq retries (most of them)

3v0k4:
See commit bodies fo more info. I've removed the test because it was not really testing much, just a chain of test doubles.


gitjul:
awesome!

PR #379, opened 2019-12-18
Sendgrid

3v0k4:
Did some cleaning while I was there.


app/services/api/google_analytics.rb
```diff
@@ -4,7 +4,7 @@
 module Api
   class GoogleAnalytics
     def register_event(event_action)
-      if A9n.host_ == 'aircasting.org'
+      if A9n.host_ == 'aircasting.habitatmap.org'
```
3v0k4: I'm going to change this to something like `A9n.google_analytics == "enabled"`. Cause, I guess it's there so that this is not run on clone.

gitjul:
👁 👍

PR #380, opened 2019-12-18
UI: update filters after HM redesign

gitjul:
Screenshot 2019-12-18 at 19 07 02


app/javascript/javascript/buildGraphOptions.js
```diff
@@ -46,14 +46,14 @@ const buildRangeSelector = ({ buttons, selected }) => ({
     states: {
       hover: {
         cursor: "pointer",
-        fill: "#09a7f0",
+        fill: "#00b2ef",
```
3v0k4: Have you considered extracting a JS file for the colors? Similarly to what we have now in CSS.

gitjul: good idea. not doing not though

PR #383, opened 2019-12-19
Update tip icons

gitjul:
Updates tooltip icons according to the design in Invision.

Screenshot 2019-12-19 at 16 07 28


3v0k4:
This is a pain point in Elm. Having to pass icons from a flag "contaminates" all the functions in between. We could try to solve it with a custom loader in Webpack. When I took a look at how to do that time ago there wasn't anything ready to use out of the box unfortunately.

PR #385, opened 2019-12-24
Average long mobile streams

3v0k4:
We decided to average long mobile sessions by putting together measurements whenever a stream contains a big amount of them. In particular
- for streams containing >4 hrs but <9 hrs of data (14,400 to 32,400 data points) we apply a 5-second avg. time interval;
- for streams containing >9 hrs of data (32,401 data points) we apply a 60-second avg. time interval.

The core of this PR is `AverageStreams` which performs what's described above. Tests should help understanding how it all works.

Besides that I did some cleaning in this PR. In particular, I've extracted some factories used in tests and started using random generators when creating test objects. Tests should separate noise from signal. It's much more readable to see

```
obj = Klass.new(a: 1)

expect(obj.call).to eq(10)
```

than

```
obj = Klass.new(a: 1, b: 123, c: 123, d: 123)

expect(obj.call(123, 123)).to eq(10)
```

In the second example all the `123` are just noise. Hiding them in a factory and making them random ensures 
- no noise in the tests;
- tests are not relying on values setup in the factories (all the signal is visible in the tests).

Where possible I've removed FactoryBot because I'd rather have to deal with the factories than hiding the complexity inside FactoryBot. In fact, in the former case I can feel the pain and adjust the design to fix the root cause. Plus the author of the gem warn against using it:

>My general advice, though, is to avoid Factory Girl as much as is reasonably possible. Not because it’s bad or unreliable software (Factory Girl is very reliable; we’ve used it successfully since 2008), but because its inherent persistence mechanism is calling #save! on the object, which will always take longer than not persisting data. (https://thoughtbot.com/blog/speed-up-tests-by-selectively-avoiding-factory-girl)

### DIFF
--- a/spec/models/stream_spec.rb
+++ b/spec/models/stream_spec.rb
@@ -1,10 +1,9 @@
 require 'rails_helper'
 
 describe Stream do
-  let(:stream) { FactoryBot.create(:stream) }
-  let!(:measurement) { FactoryBot.create(:measurement, stream: stream) }
-
   describe '#build_measurements!' do
+    let(:stream) { FactoryBot.create(:stream) }
+    let(:measurement) { FactoryBot.create(:measurement, stream: stream) }
     let(:measurement_data) { double('measurement data') }
 
     before do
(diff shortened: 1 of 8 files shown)


spec/models/stream_spec.rb
```diff
@@ -46,20 +45,14 @@
     end
   end
 
-  describe '#destroy' do
```
3v0k4: We should not test if `destroy` works. It's tested in Rails.

PR #387, opened 2020-01-08
Sidekiq: remove time loggings

3v0k4:
Now that things are stable and running well we don't need to log times anymore.


gitjul:
👀

PR #388, opened 2020-01-09
Remove Angular

3v0k4:
* remove Angular
* remove underscore.string
* upgraded jquery
* moves jquery-ui to package.json (instead of vendored files); notice that daterangepicker has some custom changes so that one remains in the repo
* removes many JS tests; the reason is that those tests have been useful to cover changes of the legacy Angular code; now that Angular is gone those tests would be only noise

Most of the internal structure of the code is just the same as it was inside Angular. Just extracted out and placed in vanilla JS files. That is aligned with our philosophy of doing small and valuable iterations for the customers. See the internal wiki for more info.

## Ideas: 

- [x]  move sessions list ctrl inside sessions map ctrl
- [x] versioner
- [x] search $scope and remove
- [x] remove sessionsUtils
- [x] remove $watch from params
- [x] remove $watch from _sessions_map_ctrl.js
- [x] map_settings.js
- [x] empty
- [x] custom_marker.js 
- [x] heat.js
- [x] sensors.js 
- [x] buildqueryparams
- [x] extract rectangles
- [x] infowindowctrl
- [x] https://github.com/esamattis/underscore.string/blob/master/dist/underscore.string.js#L1165
- [x] $on / $broadcast
- [x] extract params
- [x] __params
- [x] extract google maps
- [x] $apply
- [x] $http and spinner
- [x] __mapNG
- [x] upgrade jquery ui
- [x] review tests and noise
- [x] clean up javascript/javascript


app/assets/stylesheets/modules/filters.css.scss
```diff
@@ -68,6 +68,7 @@
   max-height: 20vh;
   max-width: $filters-width - 2 * $margin-default;
   overflow: auto;
+  z-index: 8;
```
gitjul: why 8? 
I'd prefer to use a variable that would show the rule

app/javascript/javascript/map.js
```diff
@@ -1,65 +1,50 @@
...
-  const elmApp = $window.__elmApp;
+  window.__traceMarkers = [];
+  const elmApp = window.__elmApp;
 
   var Map = function() {};
 
   Map.prototype = {
     init: function(element, options) {
       this.mapObj = googleMaps.init(element, options);
-      if (process.env.NODE_ENV !== "test") onMapInit();
```
gitjul: I'm experiencing a little bout of joy every time I see one of those

app/javascript/javascript/sessionsMap.js
```diff
@@ -0,0 +1,339 @@
...
+      : { ...defaults, isIndoor: false, isActive: true };
+
+    params.updateFromDefaults(defs);
+  }
+
+  setDefaults();
+
+  pubsub.subscribe("googleMapsChanged", function(newValue) {
+    if (params.isSessionSelected()) return;
+    // when loading the page for the first time sometimes the watch is triggered twice, first time with hasChangedProgrammatically as undefined
```
gitjul: is this still the case?

app/assets/stylesheets/modules/filters.css.scss
3v0k4: I agree :) Would you be willing to take care of that? The reason for that 8 was that without it the autocomplete dropdown is hidden behind the filters.

app/javascript/javascript/map.js
3v0k4: Hehe, as they say "you've got to crack a few eggs to make an omelette". I think that was a good trick while removing Angular!

app/javascript/javascript/sessionsMap.js
3v0k4: I don't think so but I didn't want to force my hand too much in this PR ;)

3v0k4:
@gitjul thanks a lot for taking a look!

app/javascript/javascript/sessionsMap.js
3v0k4: In any case, good catch!

app/views/layouts/map.html.haml
```diff
@@ -4,10 +4,9 @@
     %meta(name='viewport' content='width=device-width, initial-scale=1')
     %meta(name='description' content='AirCasting map showing health and environmental data like air quality, temperature, humidity, sound levels and more.')
     %title AirCasting
-    = stylesheet_pack_tag "elm", async: true
+    = stylesheet_pack_tag "elm", defer: true
```
gitjul: 👍

gitjul:
@3v0k4 sure, I was interested. the review wasn't very deep though. good stuff! there are also some `var`s left instead of const/let but didn't mark those, easy to find

PR #390, opened 2020-01-14
Increase button sizes in notes popup

gitjul:
- increases close button size in gm popups = both info window and note
- increases prev/next button sizes and adds more padding around them for better tapping

before:
Screenshot 2020-01-14 at 19 57 55

after:
Screenshot 2020-01-14 at 19 58 02


3v0k4:
👍

PR #393, opened 2020-01-15
New favicon

gitjul:


gitjul:
merging because I need to test on live website (+ won't break anything)

PR #394, opened 2020-01-20
Show a bigger, gray spinner when loading initial data

gitjul:
This PR changes the way the initial spinner looks, the one that's later replaced by the elm app.
This way it doesn't look as the spinner jumps from its position onto the map after the elm app loads but before the sessions load.

Screenshot 2020-01-20 at 16 13 48

### DIFF
--- a/app/views/maps/index.html.erb
+++ b/app/views/maps/index.html.erb
@@ -1,1 +1,1 @@
-<div class="spinner"></div>
+<div class="spinner spinner-big"></div>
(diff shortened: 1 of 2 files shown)


app/views/maps/index.html.erb
```diff
@@ -1 +1 @@
-<div class="spinner"></div>
+<div class="spinner spinner-big"></div>
```
3v0k4: 💥

gitjul: damn, thx for this! it's supposed to be `spinner--big` :D

PR #398, opened 2020-02-06
Integrate OpenAQ measurements into AirCasting

3v0k4:
* Sessions are considered active if they recorded at least a measurement in the last 4 hours (not only in the last hour); this is done to be sure that we consider most OpenAq measurements as active.
* Extracted `last_hour_average` that was duplicated in multiple places
* Added service to import OpenAq measurements (i.e. `OpenAq::ImportMeasurements`)
* Added a scheduled job to Sidekiq that runs the above every 5 minutes

More info on OpenAq and AirCasting [here](https://docs.google.com/document/d/1UVJbkcClyip9WlVy-92LAbNdSIXC0k6y3uZ_xlRTBuA/edit#heading=h.i4j2l0evr0sw).


app/javascript/javascript/singleSession.js
```diff
@@ -5,21 +5,27 @@
 // The same for the empty point at the end.
 export const measurementsToTimeWithExtremes = ({ measurements, times }) => {
   const { start, end } = times;
-  return {
+  const ms = {
```
gitjul: I don't really like this name, even though it can be easily derived what this var is from the context

app/javascript/javascript/singleSession.js
```diff
@@ -5,21 +5,27 @@
...
-    [end]: {
-      x: end,
-      y: null,
-      latitude: null,
-      longitude: null
-    }
+    ...measurementsToTime(measurements)
   };
+
+  return ms.hasOwnProperty(end)
```
gitjul: for example this in my opinion doesn't read well, just by looking at this line I don't understand what's going on

app/models/fixed_session.rb
```diff
@@ -1,13 +1,15 @@
 class FixedSession < Session
+  ACTIVE_FOR = 4.hour
```
gitjul: 👍

app/javascript/elm/src/Tooltip.elm
```diff
@@ -68,4 +68,4 @@ typeToggleFilter =
 
 
 activeToggleFilter =
-    TooltipText "Fixed sessions that have submitted a measurement in the past hour are considered active. Fixed sessions that have not submitted measurements in the past hour are considered dormant."
+    TooltipText "Fixed sessions that have submitted a measurement in the past 4 hours are considered active. Fixed sessions that have not submitted measurements in the past 4 hours are considered dormant."
```
gitjul: is this change because of the OpenAQ sensor or in general, independently from it?

app/models/stream.rb
```diff
@@ -177,4 +177,12 @@ def has_bounds?
...
       min_longitude.present?
   end
+
+  def last_hour_average
+    last_measurement_time = measurements.last.time
+    measurements.where(
+      time: last_measurement_time - 1.hour..last_measurement_time
+    )
+      .average(:value)
+  end
```
gitjul: 👍

app/services/api/to_fixed_session_hash.rb
```diff
@@ -49,15 +49,6 @@ def format_time(time)
...
   end
 
-  def last_hour_average(stream)
-    last_measurement_time = stream.measurements.last.time
-    measurements =
-      stream.measurements.where(
-        time: last_measurement_time - 1.hour..last_measurement_time
-      )
-    measurements.average(:value)
-  end
```
gitjul: 👍

app/javascript/javascript/singleSession.js
3v0k4: I'm happy to change that if you have a better name :)

app/javascript/javascript/singleSession.js
3v0k4: What would you suggest?

app/javascript/elm/src/Tooltip.elm
3v0k4: It's because before `ACTIVE_FOR` was 1 hour. We did change it to 4 hours because OpenAQ lags behind with adding measurements to their platform. That is they need a few hours to add measurements. Thus, we extended the timeframe in which we consider fixed sessions to be active to include most of them.

app/javascript/javascript/singleSession.js
gitjul: well, what does `ms` stand for here? (sorry for responding so late!)

app/javascript/javascript/singleSession.js
gitjul: it's just that if `ms` had a longer more descriptive way then this line would read better. don't know what `ms` should be instead tho without analyzing the other code around.

app/javascript/javascript/singleSession.js
3v0k4: I though Github glitched 😄 

Stands for measurements (`ms` plural of `m` where `m` stands for measurement). I guess I would have used `measurements` if that name was not already used in the arguments.

But I won't change it now 😉

app/javascript/javascript/singleSession.js
3v0k4: Makes sense! Unfortunately, I'm deep into a project and cannot take care of that. But feel free to tackle it if you feel like it!

PR #401, opened 2020-02-10
Set smaller max height of filter popup on mobile

gitjul:
This solves issues with scrolling on mobile when "more sensors" or "more parameters" are shown and the list is long.

Before:
Screenshot 2020-02-10 at 17 56 57

After:
image


3v0k4:
Looks great! 🎉

PR #403, opened 2020-02-13
Timerange: make fixed+active timerange same as mobile

3v0k4:
The diff reads really badly. This changes the timerange for the fixed tab with active enabled.

Previously it was greyed out and shown the last hour in the disabled field. Now it behaves exactly like in fixed+dormant or mobile tab. In other words, it's not disabled and initially set to the "one_year_ago..end_of_today" range. The circular arrow can be used to update the range without opening the time picker.


3v0k4:
@gitjul appreciate the discipline of going back to a PR so remote in the past. Thanks!

PR #408, opened 2020-03-11
Implement a new solution for the mobile version of filters

gitjul:
The new solution resembles the behavior of the expandable mobile menu.
This PR fixes a number of issues, including filters panel breaking whenever the screen height is less than the height of the filters panel, the need to double scroll to see the bottom part of mobile version of filters, issues with map peaking out from under filters etc.

### DIFF
--- a/app/assets/stylesheets/modules/filters.css.scss
+++ b/app/assets/stylesheets/modules/filters.css.scss
@@ -1,32 +1,57 @@
+$filter-actions-height: 60px;
+
 .filters {
   background: $gray-dark;
   color: white;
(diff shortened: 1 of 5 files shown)


3v0k4:
👁 I'm totally not the right person to judge CSS but I know this has been painful. Good job!!

gitjul:
@3v0k4 thx! have you thrown a glance at the elm changes? it's basically html but I still wanted to check if you see anything ugly there

3v0k4:
@gitjul looks great :)

PR #409, opened 2020-03-12
Fix "Profile Name" and "Tags" suggestions

gitjul:
With two alternative filter containers, profile name and tags autocomplete fields broke. Now the autocomplete is initiated per dedicated class + some css changes were introduced to adjust the dropdown menu for the mobile version and make it visible by changing the `z-index`.

This is the functionality that was broken:

image

### DIFF
--- a/app/javascript/elm/src/LabelsInput.elm
+++ b/app/javascript/elm/src/LabelsInput.elm
@@ -74,13 +74,13 @@ update msg model toCmd =
 -- VIEW
 
 
-view : Model -> String -> String -> String -> Bool -> TooltipText -> Html Msg
-view model text_ inputId placeholderText isDisabled tooltipText =
+view : Model -> String -> String -> String -> String -> Bool -> TooltipText -> Html Msg
+view model text_ inputId jsClass placeholderText isDisabled tooltipText =
     div [ class "filters__input-group" ]
         [ div [ class "tag-container" ] (List.map viewLabel <| asList model)
         , input
             [ id inputId
-            , class "input input--dark input--filters"
+            , class (String.append jsClass " input input--dark input--filters")
             , placeholder placeholderText
             , type_ "text"
             , name inputId
--- a/app/assets/stylesheets/config/variables.css.scss
+++ b/app/assets/stylesheets/config/variables.css.scss
@@ -112,9 +112,6 @@ $bold-font: "Moderat-Bold", sans-serif;
 $body-font: "Muli", sans-serif;
 
 // z-index
-$header-z-index: 999;
-
-// z-index stack for elements on top of the map
 $sessions-z-index: 4;
 $heatmap-z-index: $sessions-z-index;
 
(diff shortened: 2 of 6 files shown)


app/javascript/elm/src/LabelsInput.elm
```diff
@@ -74,13 +74,13 @@ update msg model toCmd =
...
-view : Model -> String -> String -> String -> Bool -> TooltipText -> Html Msg
-view model text_ inputId placeholderText isDisabled tooltipText =
+view : Model -> String -> String -> String -> String -> Bool -> TooltipText -> Html Msg
+view model text_ inputId jsClass placeholderText isDisabled tooltipText =
     div [ class "filters__input-group" ]
         [ div [ class "tag-container" ] (List.map viewLabel <| asList model)
         , input
             [ id inputId
-            , class "input input--dark input--filters"
+            , class (String.append jsClass " input input--dark input--filters")
```
3v0k4: You could also go with `jsClass ++ " input input--dark input--filters"`.

app/javascript/elm/src/LabelsInput.elm
```diff
@@ -74,13 +74,13 @@ update msg model toCmd =
 -- VIEW
 
 
-view : Model -> String -> String -> String -> Bool -> TooltipText -> Html Msg
-view model text_ inputId placeholderText isDisabled tooltipText =
+view : Model -> String -> String -> String -> String -> Bool -> TooltipText -> Html Msg
+view model text_ inputId jsClass placeholderText isDisabled tooltipText =
```
3v0k4: What do you think of a more specific name? Something like `inputClass` or even more specific?

3v0k4:
Nitpicked a bit :) Please consider a screenshot to show what's broken next time. I'm not totally sure I understood what was the context here. Nice and clean PR!

app/javascript/elm/src/LabelsInput.elm
3v0k4: Also, am I mistaken or `jsClass` is always `"js--profile-names-input"`?

gitjul: > What do you think of a more specific name? Something like `inputClass` or even more specific?

I wanted to underline that this is a dedicated class used by js. the fact that it's an input is not that important here. do you have a suggestion what name to use to make this clearer and more specific?

gitjul:
@3v0k4 sorry, was in a hurry, added a screenshot:)

app/javascript/elm/src/LabelsInput.elm
3v0k4: I see the value of that in the css class string (e.g. js--blah) cause it's a marker to say that the class is used in code and not for styling purposes. In here `js` is a bit more blurry cause we are already in js. Unfortunately, I'm not sure I can suggest anything.

Feel free to keep it the way it is :)

PR #415, opened 2020-05-27
fix(security): update rails 5.2.4.3

ArturT:
TODO:

Merge this after we finish testing memory grow. https://trello.com/c/64yctLqA 
I don't want to merge it now because it would restart unicorns and we would need to wait another 2 weeks to collect memory behaviour of the long period of time.


ArturT:
After upgrading Rails there is an issue with sprockets
https://github.com/rails/sprockets-rails/issues/444

PR #423, opened 2020-07-03
Use until_expired lock for OpenAQ worker to avoid deadlocks

ravicious:
I'm sparing an extended description here as I believe I explained the cause of the problem in the comment in code. Please read it first!

### `OpenAqImportMeasurementsWorker` queue priority

In the logs I seen scenarios where `OpenAqImportMeasurementsWorker` got enqueued but it took Sidekiq over 10 minutes to actually get around to starting the worker. The longest elapsed time for the worker I found was at 3.7 minutes.

Given that the lock is created when the worker gets enqueued, I think we should increase the priority for `OpenAqImportMeasurementsWorker` to ensure Sidekiq starts the worker before the lock expires. Since we should be running only one of these workers at a time and there are 15 Sidekiq threads, I think this worker could be put on the `critical` queue. Unless I'm missing something and it's really important for other workers to be fetched first.

### Guaranteeing uniqueness through other means

In short, Sidekiq (or rather Redis) can't do it without risking deadlocks that require manual intervention. sidekiq_unique_jobs should mention this in the README, just as [Sidekiq does for enterprise Unique Jobs](https://github.com/mperham/sidekiq/wiki/Ent-Unique-Jobs#use):

> This means that a second job can be pushed to Redis after 10 minutes or after the first job has successfully processed. If your job retries for a while, 10 minutes can pass, thus allowing another copy of the same job to be pushed to Redis. Design your jobs so that uniqueness is considered best effort, not a 100% guarantee. A time limit is mandatory so that if a process crashes, any locks it is holding won't last forever.

More in-depth explanation is available in [Redis docs](https://redis.io/topics/distlock#why-failover-based-implementations-are-not-enough).

As it was suggested in the comment in `OpenAqImportMeasurementsWorker` before, the most reliable solution to guarantee this would be to implement this on the database level. However, I'm not particurarly knowledgable when it comes to Mysql so I figured out it'll take me less time to create this lock workaround rather than refactoring database stuff.

However I'm open to suggestions as I'm not particularly fond of the solution I implemented. It'd be much more straightforward if sidekiq_unique_jobs supported the same locking mechanism that Sidekiq Enterprise offers.

### DIFF
--- a/app/workers/open_aq_import_measurements_worker.rb
+++ b/app/workers/open_aq_import_measurements_worker.rb
@@ -18,13 +18,94 @@ class OpenAqImportMeasurementsWorker
   # in the database will not be processed anymore from that point on (because they would always
   # raise). But duplicates should not happen because OpenAqImportMeasurementsWorker runs serially.
   #
-  # We could have solved the concurrent inserts with a different architecture (e.g. unique
-  # indexes in the database) but this seemed to be a simple and good enough solution.
-  sidekiq_options lock: :until_executed, on_conflict: :log
+  # Unfortunately, we can't use the `:until_executed` lock from sidekiq_unique_jobs because when the
+  # Sidekiq server is unexpectedly shut down while this worker is working, the lock will stay in
+  # Redis until someone manually removes it.
+  #
+  # Typically this problem is solved by making sure the lock stays either until the worker finishes
+  # or until a certain amount of time passes. [1]
+  #
+  # Unfortunately again, sidekiq_unique_jobs, unlike the enterprise version of Sidekiq, doesn't
+  # support such locking strategy. Instead, we have to hack it ourselves by using the
+  # `:until_expired` lock and Sidekiq's API to remove the lock when the worker finishes.
+  #
+  # `:until_expired` ensures that the lock expires after the time specified through
+  # `lock_expiration`. In case Sidekiq is shut down unexpectedly, the lock will eventually expire
+  # and Sidekiq will be able to enqueue this worker again.
+  #
+  # As [1] explains, this method doesn't guarantee uniqueness, but the alternative is having a
+  # system that requires a manual intervention after an unexpected restart/shutdown. [2] goes in
+  # detail as to why the timeout is required.
+  #
+  # Another unfortunate side effect of using `:until_expired` is that locks created by it aren't
+  # visible through the "Unique Digests" panel in Sidekiq dashboard as well as through
+  # `SidekiqUniqueJobs::Digests.all`. But the logs should still show info about skipping the worker
+  # because the unique digest being present. Attempting to enqueue the worker from console will
+  # return the error as well.
+  #
+  # One way to guarantee uniqueness for the concurrent inserts would be to use a different
+  # architecture (e.g. unique indexes in the database).
+  #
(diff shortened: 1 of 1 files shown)


app/workers/open_aq_import_measurements_worker.rb
```diff
@@ -18,13 +18,94 @@ class OpenAqImportMeasurementsWorker
...
+  # return the error as well.
+  #
+  # One way to guarantee uniqueness for the concurrent inserts would be to use a different
+  # architecture (e.g. unique indexes in the database).
+  #
+  # [1] https://github.com/mperham/sidekiq/wiki/Ent-Unique-Jobs#use
+  # [2] https://redis.io/topics/distlock#why-failover-based-implementations-are-not-enough
+  sidekiq_options lock: :until_expired,
+                  on_conflict: :log,
+                  lock_expiration: 30.minutes
```
ravicious, line 53: I'm going to confirm it with Michael that 30 minutes is fine. This is going to affect the potential for duplicates to occur as well as how long it will take the system to recover after an unexpected shutdown.

app/workers/open_aq_import_measurements_worker.rb
```diff
@@ -18,13 +18,94 @@ class OpenAqImportMeasurementsWorker
...
+                  } was executed synchronously. Skipping lock removal."
+      return
+    end
+
+    if Time.now - started_at < 5.seconds
+      # Sidekiq::Workers data is updated every 5 seconds.
+      # https://github.com/mperham/sidekiq/wiki/API#workers
+      logger.info 'Sleeping for 5 seconds to wait for Sidekiq workers data to refresh.'
+      sleep 5
+    end
```
ravicious, line 79: When I found out that this is the case I was like, okay, this is getting quite complex already. 🤕

But really there's no other way to approach it if we want to remove this lock manually from within the `perform` method. Within `perform`, we don't have any meta info about the payload of the worker other than `jid`. If we had `unique_digest` then I wouldn't need to do all this waiting.

ravicious:
I also considered using [sidekiq_unique_digests_cleaner](https://github.com/restaurant-cheetah/sidekiq_unique_digests_cleaner) which I found in an issue created in sidekiq_unique_jobs. It can clear unused locks at the start of Sidekiq, but I found it to be very unreliable. It requires a fix to not crash under certain conditions, then there are problems with the way it handles orphaned workers (it doesn't really). Overall I just found it to be much less reliable than the solution that I implemented here.

ravicious:
After a call with Riccardo I have to close this. I was under a wrong assumption that if once in a blue moon two workers run at the same time, nothing too bad will happen, we'll just have some duplicates in the system.

However, this is not true at all. The worker was made so that if it detects any duplicates in streams or sessions, it'll stop processing any new measurements.

https://github.com/HabitatMap/AirCasting/blob/f1ce46b7231fce8961afb4ce2db576b0355ff622/app/services/open_aq/save_measurements.rb#L45-L55

So the system depends on the fact that only one instance of this worker can be run at a time. Removing this dependency might require making extensive changes in the database structure, for which we don't have resources at the moment.

I'll outline another solution on Trello and give some pointers as to what can be done in the future.

PR #431, opened 2021-02-18
API call for uploading fixed measurements in an asynchronous way

lesniakania:
This is needed for the mobile app SD card sync.

[Trello task](https://trello.com/c/35QfLbSk/1543-api-call-for-uploading-fixed-measurements-in-an-asynchronous-way)

It's almost done, but tests are missing yet.


3v0k4:
@lesniakania can we close this PR? 🙂

lesniakania:
Yeah, it's not needed anymore.

lesniakania:
Actually, not sure. It's not necessary at this moment, but it might be at some point - for now, the Aircasting Mobile app is using the old endpoint to send AB3 measurements during sync, but it was meant only for MVP, it was supposed to be changed after. So there is a separate endpoint for that and the backend saves only ones that it does not have already.
I will see Marysia tomorrow, so I can confirm.

PR #486, opened 2022-09-16
Fix threshold alert emails

gitjul:
Note: this solution fixes the issue for outdoor sessions, but doesn't work for indoor sessions. It's deployed to production, but a better solution is in progress, which requires some work in iOS: https://github.com/HabitatMap/AirCasting/pull/489


gitjul:
https://github.com/HabitatMap/AirCasting/pull/489

PR #527, opened 2023-10-10
Sessions cleanup

mpolec:
# Sessions table clean up
The aim of this PR is to remove unused columns and duplicated data from sessions table.

## Session start and end time
We store information about session start and end time. These attributes are used for filtering when a user searches for sessions.

The problem is that we have some duplication - there are four columns in the db - `start_time`, `end_time`, `start_time_local`, `end_time_local`. Looking at data, `start_time` and `start_time_local` and `end_time` and end_time_local respectively store the same data (except 4 records for data after 2019-01-01, 287 records in total, checked in Oct '23).

I think we can rely on information stored in `start_time_local`, `end_time_local`; `start_time` and `end_time` can be removed.

## Other unused fields
Additionally, there are three unused fields - `data_type`, `instrument`, `measurements_count` which can be removed.
As a reference: these fields are set for records from 2011-2013.


BartkiewiczM (approved):
works 🚀 ✅

PR #529, opened 2023-10-18
Migration to PostgreSQL

BartkiewiczM:
App adapted to work with Postgres. 
MySQL-specific queries adapted.
Case sensitiveness cases resolved as Postgres is and mysql is not case-sensitive in strings.
Tests adapted.


db/migrate/20231012105017_change_sessions_columns_to_boolean.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeSessionsColumnsToBoolean < ActiveRecord::Migration[6.0]
+  def up
+    change_column :sessions, :contribute, 'boolean USING CASE WHEN contribute = 1 THEN TRUE ELSE FALSE END'
+  end
+
+  def down
+    change_column :sessions, :contribute, :integer, limit: 2, using: 'CASE WHEN contribute THEN 1 ELSE 0 END'
```
mpolec, line 7: It seems that currently we use `tinyint(1)` for the `contribute` column.
```
mysql> desc sessions;
+---------------------+---------------+------+-----+---------+----------------+
| Field               | Type          | Null | Key | Default | Extra          |
+---------------------+---------------+------+-----+---------+----------------+
| id                  | int(11)       | NO   | PRI | NULL    | auto_increment |
| created_at          | datetime      | YES  |     | NULL    |                |
| updated_at          | datetime      | YES  |     | NULL    |                |
| user_id             | int(11)       | YES  | MUL | NULL    |                |
| uuid                | varchar(255)  | YES  | MUL | NULL    |                |
| url_token           | varchar(255)  | YES  | MUL | NULL    |                |
| title               | text          | YES  |     | NULL    |                |
| contribute          | tinyint(1)    | YES  | MUL | NULL    |                |
| data_type           | varchar(255)  | YES  |     | NULL    |                |
| instrument          | varchar(255)  | YES  |     | NULL    |                |
| start_time          | datetime      | YES  | MUL | NULL    |                |
| end_time            | datetime      | YES  | MUL | NULL    |                |
| measurements_count  | int(11)       | YES  |     | NULL    |                |
| start_time_local    | datetime      | YES  | MUL | NULL    |                |
| end_time_local      | datetime      | YES  | MUL | NULL    |                |
| type                | varchar(255)  | NO   |     | NULL    |                |
| is_indoor           | tinyint(1)    | YES  |     | NULL    |                |
| latitude            | decimal(12,9) | YES  |     | NULL    |                |
| longitude           | decimal(12,9) | YES  |     | NULL    |                |
| last_measurement_at | datetime      | YES  | MUL | NULL    |                |
| version             | int(11)       | YES  |     | 1       |                |
+---------------------+---------------+------+-----+---------+----------------+
21 rows in set (0.00 sec)
```

db/migrate/20231011105418_change_is_indoor_to_boolean_in_sessions.rb
mpolec, line 1: Why is there a different approach to changing `is_indoor` and a different to changing `contribute`?

db/schema.rb
```diff
@@ -10,46 +10,43 @@
...
+
+  create_table "active_storage_attachments", force: :cascade do |t|
+    t.string "name", limit: 255, null: false
+    t.string "record_type", limit: 255, null: false
     t.bigint "record_id", null: false
     t.bigint "blob_id", null: false
     t.datetime "created_at", null: false
-    t.index ["blob_id"], name: "index_active_storage_attachments_on_blob_id"
-    t.index ["record_type", "record_id", "name", "blob_id"], name: "index_active_storage_attachments_uniqueness", unique: true
+    t.index ["blob_id"], name: "public_active_storage_attachments_blob_id1_idx"
```
mpolec, line 24: I wonder why index names got changed like that 🤔

mpolec (commented):
Good job 🙌  I'm glad to see that you've already solved a lot of issues.

Please take a look at comments I left. I haven't checked everything yet, but I'm submitting what I have so far, so you can already address it.

app/services/fixed_region_info.rb
mpolec, line 1: I think we need to refactor logic for fetching statistics.
Not sure about intention behind `end_time` method - is it a way to define last hour without reference to time zones? Maybe it can be done simpler without a separate db call for each stream?

spec/controllers/api/fixed/active/sessions_controller_spec.rb
```diff
@@ -248,8 +248,8 @@
             'start_time_local' => '2000-10-01T02:03:04.000Z',
             'last_measurement_value' => queried_stream.average_value,
             'is_indoor' => session.is_indoor,
-            'latitude' => session.latitude,
-            'longitude' => session.longitude,
+            'latitude' => session.latitude.to_f,
+            'longitude' => session.longitude.to_f,
```
mpolec, line 252: Is format used in our api response changed somehow?

mpolec (commented):
Adding some comments, let's talk tomorrow live about some more complicated cases.

PR #532, opened 2023-11-16
Typography

smialko:
I added some typography styles based on Figma designs for future use.

### DIFF
--- a/app/javascript/react/assets/styles/Typography/Typography.style.tsx
+++ b/app/javascript/react/assets/styles/Typography/Typography.style.tsx
@@ -1,0 +1,42 @@
+import styled, { css } from "styled-components";
+
+const H1 = styled.h1`
+  font-family: "Roboto";
+  font-weight: 600;
+  font-size: 28px;
+`;
+
+const H2 = styled.h2`
+  font-family: "Roboto";
+  font-weight: 600;
+  font-size: 24px;
+  text-transform: capitalize;
+`;
+
+const H3 = styled.h3<{ bold?: boolean }>`
+  font-weight: ${(p) => (p.bold ? "600" : "400")};
+  font-size: 18px;
+`;
+
+const H4 = styled.h4<{ bold?: boolean }>`
+  font-family: "Roboto";
+  font-weight: ${(p) => (p.bold ? "600" : "400")};
+  font-size: 14px;
+  letter-spacing: 0.14px;
+`;
+
+const H5 = styled.h5<{ bold?: boolean }>`
+  font-family: "Roboto";
+  font-weight: ${(p) => (p.bold ? "600" : "400")};
+  font-size: 12px;
+`;
+
+const H6 = styled.h6<{ bold?: boolean; uppercase?: boolean }>`
+  font-family: "Roboto";
+  font-weight: ${(p) => (p.bold ? "500" : "400")};
+  font-size: 10px;
+  text-transform: ${(p) => (p.uppercase ? "uppercase" : "none")};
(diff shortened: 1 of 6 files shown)


app/javascript/react/assets/styles/Typography/Typography.style.tsx
```diff
@@ -0,0 +1,107 @@
+import styled, { css } from "styled-components";
+
+const H1 = styled.h1`
+  font-family: "Roboto";
```
franiakat, line 4: I think there is no other font used in the calendar feature, why don't we just set it once in `global-styles.ts`?

app/javascript/react/assets/styles/Typography/Typography.style.tsx
```diff
@@ -0,0 +1,107 @@
...
+  font-weight: 600;
+  font-size: 28px;
+  font-style: normal;
+  line-height: 1.2;
+  color: #3e4449;
```
franiakat, line 9: Please add this color to variables, also we can set the font color globally, as most often that would be true, can be then removed from all the elements.

app/javascript/react/assets/styles/Typography/Typography.style.tsx
```diff
@@ -0,0 +1,107 @@
+import styled, { css } from "styled-components";
+
+const H1 = styled.h1`
+  font-family: "Roboto";
+  font-weight: 600;
+  font-size: 28px;
+  font-style: normal;
```
franiakat, line 7: It's a default, don't have to specify that, please remove.

app/javascript/react/assets/styles/Typography/Typography.style.tsx
```diff
@@ -0,0 +1,107 @@
...
+  line-height: 1.2;
```
franiakat, line 8: Could be globally set as well I think?

app/javascript/react/assets/styles/Typography/Typography.style.tsx
```diff
@@ -0,0 +1,107 @@
...
+`;
+
+const H3bold = styled.h3`
+  font-family: "Roboto";
+  font-weight: 600;
+  font-size: 18px;
+  font-style: normal;
+  line-height: 1.2;
+  color: #3e4449;
+`;
```
franiakat, line 38: Here we can have just one component with a prop:
```suggestion
const H3 = styled.h3<{bold?: boolean}>`
  font-weight: ${(p) => p.bold ? '600' : '400'};
  font-size: 18px;
`;
```

I've written it here, but most probably should work.

app/javascript/react/assets/styles/Typography/Typography.style.tsx
```diff
@@ -0,0 +1,107 @@
...
+const H6uppercase = styled.h6`
+  font-family: "Roboto";
+  font-weight: 400;
+  font-size: 10px;
+  font-style: normal;
+  line-height: 1.2;
+  letter-spacing: 0.1px;
+  text-transform: uppercase;
+  color: #3e4449;
+`;
```
franiakat, line 105: We can treat this uppercase feature as a prop as well, similar to bold.

franiakat (changes requested):
⭐ Happy to be doing reviews again!

Please see the suggestions. :)

jankrzempek (commented):
It was a great opportunity to learn, thanks. Nothing more to add 😄

franiakat (approved):
Good to go! 🎖️ 
Please use the option of squash and merge.
sqash

PR #534, opened 2023-11-22
Header buttons

smialko:
I added a style for those buttons: 
Screenshot 2023-11-22 at 09 50 05

There's a problem with a font but it'll be fixed in separate PR.


app/javascript/react/assets/styles/Typography/Typography.style.tsx
```diff
@@ -39,4 +39,12 @@ const H6 = styled.h6<{ bold?: boolean; uppercase?: boolean }>`
   letterspacing: ${(p) => (p.uppercase ? "0.1px" : "normal")};
 `;
 
-export { H1, H2, H3, H4, H5, H6 };
+const buttonText = css`
```
jankrzempek, line 42: Is it the same as `H4` but with `text-transform: uppercase;`? If so, maybe we can add this as parameter to existing H4?

app/javascript/react/components/Button/Button.style.tsx
```diff
@@ -0,0 +1,25 @@
...
+const buttonGeneralStyles = css`
+  ${buttonText}
+  height: 42px;
+  border-radius: 5px;
+  padding: 16px;
+  align-items: center;
+  gap: 10px;
+`;
+
+const Button = styled.button`
```
jankrzempek, line 14: What is the point of this? Is this some kind of a middle thing that does nothing more? If so maybe we can shorten this and change HeaderButton

```
const HeaderButton = styled.button`
  ${buttonGeneralStyles}
  justify-content: flex-end;
  display: flex;
  border: 1px solid ${lightGray};
  background-color: transparent;
`;
```

jankrzempek (commented):
Works well, and looks as designed. Two concerns in comments 😄

app/javascript/react/components/Button/Button.style.tsx
smialko: Hmmm, tbh I had some errors and I was trying different things. Also I had some code inside the `Button` and then I moved it but left the `Button` to avoid the previous errors.  But now I see it's working fine when I remove this `Button` 🤡 so I guess I'll just remove it

app/javascript/react/assets/styles/Typography/Typography.style.tsx
smialko: I'm not sure, I think it's nicer to have something dedicated to buttons and not use "headers" for button text. But that's just my intuition, let's see what Kasia will say is more common or sth :D

franiakat: Good catch :) But the duplication hurts me less, than thinking about a good name for those extracted files we could use both in h4 tag and the button :D If you have a good idea, go for it.

I think I wouldn't keep those styles in Typgraphy file, I'd remove this declaration from here and keep the styles in the button itself. : )

app/javascript/react/assets/styles/Typography/Typography.style.tsx
```diff
@@ -39,4 +39,12 @@ const H6 = styled.h6<{ bold?: boolean; uppercase?: boolean }>`
...
+  font-family: "Roboto";
```
franiakat, line 43: I assume, would be inherited from body tag, please rebase with main after merging typography :)

app/javascript/react/components/Button/Button.tsx
```diff
@@ -0,0 +1,13 @@
+import * as S from "./Button.style";
+import React from "react";
```
franiakat, line 2: We can agree that we always first import from libraries, leave an empty line, custom components and utils, empty line and styles at the end. :) It's just a matter of following convention. :)

app/javascript/react/components/Button/Button.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+interface ButtonProps {
+  onClick: () => void;
+  children: string;
```
franiakat, line 6: Can be a text node or other element like an icon, etc. This should work:
```suggestion
  children: string | JSX.Element | (JSX.Element | string)[];
```

franiakat (changes requested):
Please see the suggestions :)

app/javascript/react/components/Button/Button.style.tsx
```diff
@@ -0,0 +1,28 @@
...
+  gap: 10px;
+`;
+
+const HeaderButton = styled.button`
+  ${buttonGeneralStyles}
+  justify-content: flex-end;
+  display: flex;
+  border: 1px solid ${lightGray};
+  background-color: transparent;
+`;
```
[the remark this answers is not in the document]
smialko: I saw `generalStyles` in SquareShare & I saw that there are other buttons that have similar UI characteristics so I thought that that's what 'you' do :D

app/javascript/react/components/Button/Button.tsx
smialko: sure
@jankrzempek I think that this comment might interest you, too :D

app/javascript/react/components/Button/Button.style.tsx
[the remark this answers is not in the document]
franiakat: Oh, I see :D There was a different problem with styles not applying in buttons elsewhere, that's why it was separate. :D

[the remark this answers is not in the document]
franiakat: So if it wasn't clear - please, merge it all together, unless the `buttonText` is needed elsewhere. :)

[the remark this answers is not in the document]
smialko: Done

franiakat:
Please ping me after rebase so I can do the final review. :)

app/javascript/react/components/Button/Button.style.tsx
```diff
@@ -0,0 +1,20 @@
+import styled from "styled-components";
+import { lightGray } from "../../assets/styles/colors";
+
+const HeaderButton = styled.button`
```
franiakat, line 4: For now this is our only button, we know there are more types coming, but we yet need to think about its looks and how to code all the versions. Therefore I would keep the naming consistent and use `Button` here.

app/javascript/react/components/Button/Button.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+import { HeaderButton } from "./Button.style";
```
franiakat, line 3: Let's agree to always follow:
```suggestion
import * as S from "./Button.style";
```

Then:
<S.Button>...

franiakat (approved):
Just a minor naming chnages. :) Good to go!

PR #543, opened 2023-12-12
Threshold config

smialko:


app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.style.tsx
```diff
@@ -0,0 +1,138 @@
...
+  &::-webkit-outer-spin-button,
+  &::-webkit-inner-spin-button {
+    -webkit-appearance: none;
+    margin: 0;
+  }
+
+  input[type="number"] {
+    appearance: textfield;
+    -moz-appearance: textfield;
+  }
```
jankrzempek, line 124: don't think it has a purpose, at least not yet

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.style.tsx
```diff
@@ -0,0 +1,138 @@
...
+  cursor: move; /* fallback if grab cursor is unsupported */
+  cursor: grab;
+  cursor: -moz-grab;
+  cursor: -webkit-grab;
+
+  &:active {
+    cursor: grabbing;
+    cursor: -moz-grabbing;
+    cursor: -webkit-grabbing;
+  }
```
jankrzempek, line 135: don't think it has a purpose, at least not yet

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const [thresholds, setThresholds] = useState(initialThresholds);
+  const [sliderWidth, setSliderWidth] = useState(0);
+  const [thumbPositions, setThumbPositions] = useState<number[]>([]);
+  const sliderRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (sliderRef.current) {
+      setSliderWidth(sliderRef.current.offsetWidth);
+    }
+    const [min, first, second, third, max] = thresholds;
```
jankrzempek, line 40: It's subjective but I like more `[min, low, middle, hight, max]` becouse now first is not exactly first

jankrzempek (commented):
Impressive, so complex, and remarkable code!

When I say `don't think it has a purpose, at least not yet` - I mean that commenting this out in my end changes nothing

jankrzempek (commented):
Also those:

https://github.com/HabitatMap/AirCasting/assets/44276070/67d93651-0ece-4d57-91cd-6732e9859d38


https://github.com/HabitatMap/AirCasting/assets/44276070/53bbbbc4-8067-42f8-88a5-a4f5d9b29870

package.json
```diff
@@ -6,7 +6,7 @@
     "test": "NODE_ENV=test blue-tape -r @babel/register app/javascript/javascript/tests/**/*.test.js",
     "coverage": "NODE_ENV=test nyc --require @babel/register blue-tape app/javascript/javascript/tests/**/*.test.js",
     "coverage:lcov": "NODE_ENV=test nyc --require @babel/register --reporter=lcov blue-tape app/javascript/javascript/tests/**/*.test.js",
-    "build": "webpack --config ./config/webpack/webpack.config.js --mode production --progress",
```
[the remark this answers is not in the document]
franiakat: It was, but I'd rather have it done in a separate PR.

package.json
```diff
@@ -57,7 +57,7 @@
     "@babel/register": "^7.16.0",
     "@istanbuljs/nyc-config-babel": "^3.0.0",
     "@prettier/plugin-ruby": "^2.0.0",
-    "@types/react": "^18.2.23",
+    "@types/react": "^18.2.46",
```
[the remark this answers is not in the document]
franiakat: No, probably "a fruit of multiple rebases". 😅

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
franiakat: Definitely!

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.style.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+const SliderContainer = styled.div`
+  position: relative;
+  width: 90%;
+  height: 70px;
+`;
+
+const RangeInput = styled.input.attrs<{ $thresholds: number[] }>((props) => ({
+  min: props.$thresholds[0],
+  max: props.$thresholds[4],
```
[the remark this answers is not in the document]
franiakat: I agree, changing!

franiakat:
Those two cases from the videos are known - to be taken care of separately.

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.style.tsx
```diff
@@ -0,0 +1,112 @@
...
+  $sliderWidth: number;
+}>`
+  width: 100%;
+  position: absolute;
+  top: 0;
+  height: 8px;
+  background: transparent;
+  -webkit-appearance: none;
+  appearance: none;
+  background: transparent;
```
jankrzempek, line 28: repetition here, it is presented above already

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
```diff
@@ -0,0 +1,153 @@
...
+  min: number,
+  max: number,
+  width: number
+): number => {
+  const percentage = calculateThumbPercentage(value, min, max);
+  return percentage * width;
+};
+
+interface ThresholdsConfiguratorProps {
+  initialThresholds: number[];
```
jankrzempek, line 25: I am not a fan of this for, few reasons:
1. We are not safe using arrays, because we are never sure about the number of elements presented inside an array
2. When accessing the threshold in the code below, we **assumed** it should be index `0` as min and `4` as max
3. We are accessing those thresholds in code multiple times, which makes us more error-prone. Human factors here can badly influence it, and by mistake someday we can access it as `min` ex. `1` instead of `0`
4. To be safe, as long as we are sure the number of thresholds is finite we should apply another interface and then use `thresholds.minimum`

This is the approach I am trying to apply when using type-safe languages, please skip this if you know it's not the approach we should take 🤝

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
```diff
@@ -0,0 +1,153 @@
...
+}
+
+const ThresholdsConfigurator: React.FC<ThresholdsConfiguratorProps> = ({
+  initialThresholds,
+}) => {
+  const [thresholds, setThresholds] = useState(initialThresholds);
+  const [sliderWidth, setSliderWidth] = useState(0);
+  const [thumbPositions, setThumbPositions] = useState<number[]>([]);
```
jankrzempek, line 33: Same as for `thresholds`, I believe we can make it and use using interface and making less error-prone

jankrzempek (commented):
Thank you for applying previous suggestions. Few more to consider 🫡

jankrzempek (commented):
One more thing is `mobile` it will be done in different PR I should assume? Just to clear things out

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
smialko: +1 to that.
I think we used enum in iOS, didn't we?

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
```diff
@@ -0,0 +1,153 @@
...
+      <S.NumberInput
+        type="number"
+        value={thresholds[0]}
+        onChange={(e) => handleInputChange(0, e.target.value)}
+      />
+      {thresholds.slice(1, -1).map((value, index) => (
+        <React.Fragment key={index}>
+          <S.RangeInput
+            $min={thresholds[0]}
+            $max={thresholds[4]}
```
smialko, line 109: Hmm, I think that in iOS we would avoid doing that, we were using a struct just for thresholds. I'm not saying it's the best approach but I remember discussion that keeping them as an array and getting values like that is error prone.  
But tbh I don't know what's the best approach in frontend, feel free to ignore :D

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
```diff
@@ -0,0 +1,153 @@
+import React, { useEffect, useRef, useState } from "react";
+
+import * as S from "./ThresholdConfigurator.style";
+
+const calculateThumbPercentage = (
+  value: number,
+  min: number,
+  max: number
+): number => {
+  const percentage = (value - min) / (max - min);
```
smialko, line 10: Did you consider the case where min is negative number? F.e. temperature can be negative and I believe we measure temp.

smialko (commented):
🤘

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.style.tsx
franiakat: Removed!

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
franiakat: Yes, I agree, thanks for the thorough comment! :)
I wasn't really thinking about data structures intentionally as waiting for the API shape, although given how much code I produced it could've been done. 😅

franiakat:
> One more thing is `mobile` it will be done in different PR I should assume? Just to clear things out

Yes, as written in description - what I meant in RWD. ;)

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
franiakat: Wait, we agreed with Marta, that API won't send the values most probably. 😅

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
franiakat: Yes, it should work the same. :) I was adding test for it, but struggled with jest setup. I will add jest setup as a task on our board.

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
```diff
@@ -0,0 +1,145 @@
...
+        calculateThumbPercentage(
+          startValue,
+          thresholdValues.min,
+          thresholdValues.max
+        ) +
+        dx / sliderWidth;
+      const newValue = Math.round(
+        thresholdValues.min +
+          newPercentage * (thresholdValues.max - thresholdValues.min)
+      );
```
jankrzempek, line 56: Maybe we could add some comment to this math, as it is not so clear at first glance. Or the option second, extract something to variable and divide it into smaller? 
Subjective

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+      setThresholdValues((prevThresholds) => {
+        return {
+          ...prevThresholds,
+          [thresholdKey]: Math.min(
+            Math.max(newValue, thresholdValues.min),
+            thresholdValues.max
```
jankrzempek, line 63: Maybe we could add some comment to this math, as it is not so clear at first glance. Or the option second, extract something to variable and divide it into smaller? 
Subjective

jankrzempek (approved):
🥳🥳🥳 Perfection!

franiakat:
I have tested on experimental and staging, working there, merging.

PR #545, opened 2023-12-19
Redux introduction

jankrzempek:
**What was implemented:** Redux toolkit based on the [https://redux-toolkit.js.org/tutorials/typescript](url) and square share.
**GOAL:** Store different app states using this tool

Main file that stores all the reducers is `store/index.ts`.
`thresholdSlice.ts` is the initial version for future slider and its values

--------------------------------------------------------------------------
Example of implementation:

Import:
```
import { useAppSelector, useAppDispatch } from '../../store/hooks'
import { updateGivenIndex } from '../../store/thresholdSlice'
```

Use:
```
 const threshold = useAppSelector((state) => (state.threshold));
 const dispatch = useAppDispatch()
```

Execute:
```
console.log("DID TAP ME")
dispatch(updateGivenIndex({id: 2, value: 2}))
```


app/javascript/react/App.tsx
```diff
@@ -1,14 +1,18 @@
...
 import GlobalStyles from "./assets/styles/global-styles";
+import store from "./store/index";
+import './locales/i18n'
 
 const App = () => {
   return (
     <>
-      <GlobalStyles />
-      <CalendarPage />
+      <Provider store={store}>
```
franiakat, line 12: An empty tag is not needed anymore, the rule of returning a single child element is et thanks to adding the Provider component.

app/javascript/react/store/hooks.ts
```diff
@@ -0,0 +1,6 @@
+import { useDispatch, useSelector } from 'react-redux'
+import type { TypedUseSelectorHook } from 'react-redux'
+import type { RootState, AppDispatch } from './index'
+
+export const useAppDispatch: () => AppDispatch = useDispatch
+export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector
```
franiakat, line 6: We need to agree on prettier styling standards :D
```suggestion
import { useDispatch, useSelector } from "react-redux";
import type { TypedUseSelectorHook } from "react-redux";

import type { RootState, AppDispatch } from "./index";

export const useAppDispatch: () => AppDispatch = useDispatch;
export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector;
```

app/javascript/react/store/index.ts
```diff
@@ -0,0 +1,12 @@
+import { configureStore } from "@reduxjs/toolkit";
+import thresholdReducer from "./thresholdSlice";
```
franiakat, line 2: Please separate the imports: external libs and internal module.

app/javascript/react/store/thresholdSlice.ts
```diff
@@ -0,0 +1,52 @@
+import { createSlice } from "@reduxjs/toolkit";
+import type { PayloadAction } from "@reduxjs/toolkit";
+import type { RootState } from "./index";
```
franiakat, line 3: Imports rules. ;)

app/javascript/react/store/thresholdSlice.ts
```diff
@@ -0,0 +1,52 @@
...
+  name: "threshold",
+  initialState,
+  reducers: {
+    updateAll: (state, action: PayloadAction<ThresholdState>) => {
+      state.th1 = action.payload.th1;
+      state.th2 = action.payload.th2;
+      state.th3 = action.payload.th3;
+      state.th4 = action.payload.th4;
+      state.th5 = action.payload.th5;
+    },
```
franiakat, line 36: We can de-structure action payload object earlier and refer to particular `th` values when updating the state.
```suggestion
    updateAll: (
      state,
      { payload: { th1, th2, th3, th4, th5 } }: PayloadAction<ThresholdState>
    ) => {
      state.th1 = th1;
      state.th2 = th2;
      state.th3 = th3;
      state.th4 = th4;
      state.th5 = th5;
    },
```

app/javascript/react/store/thresholdSlice.ts
```diff
@@ -0,0 +1,52 @@
...
+    updateGivenIndex: (state, action: PayloadAction<Threshold>) => {
+      const { id, value } = action.payload;
+      const key = `th${id}` as keyof ThresholdState;
+      console.log(key)
+      if (key in state) {
+        state[key] = value;
+      } else {
+        console.log("No update possible, the key is out of the range.");
+      }
+    },
```
franiakat, line 46: I think I would change the Threshold interface as follows:
```
interface Threshold {
  name: keyof ThresholdState;
  value: number;
}
```

And then we could have this action defined:
```
    updateGivenIndex: (
      state,
      { payload: { name, value } }: PayloadAction<Threshold>
    ) => {
      state[name] = value;
    },
```

franiakat (changes requested):
Thanks a lot for a great description, I have suggested a few changes. :)

I think we may end up with thresholds being a different data structure, but let's see once we get into working on API and data structure more.

app/javascript/react/store/thresholdSlice.ts
jankrzempek: Nice, it did not came through my mind, thanks!

franiakat (approved):
Good job, Janek, thanks for applying suggestions, ready to go! 🤸‍♂️

PR #556, opened 2024-01-05
Display calendar page when feature flag enabled

franiakat:
Read `calendar` feature flag value from Flipper and display either Calendar Page or 404.


mpolec (approved):
👍

PR #576, opened 2024-02-27
[Time handling] time to time_with_timezone translation script

BartkiewiczM:
I've created a column that will store measurement time in UTC.
I've translated the time from local using sessions latitude/longitude with timezone_finder gem.


mpolec (commented):
I added some comments, please take a look.

lib/session_timezone_builder.rb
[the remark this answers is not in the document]
BartkiewiczM: Honestly I've searched for `builders` in the project and found them only in `/lib`. Should it be maybe in a new folder `services/builders` or is there any other idea?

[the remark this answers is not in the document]
BartkiewiczM: SessionTimezoneBuilder -> TimeZoneBuilder in services

lib/session_builder.rb
```diff
@@ -1,3 +1,5 @@
+require './lib/session_timezone_builder'
```
mpolec, line 1: is that needed?

lib/tasks/populate_session_timezones.rake
```diff
@@ -0,0 +1,16 @@
...
+
+namespace :sessions_timezone do
+  task populate: :environment do
+    puts "Starting to populate sessions timezones"
+
+    finder = TimezoneFinder.create
+
+    Session.find_each(batch_size: 1000) do |session|
+      timezone = TimeZoneBuilder.new.call(session.latitude, session.longitude)
+      session.update(timezone: timezone)
```
mpolec, line 11: `session.update(time_zone: time_zone)`?

spec/services/time_zone_builder_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+describe TimeZoneBuilder do
```
mpolec, line 3: I would add here: `subject { described_class.new }`
and then instead of `TimeZoneBuilder.new.call(50.0640, 19.9440)` we can have `subject.call(50.0640, 19.9440)`.

lib/session_builder.rb
```diff
@@ -19,6 +21,7 @@ def build!
     stream_data = data.delete(:streams)
 
     data = build_local_start_and_end_time(data)
+    data[:timezone] = TimeZoneBuilder.new.call(data[:latitude], data[:longitude])
```
mpolec, line 24: I would keep the naming consistent - `data[:time_zone]`.

PR #577, opened 2024-03-04
api/fixed/sessions/:id unused endpoint deletion

BartkiewiczM:
We've discovered a lot of requests to this public endpoint connected to deleted OpenAQ sessions.
We don't use this endpoint anymore in our mobile apps.


mpolec (changes requested):
Please delete the spec as well and we're good to go 👍 
```
rspec ./spec/controllers/api/fixed/sessions_controller_spec.rb:5
```

PR #578, opened 2024-03-08
AirNow data pull

BartkiewiczM:
! Before deployment create a user with username: 'US EPA AirNow' !

I've created a worker that will fetch the data from [AirNow API](https://docs.airnowapi.org/) every 56th minute of the hour (according to documentation at this point all data should be there).

We use 2 File Products:
[Hourly Data values](https://docs.airnowapi.org/docs/HourlyDataFactSheet.pdf)
[Monitoring site information](https://docs.airnowapi.org/docs/MonitoringSiteFactSheet.pdf)


app/models/air_now/stream.rb
```diff
@@ -0,0 +1,75 @@
+module AirNow
```
BartkiewiczM, line 1: naming will be changed

app/services/save_measurements.rb
```diff
@@ -135,8 +142,13 @@ def save(streams)
         .logger.warn "Stream.import failed for: #{import.failed_instances.inspect}"
     end
 
+    created_sessions = Session.where(id: session_ids)
```
BartkiewiczM, line 145: I wanted to somehow optimize this logic, cause it takes a significant amount of time if there is a lot of sessions to create, but there is not a lot after the first pull so optimization is not a must-be here.

mpolec (commented):
Thanks for your work on that 🙌 

I went through the code and I wonder whether we're able to simplify the logic in some places, so it will be easier to understand and reuse. I know you were relying on existing implementation and that's good, I just wouldn't like us be attached to some specifics which were needed for PurpleAir or OpenAQ integration and are not needed any more.

Some ideas I have:
- let's handle _locations_ and _hourly data_ in two separate steps:
  1. **locations:** based on data fetched from locations endpoint we can create/update session and stream records; there would be an additional column in sessions table needed to save `aqsid`.
  2. **hourly data:** based on data fetched from hourly data endpoint we can create measurement records, `aqsid` will be used to retrieve proper stream and session record.
  
  Some questions related to that:
    Do I understand correctly that `aqsid` is used to associate location with measurements and that we rely on it being present for fetched data?
    Do you maybe know how often does location data get changed? And what active/inactive status means?

- I would try to do all processing of measurement data together, so we don't iterate over the same array multiple times.
- there is `Http` service defined, maybe it could be reused to make the request to fetch the data.

What do you think about that?

BartkiewiczM:
Thank you for the comments!

AQSID is the parameter we are matching locations with measurements and we rely on it being present (I don't think there is any other reliable option).

The location data file is modified often. You can look up `last modified` column [here](https://files.airnowtech.org/?prefix=airnow/today/) - the file is named `monitoring_site_locations.dat`

I don't know exactly what it means to process the measurements in one place, is it about parsing, filtering, and normalizing?

I'm not sure there is a way to process locations and hourly data separately using the SaveMeasurements service. I think it's developed this way for performance and I wanted to use it cause it's a tested solution. Should we get rid of it?

app/services/air_now/process_measurements.rb
```diff
@@ -0,0 +1,51 @@
+class AirNow::ProcessMeasurements
+  def initialize(measurements)
+    @measurements = measurements
+  end
+
+  def call
```
mpolec, line 6: I would pass measurements as a param here instead of to the constructor.

app/services/air_now/process_measurements.rb
```diff
@@ -0,0 +1,51 @@
...
+    @measurements.each_with_object([]) do |measurement, processed_measurements|
+      next unless wanted_sensor_name?(measurement)
+
+      measurement[:parameter] = normalize_sensor_name(measurement)
```
mpolec, line 10: Is this needed? I see another call in `create_saveable_object`.

app/services/air_now/process_measurements.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  attr_reader :measurements
+
+  def wanted_sensor_name?(measurement)
+    ['PM2.5', 'O3', 'NO2', 'OZONE'].include?(measurement[:parameter])
+  end
+
+  def normalize_sensor_name(measurement)
+    measurement[:parameter] == 'OZONE' ? 'O3' : measurement[:parameter]
+  end
```
mpolec, line 25: I would pass `parameter` instead of the whole `measurement`. The other methods are written like this and I think it's quicker to understand.

app/services/air_now/process_measurements.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def convert_datetime(time, date)
+    full_year_date = Date.strptime(date, "%m/%d/%y").strftime("%Y-%m-%d")
+    Time.parse("#{full_year_date} #{time}")
+  end
+
+  def create_saveable_object(measurement)
+    time_utc = convert_datetime(measurement[:time], measurement[:date])
+    end_time_utc = time_utc + 1.hour
+    time_local = end_time_utc + measurement[:timezone].to_i.hours
```
mpolec, line 35: Time handling 🤔 I'll ask you offline as I don't understand it fully.

app/workers/air_now_import_measurements_worker.rb
```diff
@@ -0,0 +1,27 @@
...
+  # Unfortunately, files containing measurements belonging to sessions and streams duplicated
+  # in the database will not be processed anymore from that point on (because they would always
+  # raise). But duplicates should not happen because AirNowImportMeasurementsWorker runs serially.
+  #
+  # We could have solved the concurrent inserts with a different architecture (e.g. unique
+  # indexes in the database) but this seemed to be a simple and good enough solution.
+  sidekiq_options lock: :until_executed, on_conflict: :log
+
+  def perform
+    # return unless A9n.sidekiq_air_now_import_measurements_enabled
```
mpolec, line 23: It would be good to have such flag as well.

mpolec (commented):
I've added my comments, please take a look :)

app/services/save_measurements.rb
```diff
@@ -169,22 +183,27 @@ def save(streams)
...
                 value: measurement.value,
                 latitude: measurement.latitude,
                 longitude: measurement.longitude,
+                location: factory.point(measurement.latitude.to_f, measurement.longitude.to_f),
                 time: measurement.time_local,
                 timezone_offset: nil,
                 milliseconds: 0,
                 measured_value: measurement.value,
                 stream_id: stream_ids[i],
+                time_with_time_zone: measurement.time_local.in_time_zone.change(zone: time_zone)
```
mpolec, line 206: Can we use `timezone` parameter we get from AirNow data?

PR #581, opened 2024-03-18
Fetch fixed stream data

franiakat:
Fetch fixed stream data, save it in the redux store.


app/javascript/react/utils/getErrorMessage.ts
```diff
@@ -0,0 +1,29 @@
+// credits to KCDodds https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript
```
evulins, line 1: ❤️

app/javascript/react/utils/getErrorMessage.ts
```diff
@@ -0,0 +1,29 @@
...
+type ErrorWithMessage = {
+  message: string;
+};
+
+function isErrorWithMessage(error: unknown): error is ErrorWithMessage {
```
evulins, line 6: Out of curiosity  - could we use here a `const` assertion or there is a reason why we use in this file function, not a constant?

evulins (approved):
Looks good 👍  
image

app/javascript/react/utils/getErrorMessage.ts
franiakat: @evulins did you mean const declaration? 🙃 If so, I also stopped to think about it, I have used Kent's example directly and haven't adjusted to my regular choice to use const declaration. But actually I think it would be better to keep it unified, thanks.

PR #587, opened 2024-03-28
Calculate stream daily averages

mpolec:
Adding logic for calculating stream daily averages.

Values are calculated every hour starting from 0:30 is a station local time. Additionally, at the beginning of a day we recalculate value for the previous day to make sure we have a correct value.


BartkiewiczM:
Let's think about rounding up average values: 
- should it be done on the backend or front? 
- if backend - where? 

The question emerged from a conversation with our front-end team.

PR #592, opened 2024-04-03
Time zone translation script fix

BartkiewiczM:
The maintenance interval changed from 200mln to 50mln measurements. 
Only looping through not-processed sessions.

### DIFF
--- a/lib/tasks/populate_time_with_timezone.rake
+++ b/lib/tasks/populate_time_with_timezone.rake
@@ -1,56 +1,50 @@
 namespace :measurements do
   task populate_time_with_time_zone: :environment do
-    batch_size = 50_000
     sleep_time = 0.5
     total_processed = 0
     last_maintenance_at = 0
     maintenance_interval = 50_000_000
 
-    stream_ids_to_update_sql = <<-SQL
-      SELECT DISTINCT stream_id FROM measurements WHERE time_with_time_zone IS NULL
-    SQL
-    stream_ids_to_update = ActiveRecord::Base.connection.execute(stream_ids_to_update_sql).map { |row| row['stream_id'] }
-
-    session_ids_to_update_sql = <<-SQL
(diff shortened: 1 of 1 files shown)


lib/tasks/populate_time_with_timezone.rake
```diff
@@ -4,12 +4,25 @@ namespace :measurements do
...
+    session_ids_to_update = ActiveRecord::Base.connection.execute(session_ids_to_update_sql).map { |row| row['session_id'] }
+
+    total_to_update_sql = <<-SQL
+      SELECT COUNT(*) FROM measurements WHERE time_with_time_zone IS NULL
+    SQL
+    total_to_update = ActiveRecord::Base.connection.execute(total_to_update_sql).first['count'].to_i
     puts "Total measurements to update: #{total_to_update}"
 
-    Session.find_each(batch_size: 100) do |session|
+    Session.where(id: session_ids_to_update).find_each(batch_size: 100) do |session|
```
mpolec, line 25: How about doing it like that?
```
Session.select('streams.id, sessions.time_zone').distinct
.joins(streams: :measurements).where(measurements: { time_with_time_zone:  nil })
```
This way we fetch only data needed for the update - `stream_id` and `time_zone`


I've experienced some weird behaviour with _select_ method, not sure why it's working like that, but from what I tested it seems that the results are correct. Here is how a single record returned by this query looks like:
`#<Session:0x00005562ed5fb428 id: 2494073, time_zone: "Europe/Paris", tag_list: nil>`
- `id` refers to `streams.id`
- I have no idea why `tag_list` is added there 🤷

mpolec: And then having `stream_id`, we could simplify updating measurements:
line 36: `WHERE id IN (SELECT id FROM measurements WHERE stream_id = '#{stream_id})`

From what I've checked there are 2 377 866 measurements per stream at most, so maybe we can do it all at once? WDYT?

PR #594, opened 2024-04-04
Read stream id from query params

franiakat:
Reads the query param: `streamId` from the url and fetch data for the given fixed stream.
Renames calendars controller into fixed streams controller which it is.


BartkiewiczM (approved):
✅

PR #596, opened 2024-04-05
Account deletion with code confirmation endpoints

BartkiewiczM:
I've implemented two endpoints:
/api/user/delete_account_send_code
This endpoint saves a confirmation code for a logged-in user in the database. The code is valid for 30 minutes.
It sends a email to this user with the code for confirmation in the app.

/api/user/delete_account_confirm
This endpoint takes the code that is sent from the app and matches it with the one that's saved in the database.
If the code passes the validation the account is deleted.


app/controllers/api/delete_account_controller.rb
```diff
@@ -0,0 +1,49 @@
+module Api
+  class DeleteAccountController < BaseController
```
mpanuszewska, line 2: :)

spec/controllers/api/delete_account_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      it 'successfully deletes the user account' do
+        expect { post :delete_account_with_confirmation_code, params: { code: '1234' }, format: :json }
+        .to change(User, :count).by(-1)
+        expect(response).to have_http_status(:ok)
+        expect(json_response['message']).to eq('Account successfully deleted.')
+      end
+    end
+
+    context 'with an invalid or expired confirmation code' do
```
mpanuszewska, line 33: This is not really testing expired (but valid) confirmation codes, I'd probably add that too. Dates are always tricky.

mpanuszewska (approved):
Looks really good! 👏 

I've been wondering if we should cleanup old codes after they expire but couldn't really find a good reason so it's probably not reasonable. 

One small comment on the tests, otherwise you're good to go!

BartkiewiczM:
Thank you! I think for now the old endpoint will be used by people before they update the Android/iOS app.

PR #611, opened 2024-04-22
Flipper feature flag endpoint for frontend

BartkiewiczM:
Front-end needs to know when the feature flag is enabled 🚀


mpolec (approved):
👍

PR #613, opened 2024-04-24
Improve threshold editing behaviour

evulins:
- Added debouncing - now when the user types value, the range changes after clicking the Enter key or outside the input
- Added errors and validation.
- Fixed thumbs behaviour -  when they meet their neighbors (in terms of close values) they push each other.
- Fixed max and min values of middle thumbs - they can't have higher or lower values than max and min values.
-  Extracted handlers to separate files.

[Trello card ](https://trello.com/c/SIso24iI)


smialko (approved):
It's very impressive work! 
Working fine as far as I tested so I'm giving the approve, although the code was very complex and I might have missed some things 🥲

app/javascript/react/utils/thresholdEventHandlers.ts
```diff
@@ -0,0 +1,141 @@
...
+  const handleInputChange = (thresholdKey: keyof Thresholds, value: string) => {
+    const trimmedValue = value.trim();
+    setInputValue(trimmedValue);
+
+    if (trimmedValue === "") {
+      setErrorMessage(t("thresholdConfigurator.emptyInputMessage"));
+      return;
+    }
+
+    setInputValue(trimmedValue);
```
jankrzempek, line 37: For me at least, instead of setting InputValue twice, it is clearer to set it up only when the error appears and assign to input an empty value (I hope I understood the intention :())

```
 const trimmedValue = value.trim();

    if (trimmedValue === "") {
      setInputValue("");
      setErrorMessage(t("thresholdConfigurator.emptyInputMessage"));
      return;
    }

    setInputValue(trimmedValue);
```

app/javascript/react/utils/thresholdEventHandlers.ts
```diff
@@ -0,0 +1,141 @@
...
+      inputValue.trim() === ""
+    ) {
+      setErrorMessage(t("thresholdConfigurator.emptyInputMessage"));
+    }
+  };
+
+  const handleInputKeyDown =
+  (thresholdKey: keyof Thresholds) =>
+  (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === "Enter") {
```
jankrzempek, line 116: Can those be somewhere hardcoded in some const/utils etc? This seems to me so specific, that it would be rather better to check for ex. `Keys.Enter.value` instead of a simple string. One source of truth would be suitable here

app/javascript/react/utils/thresholdEventHandlers.ts
```diff
@@ -0,0 +1,141 @@
...
+      setInputValue(event.currentTarget.value);
+      handleInputChange(thresholdKey, event.currentTarget.value);
+    } else if (event.key === "ArrowUp" || event.key === "ArrowDown") {
```
jankrzempek, line 119: same with those keys

app/javascript/react/utils/thresholdEventHandlers.ts
```diff
@@ -0,0 +1,141 @@
...
+      event.preventDefault();
+      let newValue;
+      const step = event.shiftKey ? 10 : 1;
+
+      if (event.key === "ArrowUp") {
+        newValue = parseFloat(inputValue) + step;
+      } else {
+        newValue = parseFloat(inputValue) - step;
+      }
```
jankrzempek, line 128: What about

```
  let newValue;
      const step = event.shiftKey ? 10 : 1;

      const direction = event.key === "ArrowUp" ? 1 : -1;
      newValue = parseFloat(inputValue) + step * direction;
```

jankrzempek (changes requested):
Lots of smart, great, detailed work! 🚀
Requesting changes because of the movie and possible small error in there. Or is it an expected behavior? 

https://github.com/HabitatMap/AirCasting/assets/44276070/3c6209f4-9ac7-40b6-b25a-a085867c4ef3

app/javascript/react/utils/thresholdEventHandlers.ts
```diff
@@ -0,0 +1,141 @@
...
+        !validateValue(parsedValue, thresholdValues.min, thresholdValues.max)
+      ) {
+        setErrorMessage(
+          t("thresholdConfigurator.validValueMessage", {
+            minValue: thresholdValues.min,
+            maxValue: thresholdValues.max,
+          })
+        );
+      } else {
+        setErrorMessage("");
```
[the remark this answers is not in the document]
evulins: Yes

evulins:
> Lots of smart, great, detailed work! 🚀 Requesting changes because of the movie and possible small error in there. Or is it an expected behavior?
> 
>  possible_error.mov

I fixed possible errors. I don't know what you mean exactly in this video but this is expected behaviour:
- the user clicks on input and changes the value manually or by using arrows or using slider
- when user is using arrows or is typing numbers, the value is updated only when user clicks Enter or change the focus by clicking outside the active input
- when input is empty and the user clicks enter or clicks outside the input error message shows. After that when he moves the empty input or the input is focused again the previous value appears and error message is being removed.

smialko (approved):
Looks and works great!

jankrzempek (approved):
Great, big chunk of work! 🥳

I am pointing out two more concerns, but I am not sure if this ticket is about them:
1. When I change the highest threshold value to something lower, or it is just changed when editing value - the line stretches outside the screen.

PR #616, opened 2024-04-26
Page not found

smialko:
Add styling for page not found. 
The base branch is Kasia's branch because it had a ready-to-use setup for displaying the page. 

So if you want to review my code you can do it by reviewing just this commit: https://github.com/HabitatMap/AirCasting/pull/616/commits/97f45828eddcae74e24f630fa64c11b456c666c2 
or files: 
`[app/javascript/react/pages/NotFoundPage/NotFoundPage.tsx]`
(https://github.com/HabitatMap/AirCasting/pull/616#diff-ed105342d99960725e8e7d2260e76bc1bf887d42f9094ab94e5ed165149d718d)

`[app/javascript/react/pages/NotFoundPage/NotFoundPage.style.tsx]`
(https://github.com/HabitatMap/AirCasting/pull/616#diff-b07d411770829ef3d1960ee1678a6c9e7a75cae4823e3097d539b12e8c45bb91)


evulins (commented):
I left some  minor suggestions :)

app/javascript/react/pages/NotFoundPage/NotFoundPage.style.tsx
```diff
@@ -0,0 +1,19 @@
+import styled from "styled-components";
+
+const NotFoundContainer = styled.div`
```
jankrzempek, line 3: Can we approach it more universally, so it can be reused later easily if needed? I mean naming, what about just `Container` or `PageContainer`

app/javascript/react/pages/NotFoundPage/NotFoundPage.style.tsx
```diff
@@ -0,0 +1,19 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 50vh;
+`;
+const NotFoundContent = styled.div`
```
jankrzempek, line 9: Same here, what about `ContentContainer`? It is imported as NotFoundPage.style and I think it says enough about it's purpose, so we could shorten the names. It's up to you, just suggesting 😄

app/javascript/react/pages/NotFoundPage/NotFoundPage.tsx
```diff
@@ -1,7 +1,22 @@
...
 
-const NotFoundPage = () => {
-  return <div>Page Not Found</div>;
+const NotFoundPage: React.FC = () => {
+  return (
+    <NotFoundContainer>
+      <NotFoundContent>
+        <Title>404 - Page Not Found</Title>
+        <Description>
+          Oops! The page you’re looking for doesn’t exist.
```
jankrzempek, line 15: To translations, please

app/javascript/react/pages/NotFoundPage/NotFoundPage.tsx
```diff
@@ -1,7 +1,22 @@
...
+  Description,
+} from "./NotFoundPage.style";
 
-const NotFoundPage = () => {
-  return <div>Page Not Found</div>;
+const NotFoundPage: React.FC = () => {
+  return (
+    <NotFoundContainer>
+      <NotFoundContent>
+        <Title>404 - Page Not Found</Title>
```
jankrzempek, line 13: 1. To translations please
2. Will it ever be used for ex. 403 error or any other? If so, it could be adjusted to take number of error as param and give you message that matches the error/ or just generic message and proper error code - only when it is the case 😁

jankrzempek (approved):
Few suggestions/comments added. Nice work 🚀

app/javascript/react/pages/NotFoundPage/NotFoundPage.tsx
smialko: Nice, although I think it's overkill for MVP and don't want to spend too much time as it doesn't bring that much value ;D  I'll change it to just "Page not found", I wrote it as something generic but you're totally right, I shouldn't use invalid error code 🤡

PR #618, opened 2024-04-30
Reset values button

evulins:
Clicking "Reset to default" will reset all thresholds and inputs values.

[Trello card ](https://trello.com/c/uAehOSid/1798-reset-to-default-thresholds)

### DIFF
--- a/app/javascript/react/components/molecules/Calendar/HeaderToggle/HeaderToggle.tsx
+++ b/app/javascript/react/components/molecules/Calendar/HeaderToggle/HeaderToggle.tsx
@@ -42,24 +42,13 @@ const HeaderToggle: React.FC<Props> = ({
 
     return (
       <S.ResetButton onClick={resetThresholds}>
-        {isMobile && (
-          <>
-            <img
-              src={returnArrow}
-              alt={t("thresholdConfigurator.altResetButton")}
-            />
-            {t("thresholdConfigurator.resetButtonMobile")}
-          </>
-        )}
-        {!isMobile && (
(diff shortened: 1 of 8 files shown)


app/javascript/react/components/molecules/Calendar/HeaderToggle/HeaderToggle.tsx
```diff
@@ -29,25 +37,52 @@ const HeaderToggle: React.FC<Props> = ({ titleText, componentToToggle }) => {
...
+        {!isMobile && (
+          <>
+            {t("thresholdConfigurator.resetButton")}
+            <img
+              src={returnArrow}
+              alt={t("thresholdConfigurator.altResetButton")}
+            />
+          </>
+        )}
+      </S.ResetButton>
```
jankrzempek, line 63: I can see that this approach differs from the one that I chose. Let's talk after Majówka about which one we should continue to use because I think we should establish one rule for dealing with mobile/web.
My approach can be seen in `WeekView.style.tsx` and the use of `MobileLablel` and `DesktopLabel`

jankrzempek (approved):
Nice work! 🥳

app/javascript/react/components/molecules/Calendar/HeaderToggle/HeaderToggle.tsx
evulins: @jankrzempek after Michaels review this code is no longer needed.

Of course, we can still talk about mobile/desktop view approach

PR #620, opened 2024-05-06
Save threshold values

smialko:
The values have been stored and updated in the reducer. Now range colours affect the colours in the calendar

[Trello card ](https://trello.com/c/0nFDbPZH/1745-threshold-configurator-saves-values-in-the-store)


app/javascript/react/utils/thresholdEventHandlers.ts
```diff
@@ -3,7 +3,10 @@ import { useTranslation } from "react-i18next";
 
 import { updateAdjacentThresholds } from "./tresholdsUpdateAdjacent";
 import { KeyboardKeys } from "../types/keyboardKeys";
-import { SetStateAction, useEffect } from "react";
+import { useEffect } from "react";
+import { useDispatch } from "react-redux";
```
jankrzempek, line 7: those should be above, before new line as those are external right?

app/javascript/react/components/ThresholdConfigurator/ThresholdConfigurator.tsx
```diff
@@ -12,19 +12,16 @@ import {
 import * as S from "./ThresholdConfigurator.style";
 
 import HeaderToggle from "../molecules/Calendar/HeaderToggle/HeaderToggle";
-
-interface ThresholdsConfiguratorProps {
-  initialThresholds: Thresholds;
-}
+import { selectThreshold, updateAll } from "../../store/thresholdSlice";
+import { useDispatch, useSelector } from "react-redux";
```
jankrzempek, line 16: External one, shoudn't be before new line above?

jankrzempek (changes requested):
This AppDispatch makes me wonder, please take a look, nice work 🫡

app/javascript/react/utils/thresholdEventHandlers.ts
```diff
@@ -178,11 +179,13 @@ export const useThresholdHandlers = (
...
 
-    const resetThresholds = () => {
-      debouncedHandleInputChange.cancel();
-      setInputValue("");
-      setThresholdValues(initialThresholds);
-    };
+  const resetThresholds = () => {
+    debouncedHandleInputChange.cancel();
+    dispatch(resetToInitialValues());
+    setThresholdValues(initialState);
```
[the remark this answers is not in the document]
evulins: This is a tricky one. I didn't find a better solution for this case.  Values that we have from slice in the Threshold component are updated when the user changes the input values. 

This solution is not permanent. It will probably change again in the future once we get the initial values from the backend. Right now they are mocked.

jankrzempek (approved):
🚀

PR #624, opened 2024-05-07
AirNow - fetching data from the last 24h

BartkiewiczM:
Populate cache before deploy to production.

### DIFF
--- a/app/services/air_now/import_data.rb
+++ b/app/services/air_now/import_data.rb
@@ -4,8 +4,20 @@ class AirNow::ImportData
   end
 
   def call
+    current_utc = DateTime.now.new_offset(0).beginning_of_hour - 1.hour
+
+    hourly_data = ""
+
+    (current_utc - 24.hours).step(current_utc, 1/24.0).each do |utc_time|
+      data = http_client.get(hourly_data_endpoint(utc_time))
+      measurements_to_import = substract_cached_data(data, utc_time)
+      hourly_data.concat(measurements_to_import)
+
+      measurements_to_import_count = measurements_to_import.split("\n").size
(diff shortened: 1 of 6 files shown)


app/services/air_now/import_data.rb
```diff
@@ -4,8 +4,15 @@ def initialize(http_client: Http.new)
   end
 
   def call
+    current_utc = DateTime.now.new_offset(0).beginning_of_hour - 1.hour
```
mpolec, line 7: I would use `Time.current` just to be on a safe side.

https://thoughtbot.com/blog/its-about-time-zones

app/services/air_now/import_data.rb
```diff
@@ -18,10 +25,9 @@ def locations_endpoint
     'https://s3-us-west-1.amazonaws.com//files.airnowtech.org/airnow/today/monitoring_site_locations.dat'
   end
 
-  def hourly_data_endpoint
-    current_utc = DateTime.now.new_offset(0) - 1.hour
-    formatted_date = current_utc.strftime('%Y%m%d')
-    formatted_hour = current_utc.strftime('%H')
+  def hourly_data_endpoint(utc_time)
+    formatted_date = utc_time.strftime('%Y%m%d')
+    formatted_hour = utc_time.strftime('%H')
```
mpolec, line 30: Is that the same as below?
`formatted_time = utc_time.strftime('%Y%m%d%H')`
If so it would make it easier for me to read :)

mpolec (commented):
Added just some cosmetics suggestions, as discuss I think we can push it forward and improve on once we have some feedback from production.

PR #625, opened 2024-05-07
Fix tests

mpolec:
Fixes for:
/spec/models/stream_spec.rb
/spec/services/save_measurements_spec.rb


BartkiewiczM (approved):
Everything looks good ✅

PR #629, opened 2024-05-08
Fetching additional stream daily averages

BartkiewiczM:


app/controllers/api/v3/stream_daily_averages_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+  module V3
+    class StreamDailyAveragesController < BaseController
+      def show
+        averages = StreamDailyAveragesRepository.new.from_time_range(stream_id: params[:id], start_date: params[:start_date], end_date: params[:end_date])
+        serialized_averages = StreamDailyAveragesSerializer.new.call(averages)
+
+        if serialized_averages.present?
+          render json: serialized_averages, status: :ok
+        else
+          render json: { error: 'No data available' }, status: :not_found
```
mpolec, line 11: Not sure about cases when there is no data - should it empty array or an error? 🤔

app/controllers/api/v3/stream_daily_averages_controller.rb
```diff
@@ -0,0 +1,16 @@
+module Api
+  module V3
+    class StreamDailyAveragesController < BaseController
+      def show
```
mpolec, line 4: Maybe we could use interactor here and move into direction of using `Result` to handle success/failure scenarios. WDYT?

mpolec (commented):
Looking good 😎 Left some comments, please take a look 👁️

config/routes.rb
```diff
@@ -101,6 +101,7 @@
 
     namespace :v3 do
       resources :fixed_streams, only: %i[show]
+      resources :stream_daily_averages, only: %i[show]
```
mpolec, line 104: Sorry, haven't realised that before, but I think it should be `index` not `show` as we return collection of `stream_daily_averages`.

app/interactors/fixed_streams/stream_daily_averages_interactor.rb
```diff
@@ -0,0 +1,23 @@
+module FixedStreams
+  class StreamDailyAveragesInteractor
```
mpolec, line 2: When it comes to naming, I would name a module the same as its corresponding controller, and a class after the action it refers to.

```
module StreamDailyAverages
  class IndexInteractor
```

app/interactors/fixed_streams/stream_daily_averages_interactor.rb
```diff
@@ -0,0 +1,23 @@
...
+      stream_daily_averages_serializer: StreamDailyAveragesSerializer.new
+    )
+      @stream_daily_averages_repository = stream_daily_averages_repository
+      @stream_daily_averages_serializer = stream_daily_averages_serializer
+    end
+
+    def call(stream_id:, start_date:, end_date:)
+      averages = stream_daily_averages_repository.from_time_range(stream_id: stream_id, start_date: start_date, end_date: end_date)
+      serialized_averages = stream_daily_averages_serializer.call(averages)
+      serialized_averages ? nil : serialized_averages = []
```
mpolec, line 14: Is that needed?

spec/repositories/stream_daily_averages_repository_spec.rb
```diff
@@ -21,4 +21,33 @@
...
+        create_stream_daily_average!({ stream: stream, date: 1.day.ago })
+      stream_daily_average2 =
+        create_stream_daily_average!({ stream: stream, date: 2.days.ago })
+      old_stream_daily_average =
+        create_stream_daily_average!({ stream: stream, date: 5.months.ago })
+      create_stream_daily_average!
+
+      result1 = subject.from_time_range(stream_id: stream.id, start_date: 3.days.ago, end_date: 1.day.ago)
+      result2 = subject.from_time_range(stream_id: stream.id, start_date: 3.days.ago, end_date: 2.days.ago)
+      result3 = subject.from_time_range(stream_id: stream.id, start_date: 6.months.ago, end_date: 1.day.ago)
```
mpolec, line 38: I think it's enough to have one check here. In general, I would go with an approach that in one test case (so within `it` block) a tested method (`from_time_range` in this case) is called once. 

If there is something more to test, for example some edge cases, I would add them as separate test cases and add `context` block if needed.

spec/requests/stream_daily_averages_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      }
+    ]
+
+    get "/api/v3/stream_daily_averages/#{stream.id}?start_date=#{Date.current.prev_day}&end_date=#{Date.current}"
+
+    expect(response.status).to eq(200)
+    expect(JSON.parse(response.body)).to eq(expected_response.as_json)
+  end
+
+  it 'returns not_found status' do
```
mpolec, line 35: Can you update the description, please?
I would also add `context` here, so it's clear that we're getting `[]` when there is no data.

mpolec (commented):
Nice, thanks for the improvements 🙌 
Some small fixes to be done and we're good to go 🐐

mpolec (approved):
💯

PR #635, opened 2024-05-14
[WIP] Default thresholds + fixed stream endpoint with thresholds

BartkiewiczM:


app/repositories/default_thresholds_repository.rb
```diff
@@ -0,0 +1,5 @@
+class DefaultThresholdsRepository
+  def stream_thresholds(stream:)
```
mpolec, line 2: `find` or `find_for` ?

db/migrate/20240513144147_create_default_thresholds.rb
```diff
@@ -0,0 +1,13 @@
+class CreateDefaultThresholds < ActiveRecord::Migration[6.1]
+  def change
+    create_table :default_thresholds do |t|
```
mpolec, line 3: I would add `timestamps` as well.

mpolec: And consider combined index for `sensor_name` and `unit_symbol`.

db/schema.rb
```diff
@@ -189,6 +204,12 @@
     t.index ["name"], name: "index_tags_on_name"
   end
 
+  create_table "temp_session_offsets", id: :bigint, default: nil, force: :cascade do |t|
+    t.decimal "latitude"
+    t.decimal "longitude"
+    t.text "timezone"
+  end
+
```
mpolec, line 212: Can you check the db migrate status on you machine? It seems there are some old and no needed things.

app/serializers/threshold_serializer.rb
```diff
@@ -0,0 +1,11 @@
+class ThresholdSerializer
+  def call(thresholds)
+    {
+      min: thresholds[:threshold_very_low],
+      low: thresholds[:threshold_low],
+      middle: thresholds[:threshold_medium],
+      high: thresholds[:threshold_high],
+      max: thresholds[:threshold_very_high],
```
mpolec, line 8: I think I would add this to stream serializer. We don't have any place where threshold data is needed independently. I think it will be easier to grasp when we have it in one file.

app/repositories/default_thresholds_repository.rb
```diff
@@ -0,0 +1,5 @@
...
+    DefaultThreshold.find_by(sensor_name: stream.sensor_name, unit_symbol: stream.unit_symbol)
```
mpolec, line 3: I think I would move here logic returning values from stream when there is no default config available.

mpolec (commented):
Looking good. I like the idea with the new abstraction for the default configuration 🙌 

I would move some pieces around to have class responsibilities more clearly defined, we can talk about it tomorrow ;)

lib/tasks/populate_threshold_sets.rake
```diff
@@ -0,0 +1,18 @@
+namespace :threshold_sets do
```
mpolec, line 1: How are we going to mark default sets?

app/serializers/stream_serializer.rb
```diff
@@ -0,0 +1,24 @@
...
+      update_frequency: update_frequency,
+      last_update: stream.session.last_measurement_at,
+      session_id: stream.session.id,
+      end_time: stream.session.end_time_local,
+      start_time: stream.session.start_time_local,
+      min: stream.threshold_very_low.to_s,
+      low: stream.threshold_low.to_s,
+      middle: stream.threshold_medium.to_s,
+      high: stream.threshold_high.to_s,
+      max: stream.threshold_very_high.to_s,
```
mpolec, line 21: I think logic for threshold went missing, on master we have:
https://github.com/HabitatMap/AirCasting/blob/c51d027d2f77d69e760e637bc1c585af43180bda/app/serializers/fixed_stream_serializer.rb#L13

db/migrate/20240513144147_create_thresholds_set.rb
```diff
@@ -0,0 +1,17 @@
...
+  def change
+    create_table :threshold_sets do |t|
+      t.string :sensor_name
+      t.string :unit_symbol
+      t.boolean :is_default
+      t.float :threshold_very_low
+      t.float :threshold_low
+      t.float :threshold_medium
+      t.float :threshold_high
+      t.float :threshold_very_high
```
mpolec, line 11: A wondering:
On the frontend we have min, low, middle, high, max. Is it something we'd like to have across the whole codebase?

mpolec (commented):
Added some comments, please take a look 👀 
Could you check specs, please?
`rspec ./spec/services/air_now/import_data_spec.rb:5` is falling for me.

lib/tasks/populate_threshold_sets.rake
BartkiewiczM: I think we can do this manually after it's implemented.

PR #636, opened 2024-05-14
Adjust logic for stream daily average calculation for AirNow streams

mpolec:


app/repositories/measurements_repository.rb
```diff
@@ -3,15 +3,13 @@ def from_last_24_hours(stream_id:)
     Measurement.where(stream_id: stream_id).reorder(time: :desc).limit(1440)
   end
 
-  def stream_daily_average_value(stream_id:, beginning_of_day:)
-    end_of_day = beginning_of_day + 1.day
-
+  def stream_daily_average_value(stream_id:, time_with_time_zone:)
     Measurement
       .where(
         'stream_id = ? AND time_with_time_zone >= ? AND time_with_time_zone < ?',
```
mpolec, line 9: My bad, I've missed it - as we use `end_of_day` right now, we should have `<=` in the last condition:
`stream_id = ? AND time_with_time_zone >= ? AND time_with_time_zone <= ?`

spec/services/stream_daily_averages/fixed_active_sessions_traverser_spec.rb
```diff
@@ -17,10 +17,23 @@
...
+    end
+
+    it 'traverses over AirNow streams and calls streams interactor for associated stream' do
+      time_zone = 'Europe/Warsaw'
+      user = User.create!(email: 'airnow@example.com', username: 'US EPA AirNow', password: 'password')
+      fixed_session = create_session!(type: 'FixedSession', time_zone: time_zone, user: user)
+      stream = create_stream!(session: fixed_session)
+
+      expect(stream_interactor).to receive(:call).with(
+        {stream_id: stream.id, time_zone: time_zone, is_air_now_stream: true}
```
mpolec, line 36: I would add `context` block here to clearly indicate that there is another case for AirNow stream.

mpolec (commented):
Could you also adjust this spec?
`spec/repositories/fixed_sessions_repository_spec.rb`
(There is a check for loaded associations).

BartkiewiczM (approved):

PR #652, opened 2024-05-27
Calendar controls - improvements

jankrzempek:
Relates to both - Michaels comments:
- https://trello.com/c/Oht690FY/1795-controls-for-viewing-earlier-months
- https://trello.com/c/9kFI8CsY/1874-calendar-navbar-data-range

We need end time to handle disabling and enabling calendar control in well manner. It can be different for both mobile and fixed so we need to make sure to make it as best as possible.
1. For FIXED there is no end_time so we have to use last_update.
2. For MOBILE there is no last_update so we have to use end_time.
3. In case we're missing something, let's assume that the current date is the one we can't pass because there can be no data from the future


evulins (approved):
🚀

PR #658, opened 2024-05-30
Date format one source and implementation - calendar

jankrzempek:
Michael comment on task: https://trello.com/c/9kFI8CsY/1874-calendar-navbar-data-range

### DIFF
--- a/app/javascript/react/types/dateFormat.ts
+++ b/app/javascript/react/types/dateFormat.ts
@@ -1,0 +1,6 @@
+enum DateFormat {
+    default = "YYYY-MM-DD",
+    us = "MM/DD/YYYY",
+}
+
+export { DateFormat }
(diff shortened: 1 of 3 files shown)


app/javascript/react/types/dateFormat.ts
```diff
@@ -0,0 +1,6 @@
+enum DateFormat {
+    default = "YYYY-MM-DD",
+    us = "MM/DD/YYYY",
```
smialko, line 3: Just thinking out loud - why do we have two differents formats, not just the US one? Asking because I saw Michael's comment that he wants to have an American date format 🤔

smialko (approved):
Looking good in general, just one question :)

evulins (approved):

app/javascript/react/types/dateFormat.ts
jankrzempek: Good point, so I just checked to ensure myself. The reason is that the backend sends us the date in default format (let's name it like that). To receive it we need to have default type, and US type is needed - you know why 😄

PR #663, opened 2024-06-05
UI changes to make it similar to production

smialko:
Small UI changes discussed in comment with Michael here: https://trello.com/c/alrOo9tK/1814-display-sessions-fixed

### DIFF
--- a/app/javascript/react/components/Map/Markers/SingleMarker/SingleMarker.style.tsx
+++ b/app/javascript/react/components/Map/Markers/SingleMarker/SingleMarker.style.tsx
@@ -20,8 +20,8 @@ const MarkerContainer = styled.div`
 `;
 
 const ShadowCircle = styled.div<MarkerProps>`
-  height: 5rem;
-  width: 5rem;
+  height: 4rem;
+  width: 4rem;
   border-radius: 50%;
   background: radial-gradient(
     circle at center,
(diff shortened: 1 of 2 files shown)


app/javascript/react/components/Map/Markers/SingleMarker/SingleMarker.style.tsx
```diff
@@ -53,16 +53,16 @@ const DataContainer = styled.div`
...
+  height: 1.2rem;
   border-radius: 50%;
   background-color: ${(props) => props.color};
   pointer-events: none;
 `;
 
 const MarkerText = styled(H4)`
   color: ${gray400};
-  font-size: 1.2rem;
+  font-size: 12px;
```
evulins, line 65: Is it different size from 1.2rem?

evulins (approved):

app/javascript/react/components/Map/Markers/SingleMarker/SingleMarker.style.tsx
smialko: According to chatgpt yes 🤔

smialko: Should I change it?

jankrzempek (commented):
Everything perfect! One thing I am not sure about, is the outline of the session. currently, on prod each "marker" has an outline of the color that matches the threshold and Michael says in the comment: "when the outline is removed, the visual contrast between the markers and the grey default map is reduced, which to me doesn’t look good". How we approach this?

app/javascript/react/components/Map/Markers/SingleMarker/SingleMarker.style.tsx
evulins: If it looks the same i would keep 1.2rem if not 12px

smialko:
@jankrzempek , the latest Michael's comment about outline: please retain the feature of the new marker design that alters the halo of the selected session. this looks good and is an effective way to indicate a relation between the graph and the selected marker.

PR #667, opened 2024-06-06
Accept int or float in east/west/north/south values in fixed_sessions requests

BartkiewiczM:


app/models/api/fixed_sessions.rb
```diff
@@ -8,6 +8,12 @@ module Types
 
   Schema =
     Dry::Validation.Schema do
+      configure do
+        def number?(value)
+          true if Float(value)
```
mpolec, line 13: I think we need to adjust it as for non numeric values it will return `ArgumentError`.
```
Float("a")
ArgumentError: invalid value for Float(): "a"
from (pry):6:in `Float'
```

I guess we could go with something like: 
`value.is_a?(Float) || value.is_a?(Integer)`

BartkiewiczM: Perfect, pushed.

mpolec (approved):
Great 🙌

PR #668, opened 2024-06-06
Update frequency changed for AirNow streams

BartkiewiczM:
For now, I don't have a better idea of how to adjust this logic. 
I did not want to change the db schema (adding update frequency to sessions) for this small change as it would be more complex and we don't have a guarantee of updates in this time interval, so it would not be true always.

### DIFF
--- a/app/serializers/fixed_stream_serializer.rb
+++ b/app/serializers/fixed_stream_serializer.rb
@@ -12,6 +12,7 @@ class FixedStreamSerializer
 
   def serialized_stream(stream)
     thresolds = Stream.thresholds(stream.sensor_name, stream.unit_symbol)
+    update_frequency = stream.session.user.username == 'US EPA AirNow' ? '1 hour' : '1 minute'
 
     {
       id: stream.id,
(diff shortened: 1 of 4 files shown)


app/serializers/fixed_stream_serializer.rb
```diff
@@ -12,6 +12,7 @@ def call(stream:, measurements:, stream_daily_averages:)
 
   def serialized_stream(stream)
     thresolds = Stream.thresholds(stream.sensor_name, stream.unit_symbol)
+    update_frequency = stream.session.user.username == 'US EPA AirNow' ? '1 hour' : '1 minute'
```
mpolec, line 15: There is `delegate` used in the Session class:
https://github.com/HabitatMap/AirCasting/blob/294f1969339a6310af7084995757e7f25b358fe9/app/models/session.rb#L25

Thanks to delegation we can call `username` directly on a session object, so
`stream.session.user.username` can be shorten to `stream.session.username`.

spec/requests/fixed_streams_spec.rb
```diff
@@ -58,6 +58,41 @@
       expect(response.status).to eq(200)
       expect(JSON.parse(response.body)).to eq(expected_response.as_json)
     end
+
+    it 'returns 1h update frequency for US EPA AirNow user' do
```
mpolec, line 62: I wonder how we should test this change 🤔 
My thoughts:
There are two possible cases for update frequency value - we can get `1 hour` or `1 minute`. I would say we should test both of them. That's why I would add a unit test for the serializer and check it there. This spec I would leave as it was.
WDYT?

mpolec (commented):
Nice, I agree it's good enough for now solution 👍

mpolec (approved):
LGTM 😎

PR #677, opened 2024-06-14
Graph

evulins:
Done:

displaying graph,
displaying scrollbar,
displaying tooltip,
displaying custom arrows for navigation,
while using scrollbar, we are also changing values in Stream info.
Scrollbar is not styled as it is in the designs.

[Trello card](https://trello.com/c/dvX1cq18/1828-session-details-graph)


app/javascript/react/components/Map/Map.tsx
```diff
@@ -217,6 +217,25 @@ const Map = () => {
...
+            onMarkerClick={handleMarkerClick}
+            selectedStreamId={selectedStreamId}
+          />
+        )}
+        {selectedStreamId && !fixedSessionTypeSelected && (
+          <StreamMarkers
+            sessions={mobileStreamPoints}
+            unitSymbol={unit_symbol}
+          />
+        )}
```
JoannaThelema, line 238: I think this shouldn't be in this PR. I'll clean up this in `mobile session fixes` 💪

PR #678, opened 2024-06-17
Fix for mobile session

JoannaThelema:
and 
>markers change location as you zoom, this is likely related to the item above.

Additionally, I found out that centering the marker 
```
  position: absolute;
  transform: translate(-50%, -50%); /* Center the marker */
```
is breaking the tooltip so I implementing a dirty workaround `#DirtyButWorks` 🙈 

### What was done:
* I renamed `SingleMarker` to `SessionFullMarker` and I added `SessionDotMarker`
* I separated the `StreamMarker` from the `SessionDotMarker`,  which is visible when multiple markers overlap. Styles are mostly copied from `SessionFullMarker` to ensure the dot is visible in the same position. 
* tooltip fix


app/javascript/react/components/Map/Markers/MobileMarkers.tsx
```diff
@@ -115,14 +114,13 @@ const MobileMarkers = ({
...
-        <MarkerContainer
+        <DotMarker
+          color={red}
           onClick={() => {
             onMarkerClick(Number(session.point.streamId), Number(session.id));
             centerMapOnMarker(session.point);
           }}
-        >
-          <StreamMarker color={red} />
-        </MarkerContainer>
```
JoannaThelema, line 125: I separated the stream marker from the dot marker visible when multiple markers are overlapping. Styles are mostly copied from SingleMarker to ensure the dot is visible in the same position.

app/javascript/react/components/Map/Map.tsx
```diff
@@ -217,25 +212,6 @@ const Map = () => {
...
-            onMarkerClick={handleMarkerClick}
-            selectedStreamId={selectedStreamId}
-          />
-        )}
-        {selectedStreamId && !fixedSessionTypeSelected && (
-          <StreamMarkers
-            sessions={mobileStreamPoints}
-            unitSymbol={unit_symbol}
-          />
-        )}
```
JoannaThelema, line 238: Leftover after graph rebase 💪

evulins (approved):
👍

PR #691, opened 2024-07-01
Refresh button

bazelajoanna:
Style refresh button

image

image

[Trello task](https://trello.com/c/sAQMX84m/1817-fixed-mobile-refresh)

### DIFF
--- a/app/javascript/react/components/RefreshMapButton/RefreshMapButton.style.tsx
+++ b/app/javascript/react/components/RefreshMapButton/RefreshMapButton.style.tsx
@@ -10,7 +10,7 @@ const RefreshButton = styled(Button)`
   background-color: ${blue};
   font-size: 1.4rem;
   padding: 0 1rem;
-  height: 4rem;
+  height: 3rem;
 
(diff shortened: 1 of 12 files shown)


app/javascript/react/components/RefreshMapButton/RefreshMapButton.style.tsx
```diff
@@ -0,0 +1,27 @@
...
+    margin: 0 16px;
+    height: 42px;
+  }
+
+  @media ${media.largeDesktop} {
+    font-size: 16px;
+    padding: 0 50px;
+  }
+`;
+
```
evulins, line 26: Same here, can we use `rem` for padding, margin, font-size, height, width,

JoannaThelema (approved):
not tested, but the code looks good 🙌

evulins (approved):
Nice work! 

I was wondering if we can change the styling for this button on mobile devices. For instance, we could make button a little more bigger( we can set different padding)

Screenshot 2024-07-01 at 10 27 09

PR #698, opened 2024-07-08
Add MapButtons component

JoannaThelema:


app/javascript/react/components/MapButtons/MapButtons.tsx
```diff
@@ -0,0 +1,139 @@
...
+    setShowConfirmation(true);
+  };
+
+  const updateButtonPosition = () => {
+    if (buttonRef.current) {
+      const rect = buttonRef.current.getBoundingClientRect();
+      setButtonPosition({ top: rect.top, left: rect.left });
+    }
+  };
+
```
bazelajoanna, line 64: I'd close confirmation popup after 3s, as we have in modal.

```
useEffect(() => {
    if (showConfirmation) {
      const timer = setTimeout(() => setShowConfirmation(false), 3000);
      return () => clearTimeout(timer);
    }
  }, [showConfirmation]);
```

JoannaThelema: Good catch! Done 🙌

bazelajoanna (approved):

Sonje (approved):

PR #703, opened 2024-07-12
Save and update params in URL

evulins:


JoannaThelema:
Continued in: https://github.com/HabitatMap/AirCasting/pull/715

PR #705, opened 2024-07-16
Fix reset ranges design button on mobile

Sonje:
**Changes:**
- reuse `ResetButton` on `CalendarPage` instead of adding and styling a new one
- added grid within reset button wrapper to make sure it's behaving as intended in all cases we have (centered with accurate space for icon) (I can use flex if you want, but I'm a big grid lover as you might have noticed)
- fixed reset button text from reset legend to reset ranges (probably has been missed during rebases)
- setup default reset button text to desktop version, so if no one provides Reset Button with differentt txt reset to default will be displayed
- make sure the button width on mobile is not smaller than 100px -> if the button is smaller in a future we might just use the smaller gaps or reuse no text, only icon button or whatever - I did not focus on that as we don't have a design for it and these cases would be rare)
- simplify reset Button logic a bit, as it was kinda complicated

### DIFF
--- a/app/javascript/react/components/Map/Map.style.tsx
+++ b/app/javascript/react/components/Map/Map.style.tsx
@@ -64,21 +64,21 @@ const ThresholdContainer = styled.div`
   box-shadow: 2px 2px 4px 0px ${colors.gray900};
 
   @media (${media.desktop}) {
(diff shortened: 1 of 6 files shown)


Sonje:

bazelajoanna (approved):

PR #706, opened 2024-07-16
Fix map legend being stretched out on mobile iOS/Safari

Sonje:
This also causes the close button to be hidden behind other UI elements (another issue reported). Fixed by removing modal's height, it should be propagated by grid without a need for specific height.

Before:

After:
Simulator Screenshot - iPhone 15 - 2024-07-16 at 13 18 01 | width=100


Sonje:

PR #710, opened 2024-07-17
Fix input field click zoom in the view on ios

Sonje:
Clicking input field on Map Legend thresholds caused a whole view zoom in

This happens because iOS requires 16px minimum font size, but we use 12px in our designs and I followed them. 
I changed the font-size and threshold width a bit to adjust the view.


Sonje:

bazelajoanna (approved):

JoannaThelema (approved):

PR #711, opened 2024-07-18
Create General Filters component

bazelajoanna:
Create SessionFilters component with toggle button to switch between Fixed and Mobile sessions.

image


app/javascript/react/components/Map/Map.tsx
```diff
@@ -93,15 +91,14 @@ const Map = () => {
   const [selectedSessionId, setSelectedSessionId] = useState<number | null>(
     null
   );
-  const [selectedSessionType, setSelectedSessionType] = useState<SessionType>(
-    SessionTypes.FIXED
-  );
   const [selectedStreamId, setSelectedStreamId] = useState<number | null>(null);
 
+  // Selectors
+  const selectedSessionType = useAppSelector(selectSelectedSessionType);
```
JoannaThelema, line 97: I didn't know we have `useAppSelector` 😅 Makes sense to use it everywhere for consistency but this is a note for me 🙈

app/javascript/react/components/SessionFilters/SessionFilters.style.tsx
```diff
@@ -0,0 +1,50 @@
...
+  color: ${(props) => (props.$isActive ? white : gray300)};
+  cursor: pointer;
+
+  &:first-child {
+    border-radius: 10px 0 0 10px;
+  }
+
+  &:nth-child(2) {
+    border-radius: 0 10px 10px 0;
+  }
```
JoannaThelema, line 38: Nice! I didn't know this trick 👍

JoannaThelema (approved):
To be consistent I would name all icons in the same pattern eg: `pinIcon` WDYT?

Overall, beautiful buttons 😄

Sonje (approved):

app/javascript/react/components/Map/Map.tsx
bazelajoanna: I didn't know either :D I've just noticed that so I decided to use :)

PR #712, opened 2024-07-19
Display session type filters on mobile device

bazelajoanna:
- add "Filters" button on Map to display filters on mobile devices,
- add filters modal with session type filters
image


app/javascript/react/components/SessionFilters/MobileSessionFilters.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  let sessionsCount;
+
+  useMemo(() => {
+    if (selectedSessionType === SessionTypes.FIXED) {
+      sessionsCount = fixedSessionsState.sessions.length;
+    } else if (selectedSessionType === SessionTypes.MOBILE) {
+      sessionsCount = mobileSessionsState.sessions.length;
+    }
+  }, [fixedSessionsState, mobileSessionsState, selectedSessionType]);
```
Sonje, line 32: to utilize `useMemo` optimization it would be better to use `useMemo` to directly set the _sessionsCount_. 

I'd consider using a `switch` statement instead of `if/else` - while both are valid, switch handles memory allocation and memory address usage more efficiently. I have no strong feelings about that, but I think it might be better in the Aircasting case handling lots of data.

Sonje (approved):
Nice job! I have added an optimization suggestion - you can ignore it, but I think it might improve performance in a long run.

JoannaThelema (commented):
The code looks good 🙌 I found one bug:
For 1024 width there are no filters:
Screenshot 2024-07-19 at 20 05 42

app/javascript/react/components/SessionFilters/MobileSessionFilters.tsx
bazelajoanna: Thank you!

JoannaThelema (approved):

Sonje:
Ready, set, merge! 🎉

PR #713, opened 2024-07-19
Fix markers appearing while they should remain clustered

Sonje:
### Problem:
The pulsation solution was affecting clusters and markers incorrectly. When a session was assigned a `pulsatingSessionId`, the markers were not clustering correctly and appeared individually instead of within a cluster. It was caused by incorrect pulsation clearing and marker clustering.

### Fixes:

- handle pulsating markers: refactored the code around pulsating markers - implemented a separate `MarkerClusterer` for handling pulsating markers to ensure they do not interfere with the main clusterer
- clear previous pulsating markers: before applying pulsation, clear any existing pulsating markers to avoid conflicts
- maintain clusters: ensured that markers remain part of the main clusterer even when they are pulsating by adding markers back to the main clusterer after the pulsation ends
- properly reset clusters: when `pulsatingSessionId` is set to null, clear the pulsating cluster and ensure the main clusterer is updated with all valid markers
- prevent markers from refreshing: to avoid refreshing all other markers when the pulsation stops, memoized the sessions and markers using `useMemo` to minimize unnecessary re-renders

### Optimizations:

- memoize sessions and markers: used `useMemo` to memoize sessions and markers to avoid unnecessary re-renders and improve performance
- update only relevant parts: ensured that only the pulsating markers or clusters are updated without affecting the rest of the markers, leading to a smoother user experience
- stable references for callbacks: ensured stable references for the `centerMapOnMarker` and `updateClusterer` functions by using `useCallback`
- type safety for callbacks: explicitly typed the callback functions for better type safety and readability

### Steps to reproduce the initial problem:

1. Click on a cluster on loaded map
2. Hover over sessions on session list
3. Zoom out so the markers should be clustered
4. Notice that hovered sessions are not being clustered
5. To re-cluster them we are forced to hover over session that should be clustered

### DIFF
--- a/app/javascript/react/components/Map/Markers/FixedMarkers.tsx
+++ b/app/javascript/react/components/Map/Markers/FixedMarkers.tsx
@@ -4,6 +4,7 @@ import React, {
   useLayoutEffect,
   useRef,
   useState,
+  useMemo,
 } from "react";
 import { useSelector } from "react-redux";
 
@@ -64,6 +65,10 @@ const FixedMarkers = ({
     null
   );
 
+  // Memoize the sessions and markers to avoid unnecessary re-renders
+  const memoizedSessions = useMemo(() => sessions, [sessions]);
+  const memoizedMarkers = useMemo(() => markers, [markers]);
+
   // Initialize clusterer
   useLayoutEffect(() => {
     if (map) {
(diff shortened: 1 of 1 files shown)


Sonje:
[fixed: markers are not clustering properly when zooming out, clustering is triggered by hovering over sessions list](https://trello.com/c/jwFCIiCH/1915-fixed-markers-are-not-clustering-properly-when-zooming-out-clustering-is-triggered-by-hovering-over-sessions-list)

bazelajoanna (approved):

PR #714, opened 2024-07-23
Fix modal for small screen sizes and arrow in the graph

evulins:
Changes made in this PR:

- disabled  arrows on the graph when there is no more data
- fix displaying modal with graph on small desktop and tablet screen sizes

[Trello task](https://trello.com/c/Edp8dmhP/1922-modal-for-tablet-sizes-does-not-scale-well)

### DIFF
--- a/app/javascript/react/components/Map/Map.tsx
+++ b/app/javascript/react/components/Map/Map.tsx
@@ -47,6 +47,7 @@ import {
 import { SessionTypes } from "../../types/filters";
 import { SessionList } from "../../types/sessionType";
 import { UserSettings } from "../../types/userStates";
+import { screenSizes } from "../../utils/media";
 import { pubSub } from "../../utils/pubSubManager";
 import useMobileDetection from "../../utils/useScreenSizeDetection";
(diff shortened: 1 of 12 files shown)


app/javascript/react/components/Map/Map.tsx
```diff
@@ -79,6 +80,8 @@ const Map = () => {
   const navigate = useNavigate();
   const { t } = useTranslation();
 
+  const isMobile1 = useMobileDetection(screenSizes.smallDesktop);
+
```
Sonje, line 84: I think it's not used anywhere, we can get rid of it, unless you had an idea what to do with it?

Sonje (approved):
Thanks @evulins for really thorough work around the issues in graph and TC on different resolutions. I added a comment about a possible leftover in the code in the Map component, otherwise your PR looks good code-wise and functionality-wise.
 
**Side note**: Good job with doing that mixed design for tablet widths (768<width<1024) as we talked in the office.

app/javascript/react/components/Map/Map.tsx
evulins: Sorry about that, this is leftover.

PR #716, opened 2024-07-23
Fix low longitude altitude mobile sessions being covered by a session modal

Sonje:
I decided to go with this simple solution. 

### Simple Solution
This is "proste jak szpadel" (eng. as simple as a spade) solution with a few hardcoded values that will work unless device has height of 300 pixels which is not possible these days. I added some hardcoded offset for low latitude altitudes that always look good. Nothing fancy. I also decoupled the functions so it's easier to tackle it in the future if this solution is not enough.

[Here's a screen recording of the changes(too big for github attachment)](https://drive.google.com/file/d/1V1l8FrrRHt6RaSbAUJv2feyKWs0NnaWp/view?usp=sharing)

### Story
Long story short: I have a work in progress calculations of visible map view in comparison to open session modal and recalculations after each session modal height change, the fitbounds function we are now using won't work, because the map bounds will always be full vh, so when I started recalculating zoom, lat/lng fractions, and taking into account Globe I knew I started rewriting the google library, so I decided to stop (remnants are available here: [branch:stop-rewriting-google-maps-library-because-of-a-stupid-modal](https://github.com/HabitatMap/AirCasting/tree/stop-rewriting-google-maps-library-because-of-a-stupid-modal))


Sonje:
[mobile device > mobile sessions: center session inside map area](https://trello.com/c/KRCLB3V3/1925-mobile-device-mobile-sessions-center-session-inside-map-area)

evulins (approved):
Code looks good. I tested it and seems to be ok (looks the same as on the recording that you attached) Great job! 👏

PR #718, opened 2024-07-25
feat: crowdmap

JoannaThelema:
Ticket: [Display Crowdmap (Mobile)](https://trello.com/c/jVYeXbfq/1854-display-crowdmap-mobile)

### DIFF
--- a/app/javascript/react/components/Map/Markers/CrowdMapMarkers.tsx
+++ b/app/javascript/react/components/Map/Markers/CrowdMapMarkers.tsx
@@ -8,6 +8,7 @@ import {
   selectCrowdMapRectangles,
 } from "../../../store/crowdMapSlice";
 import { useAppDispatch } from "../../../store/hooks";
+import { selectMobileSessionsStreamIds } from "../../../store/mobileSessionsSelectors";
 import { Session } from "../../../types/sessionType";
 import { useMapParams } from "../../../utils/mapParamsHandler";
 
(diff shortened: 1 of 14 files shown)


app/javascript/react/components/Map/Markers/CrowdMapMarkers.tsx
```diff
@@ -0,0 +1,133 @@
...
+      usernames,
+    ]
+  );
+
+  useEffect(() => {
+    dispatch(fetchCrowdMapData(filters));
+  }, [filters]);
+
+  useEffect(() => {
+    if (crowdMapRectangles.length > 0) {
```
evulins, line 82: Can we change it to const and then call it later in the useEffect?

app/javascript/react/components/Map/Markers/CrowdMapMarkers.tsx
```diff
@@ -0,0 +1,133 @@
...
+      );
+      rectanglesRef.current.push(...newRectangles);
+    }
+
+    // Cleanup function to remove rectangles on unmount
+    return () => {
+      rectanglesRef.current.forEach((rectangle) => rectangle.setMap(null));
+      rectanglesRef.current = [];
+    };
+  }, [crowdMapRectangles, map]);
```
evulins, line 104: We need to add ` thresholds ` here as well. Right now square colors are not changing when we adjust the threshold values.

JoannaThelema: Good catch! Thx! 🙌

evulins (approved):
Good job! 👍 I don't have any further comments regarding this matter.

PR #731, opened 2024-08-01
Fix cluster position problems when map view shows IDL + default cluster zooming too close

Sonje:
### Tests detected issues:

1. if you zoom out far enough you can reach IDL, which causes `getClusterPixelPosition` calculations for the clusters right of the line go bonkers
2. if you click on a zoom in, the default zoom in zooms too close, so some of the markers are covered by the UI elements

### Changes:
**ad 1a.** I added a calculation to determine the total width of the world in pixels at the current zoom level
**ad 1b.** to adjust  xOffset for longitude wrap-around I introduced conditions to adjust `xOffset` if it falls outside the range of `-totalWorldPixels / 2` to` totalWorldPixels / 2` and ensured `xOffset` is within the range` [0, totalWorldPixels]` using modulo operation
**ad 2.** added a minimal zoom out on handleZoomIn to ensure none of the markers is hidden behind UI elements

### To test:
Zoom out as far as you can to reach the day line then click on clusters
Zoom in the cluster and check if the markers are not hidden by the UI

### DIFF
--- a/app/javascript/react/components/Map/Markers/FixedMarkers.tsx
+++ b/app/javascript/react/components/Map/Markers/FixedMarkers.tsx
@@ -289,6 +289,10 @@ const FixedMarkers = ({
         selectedCluster,
         map
       );
+      const currentZoom = map.getZoom();
+      if (currentZoom !== null && currentZoom !== undefined) {
+        map.setZoom(currentZoom - 0.5);
+      }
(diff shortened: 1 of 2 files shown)


Sonje:
[fixed: clicking cluster](https://trello.com/c/XasyvVfG/1816-fixed-clicking-cluster)

PR #732, opened 2024-08-02
Allow user to select real-time map updates

Sonje:
### Changes:
- extracted `Toggle` and `CheckboxButton` components to be reusable across different components
- adjusted desktop and mobile headers so all the buttons fit the screen - there are a lot ifs and breakpoints to actually fit the screen (I didn't want to rewrite all the styling, but I removed lots of margins in favour of gaps and paddings)
- added `RealtimeMapUpdatesButton` component (aligned with the current design) which is controlled through `realtimeMapUpdatesSlice`
- when `RealtimeMapUpdatesButton` is enabled, then we'll update sessions whenever map bounds change (added a useEffect to Map component based on the provided requirements)
- I named the slice and component realtime, because it actually acts almost as realtime changes when user interacts with the map


Sonje:
[fixed & mobile: update search when map moves](https://trello.com/c/1g58LRdf/1818-fixed-mobile-update-search-when-map-moves)

evulins (approved):
Amazing work! After testing, I have only one small comment. The toggle dot doesn’t appear to be centered in the input.
Screenshot 2024-08-05 at 21 11 14

JoannaThelema:
Screenshot 2024-08-06 at 10 57 50

Is it intended behavior that for some resolutions we don't display `get airbeam` and `hamburgr menu`?

JoannaThelema:
We have a map-type picker between the refresh and update map buttons. This happens only for some resolutions, it's not a global issue 💪 
Screenshot 2024-08-06 at 11 03 08

JoannaThelema:
Screenshot 2024-08-06 at 11 07 35
Sometimes update map jumps below but the map type picker stays. If it's intended behavior ignore this comment 🙂

JoannaThelema:
And I'm confused because in designs for mobile devices, we don't have a refresh button at all 😅 BTW I prefer your version with a refresh button 💪 
Screenshot 2024-08-06 at 11 12 49

app/javascript/react/components/CheckboxButton/CheckboxButton.tsx
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import * as S from "./CheckboxButton.style";
+
+interface CheckboxButtonProps {
+  label: string;
+  isChecked: boolean;
+  onChange: (isChecked: boolean) => void;
```
JoannaThelema, line 7: ❤️

app/javascript/react/utils/media.ts
```diff
@@ -5,19 +5,28 @@ const sizes = {
   mobile: "768px",
   smallDesktop: "769px",
   desktop: "1024px",
+  mediocreDesktop: "1091px",
+  mediumDesktop: "1240px",
   largeDesktop: "1440px",
+  hugeDesktop: "1620px",
```
JoannaThelema, line 11: 💪

JoannaThelema (commented):
I just briefly went through the code. Looks good to me 🙌

Sonje:
> Screenshot 2024-08-06 at 10 57 50
> Is it intended behavior that for some resolutions we don't display `get airbeam`?

yes it is, there's not enough space to display this icon with all the buttons. This happens for resolutions from mobile to 1091px width, almost no one uses desktop width below 1280px,  wouldn't fuss about it since get air beam is displayed in the menu.

Sonje:
> Screenshot 2024-08-06 at 11 07 35 Sometimes update map jumps below but the map type picker stays. If it's intended behavior ignore this comment 🙂

can you provide me with a video, I don't know what do you exactly mean - if there's not enough space, the header elements will be wrapped  one by one into the next line as it was previously coded using the flex design.

Sonje:
> And I'm confused because in designs for mobile devices, we don't have a refresh button at all 😅 BTW I prefer your version with a refresh button 💪 Screenshot 2024-08-06 at 11 12 49

We don't have refresh button because according to the designs we were supposed to hide the refresh button if the update when map moves has been checked. This requirement has been changed since (see the comments in the trello ticket: https://trello.com/c/1g58LRdf/1818-fixed-mobile-update-search-when-map-moves

app/javascript/react/utils/media.ts
Sonje: this is actually rough - I wanted to keep as many header elements as I could across as small desktop sizes as possible. We should have coded all of the buttons from the beginning without functionalities, that would save us those weird layout shifts when changing the resolution

Sonje:
> We have a map-type picker between the refresh and update map buttons. This happens only for some resolutions, it's not a global issue 💪 Screenshot 2024-08-06 at 11 03 08

That is because from 1023 to 768 we are using tablet size + mobile header element. I did not want to add another if for this use case and decided to keep the mobile order instead of the desktop for this range width (some tablets). This is a very rare width, so I wouldn't bother at this moment since we don't have tablet designs.

JoannaThelema:
> > Screenshot 2024-08-06 at 11 07 35 Sometimes update map jumps below but the map type picker stays. If it's intended behavior ignore this comment 🙂
> 
> can you provide me with a video, I don't know what do you exactly mean - if there's not enough space, the header elements will be wrapped one by one into the next line as it was previously coded using the flex design.

Sounds good enough for me. I was surprised by the order of elements, but again if this is intended behavior ignore this comment 💪

Sonje:
> Amazing work! After testing, I have only one small comment. The toggle dot doesn’t appear to be centered in the input. Screenshot 2024-08-05 at 21 11 14

I added tiny styling change, please let me know if you can see the dot centered on your setup -- thanks!

PR #745, opened 2024-08-14
fix: trigger fetching data when changing location

JoannaThelema:
Ticket: [search location: sessions should refresh after a location is entered](https://trello.com/c/03L4o2hF/1924-search-location-sessions-should-refresh-after-a-location-is-entered)


evulins (approved):
👍

PR #746, opened 2024-08-14
fix: calendar: can't go back

JoannaThelema:


Sonje:
Already fixed in https://github.com/HabitatMap/AirCasting/pull/778

PR #748, opened 2024-08-19
Add threshold continuous uniform distribution button

Sonje:
### Changes:
- added a button that distributes threshold thumbs within min and max measurements taken for the given session details or calendar page
- added error when min and max are the same value (reused threshold configurator error, it gets cleared after 3 seconds of button interaction)
- created a reusable threshold button, so it's easier to apply styles for different designs
- refactored reset button to utilise the default threshold button
- the changes as per ticket are only applied to desktop design
- the distribution also works with negative values
- fixed incorrect typing for fixed stream `sessionId` property - we receive it as a number from API, though it was typed incorrectly as a string everywhere (and comparisons were frequently forced to use string conversion unnecessarily

### Error:
image

### 0-1 example (users can't type the fractions but we have to show them for the proper thumb placement)

Zrzut ekranu 2024-08-19 183618

### Calendar preview

https://github.com/user-attachments/assets/23b8ccad-a4b6-467d-b1dd-b73e95b9a38b

### Negative values example

https://github.com/user-attachments/assets/b3d0e2e1-f0a8-4c9d-8413-cacd4015d490


Sonje:
[Session Details: Adjust Breakpoint Values to Measurements](https://trello.com/c/dhlzzfsK/1836-session-details-adjust-breakpoint-values-to-measurements)

JoannaThelema:
This marker is available in the New York location. 
Screenshot 2024-08-21 at 10 17 42

1. Sometimes the graph is cut off, I suspect it's because we're rounding displayed values and we may have 0.8 on the graph.
2. I see that we're missing one threshold button (between orange and red) 
Screenshot 2024-08-21 at 10 06 44

3. When I reload the page, values are back to default (not how it works on current prod)


##### The same point from prod:
Screenshot 2024-08-21 at 10 33 30

JoannaThelema:
Since we don't have a separate ticket for mobile devices and we're missing reset button only on calendar page I would do it in this pr 💪

app/javascript/react/components/ThresholdConfigurator/ThresholdButtons/UniformDistributionButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const onClick = () => {
+    let min = 0;
+    let max = 0;
+
+    if (
+      mobileStream.status === StatusEnum.Fulfilled &&
+      mobileStream.data.id === sessionId
+    ) {
+      min = Math.round(mobileStream.minMeasurementValue!);
```
JoannaThelema, line 40: I think we need to round the min value always to the lower one e.g. `Math.flor`

app/javascript/react/components/ThresholdConfigurator/ThresholdButtons/UniformDistributionButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+      max = Math.round(mobileStream.maxMeasurementValue!);
```
JoannaThelema, line 41: I think we need to round the max value always to the higher one e.g. `Math.ceil`

app/javascript/react/components/ThresholdConfigurator/ThresholdButtons/UniformDistributionButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+    } else if (
+      fixedStream.status === StatusEnum.Fulfilled &&
+      fixedStream.data.stream.sessionId === sessionId
+    ) {
+      min = Math.round(fixedStream.minMeasurementValue!);
+      max = Math.round(fixedStream.maxMeasurementValue!);
```
JoannaThelema, line 47: Same for fixed stream

JoannaThelema (commented):
The code looks good 🙌 
I found one additional bug: loading thresholds from URL is broken for map and modal view 😅

app/javascript/react/store/thresholdSlice.ts
```diff
@@ -26,10 +26,10 @@ export interface ThresholdState {
...
   defaultValues: {
     min: 0,
-    low: 0,
-    middle: 0,
-    high: 0,
-    max: 0,
+    low: 9,
+    middle: 35,
+    high: 55,
+    max: 150,
```
JoannaThelema, line 32: Why do we need to change defaults? I think this is creating the bug I mentioned earlier. If we want to keep it like that we need to improve conditions for loading thresholds
```
  useEffect(() => {
    if (!isFirstRenderForThresholds.current) {
      dispatch(resetUserThresholds());
    }
    // #DirtyButWorks :nervous-laugh: -> refactor when moving thresholds to url
    if (defaultThresholds.max !== 0) {
      isFirstRenderForThresholds.current = false;
    }
  }, [defaultThresholds]);

  useEffect(() => {
    if (isFirstRenderForThresholds.current) {
      dispatch(setUserThresholdValues(initialThresholds));
    }
    // #DirtyButWorks :nervous-laugh: -> refactor when moving thresholds to url
    if (initialThresholds.max === 0) {
      isFirstRenderForThresholds.current = false;
    }
  }, [initialThresholds]);

**Comment by Sonje** [comment]
@JoannaThelema thanks so much for looking into the issues.
CR changes:
- changed threshold distribution logic from discrete uniform distribution to continuous uniform distribution as we have in production right now,
- fixed issue when slider would sometimes cover the max value (it was a threshold slider bug),
- in the meantime decoupled the logic and calculations from click button
- adjusted the code and styling so the new button can be added to mobile calendar page
- added the button to mobile calendar page with a placeholder button/s - the design is used in mobile map legend so had to adjust it accordingly - added the comment to remove them after adding theme switch functionality

File: app/javascript/react/components/ThresholdConfigurator/ThresholdSlider.tsx [hunk 5]
```diff
@@ -190,7 +190,23 @@ const ThresholdSlider: React.FC<ThresholdSliderProps> = ({
...
+            ? "auto"
+            : `${calculateThumbPosition(
+                value,
+                min,
+                max,
+                sliderWidth,
+                isMobile
+              )}px`,
+        right: thresholdKey === "max" ? "-1px" : "auto",
+      }}
```
JoannaThelema, line 209: 🚀

JoannaThelema (approved):
Great job! 🙌 Let's merge 😎

PR #749, opened 2024-08-20
feat: crowd map marker click

JoannaThelema:
Ticket: [mobile sessions: clicking crowdmap square](https://trello.com/c/LIyIxBaj/1929-mobile-sessions-clicking-crowdmap-square)


app/javascript/react/components/Map/Markers/CrowdMapMarkers.tsx
```diff
@@ -90,21 +104,81 @@ const CrowdMapMarkers = ({ pulsatingSessionId, sessions }: Props) => {
...
+    map && map.addListener("zoom_changed", handleMapInteraction);
+
+    return () => {
+      if (map) {
+        google.maps.event.clearListeners(map, "click");
+        google.maps.event.clearListeners(map, "touchend");
+        google.maps.event.clearListeners(map, "dragstart");
+        google.maps.event.clearListeners(map, "zoom_changed");
+      }
+    };
```
Sonje, line 124: just a food for thought - we use almost the same logic in fixed markers - it would be good to extract it into a separate utility function which would take on event names as argument  🤔

app/javascript/react/components/Map/Markers/RectangleInfo/RectangleInfo.tsx
```diff
@@ -0,0 +1,61 @@
...
+  color,
+  average,
+  numberOfContributors,
+  numberOfSamples,
+}: RectangleProps) => {
+  const { unitSymbol } = useMapParams();
+  const { t } = useTranslation();
+
+  return (
+    <>
```
Sonje, line 33: React fragment seems to be redundant here

Sonje (commented):
Nice job with the crowd map marker information. I added tiny comments and tested it.

I have 2 concerns - one is of UX nature and the other is about the design:

### Fetching data - nothing happens for a few seconds after  click
I know it's not frontend's fault but requests take substantial amount of time especially when there's big chunk of data. Maybe it could be possible to show the cluster information with loader (or just fetching text like in production) and show data once its available?

### Design - blur
That blurred colour circle (just like in normal cluster info) feels off when the information bubble is that big. Can we either remove it altogether or add the AQI coloured blur around the whole crowd map cluster information?

Sonje (approved):
giphy (1)

JoannaThelema:
> Nice job with the crowd map marker information. I added tiny comments and tested it.
> 
> I have 2 concerns - one is of UX nature and the other is about the design:
> 
> ### Fetching data - nothing happens for a few seconds after click
> I know it's not frontend's fault but requests take substantial amount of time especially when there's big chunk of data. Maybe it could be possible to show the cluster information with loader (or just fetching text like in production) and show data once its available?
> 
> ### Design - blur
> That blurred colour circle (just like in normal cluster info) feels off when the information bubble is that big. Can we either remove it altogether or add the AQI coloured blur around the whole crowd map cluster information?

Loader added, blurred color circle removed.

PR #752, opened 2024-08-22
Add crowdmap toggle and adjust related components

Sonje:
### Changes
Refined the crowdmap toggle
- on desktop, it directly toggles between `CrowdMapView` and `MapView`
- on mobile, when filters are open, it updates `previousUserSettings` instead of `currentUserSettings` (I am aware it's not perfect, but I think that's the quickest solution)
- the crowdmap toggle respects the active view when switching between map types and handles the case when the filters view is active on mobile
- I made sure the CrowdMap toggle is ready for grid cell sizes (just uncomment the grid-gap property and add a new component or code below `CrowdMapToggleWrapper` within `CrowdMapSettingsContainer`

Adjusted default toggle component:
- added new props to default toggle so we can use it without labels and if needed on mobile have a bigger toggle (same as on desktop) - this design varies between components, so I'd rather keep it flexible

### Design:
- I am totally aware that the toggle button on mobile is on the right in the designs, but I want to convince Michael that it is inconsistent with dormant session and that we should keep one way of showing the toggle in filters (we can decide whether it's right or left, because otherwise it's bad for UX). Let me know if you think that instead of complaining I should implement it as per designs against my better judgment. 

Mobile preview:

https://github.com/user-attachments/assets/06e31d95-39a0-402e-8557-eda75da8ea28

### DIFF
--- a/app/javascript/react/components/SessionFilters/CrowdmapToggle.tsx
+++ b/app/javascript/react/components/SessionFilters/CrowdmapToggle.tsx
@@ -1,29 +1,111 @@
-import React from "react";
+import React, { useEffect, useState, useCallback } from "react";
+import { useTranslation } from "react-i18next";
+import { debounce } from "lodash";
 
 import { UserSettings } from "../../types/userStates";
-import { useMapParams } from "../../utils/mapParamsHandler";
+import { UrlParamsTypes, useMapParams } from "../../utils/mapParamsHandler";
 import * as S from "./SessionFilters.style";
-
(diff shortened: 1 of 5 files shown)


Sonje:
[Session Filter [Mobile]: Crowdmap Toggle](https://trello.com/c/YfvI7JM3/1853-session-filter-mobile-crowdmap-toggle)

evulins (approved):
Good job! 🥇

PR #757, opened 2024-08-23
Add outdoor indoor filter

bazelajoanna:


JoannaThelema (commented):
I would keep one parameter in the URL and I would use `UserSettings` instead of true/false flag, but I'm leaving this decision to you 💪 For sure let's be consistent and use only one param in the URL for indoor.

app/javascript/react/components/Map/Map.tsx
```diff
@@ -347,7 +383,7 @@ const Map = () => {
     }
 
     if (selectedStreamId) {
-      dispatch(setMarkersLoading(true));
+      isMobile && !isIndoorParameterInUrl && dispatch(setMarkersLoading(true));
```
JoannaThelema, line 386: I think this line can be removed

app/javascript/react/components/Map/Map.tsx
```diff
@@ -260,12 +290,18 @@ const Map = () => {
 
   useEffect(() => {
     if (streamId && currentUserSettings === UserSettings.ModalView) {
-      dispatch(setMarkersLoading(true));
+      !isIndoorParameterInUrl && dispatch(setMarkersLoading(true));
```
JoannaThelema, line 293: this line also can be removed

app/javascript/react/components/Map/Markers/FixedMarkers.tsx
```diff
@@ -339,6 +341,7 @@ const FixedMarkers = ({
   });
 
   return (
+    // !isIndoorParameterInUrl && (
```
JoannaThelema, line 344: leftover

app/javascript/react/components/SessionsListView/SessionsListTile/SessionListTile.tsx
```diff
@@ -13,7 +13,7 @@ interface SessionListTile {
   id: number;
   sessionName: string;
   sensorName: string;
-  averageValue: number;
+  averageValue: number | "no data";
```
JoannaThelema, line 16: hmm... I would use `averageValue?: number;` and later you can check if it's undefined

PR #758, opened 2024-08-26
Timelapse play

evulins:
List of completed tasks in this PR:

- Added timelapse axis with a time range
- Added functionality for navigation buttons
- Added time range buttons
- Added timelapse slice and selectors
- Disabled timelapse button when there is no data


JoannaThelema (approved):
😻

PR #762, opened 2024-08-27
Fix sensors

evulins:
Changes made in this pull request:
- Excluded air beam mini from the custom sensor list
- Fixed default sensor parameters for fixed sessions
- Hid filters on timelapse view

[Trello card](https://trello.com/c/xOJY2hMN/1846-session-filter-both-sensor-picker)


bazelajoanna (approved):
🦃

PR #766, opened 2024-08-28
fix: clear sessions only when user turns on realtimeMapUpdates

Sonje:
We were clearing session list when update map button was unchecked, and this action does not trigger a refetch, so I added a guard against this case

https://github.com/user-attachments/assets/7af5eff3-15f3-4c7e-a802-967ee41de5d7


Sonje:
[Mobile sessions > sessions list: missing design and feature](https://trello.com/c/DwsowEbV/1932-mobile-sessions-sessions-list-missing-design-and-feature)

JoannaThelema (approved):

PR #767, opened 2024-08-28
Fix: session lists styling

Sonje:
Tiny fixes, for session list tiles being distributed weirdly, and mobile session list tiles being clustered. I also added the session list padding for mobile, because the last child of session tiles was cut off on mobile devices.


bazelajoanna (approved):
🦃

PR #768, opened 2024-08-28
Fix sessions count on sessions list

bazelajoanna:


Sonje (approved):
👏 🦃 

image



imfa1-imfa2

PR #771, opened 2024-08-29
[Calendar] add calendar data card dynamic color styles

Sonje:
**Changes**
- Adjust the calendar card colors as per requirements -> they should change the color based on thresholds and avg value we receive.

**Design note:** _Based on the color theory the colors we are using for threshold are quite connected to each other, so I decided to adjust colors, shades and % of gradient we are applying in each case, while trying to retain a modern gradient look. Personally yellow is the hardest ;)_ 

Based on designs by yours truly:
https://www.figma.com/design/wxBRnvjC2Y8T1FBNaiIAyv/AC---Calendar-Data-Card?node-id=0-1&node-type=CANVAS&t=rakNIwwu4uoFwNVV-0

### DIFF
--- a/app/javascript/react/assets/styles/colors.ts
+++ b/app/javascript/react/assets/styles/colors.ts
@@ -21,6 +21,9 @@ const red = "#E95F5F";
 const orange = "#FCA443";
 const yellow = "#FFD960";
 const green = "#96D788";
+const lightCrimson = '#ED6B86'
+const rose = '#FF38B5'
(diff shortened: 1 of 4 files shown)


Sonje:
[calendar: color "Avg for date" box according to the values set on the TC](https://trello.com/c/pIsGcxJ4/1892-calendar-color-avg-for-date-box-according-to-the-values-set-on-the-tc)

JoannaThelema (approved):

PR #778, opened 2024-08-30
Fix Calendar's back to previous view on mobile issues

Sonje:
### Changes
- extracted a separate back button component, so we can reuse it on desktop
- enhanced logic around going back from calendar to previous view (fixed the problem with map being centered on the session when going back from Calendar page instead of the previous map preview)
- fixed issues when going from calendar back to session details modal caused issues when closing the modal (the calendar view was invoked instead of a map view)
- added experimental browser back handling - we are listening to `popstate` event and in case it goes off, we are applying our own back logic, the same as in the button;). This is far from ideal solution, but fixing what is going on in history would take much more time, which we don't have at this moment
- extracted back hook we are using in two places to keep the logic in one place

### DIFF
--- a/app/javascript/react/components/Navbar/BackButton.tsx
+++ b/app/javascript/react/components/Navbar/BackButton.tsx
@@ -10,7 +10,7 @@ import { UserSettings } from "../../types/userStates";
 import backArrowIcon from "../../assets/icons/backArrowIcon.svg";
 import backArrowIconDesktop from "../../assets/icons/backArrowIconDesktop.svg";
 
-const BackToMapButton = () => {
+const BackButton = () => {
   const { t } = useTranslation();
 
   const {
@@ -64,4 +64,4 @@ const BackToMapButton = () => {
   );
 };
 
-export { BackToMapButton };
+export { BackButton };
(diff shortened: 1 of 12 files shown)


Sonje:
[fixed > calendar: can't go back](https://trello.com/c/itqpCiiy/1951-fixed-calendar-cant-go-back)

Sonje:
[mobile devices > fixed: returning from calendar doesn't preserve former map view](https://trello.com/c/8hsx2M7j/1960-mobile-devices-fixed-returning-from-calendar-doesnt-preserve-former-map-view)

app/javascript/react/components/Navbar/BackButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+      case UserSettings.SessionListView:
+        return t("navbar.goBackToSessions");
+      case UserSettings.MapView:
+        return t("navbar.goBackToMap");
+      case UserSettings.ModalView:
+        return t("navbar.goBackToSession");
+      default:
+        return t("navbar.goBackToMap");
+    }
+  }, [previousUserSettings, t]);
```
evulins, line 35: ❤️

evulins (approved):
Code looks great! 🥇   On minor thing to fix -> closing sessions list on mobile devices.

PR #792, opened 2024-09-05
Improve fixed markers and clusters performance

Sonje:
The threshold update now only affects the visual appearance of clusters, avoiding a full re-render of markers and clusters, which optimizes performance, especially in large datasets.

### Threshold Update Handling:

**Before**: When thresholds were updated, all markers and clusters were reloaded, causing unnecessary performance overhead and loading took sometimes up to 25 seconds.
**After**: A new function `updateClusterStyle` was added, which dynamically updates the styles (colors) of clusters and markers without reloading them. This function is called whenever thresholds change, ensuring that only the color and visual properties are updated based on the new thresholds.

### Cluster Rendering:

**Before**: The `customRenderer` directly created new cluster elements every time thresholds changed.
**After**: The `customRenderer` now references existing clusters and updates their styles using the `updateClusterStyle` function. Additionally, the `customRenderer` now stores a reference to each cluster element (clusterElementsRef), so these elements can be updated without recreating them.

### DIFF
--- a/app/javascript/react/components/Map/Markers/ClusterConfiguration.tsx
+++ b/app/javascript/react/components/Map/Markers/ClusterConfiguration.tsx
@@ -60,7 +60,28 @@ const calculateClusterStyleIndex = (
   return styleIndex;
 };
 
-export const customRenderer = (thresholds: Thresholds) => ({
+export const updateClusterStyle = (
+  clusterElement: google.maps.marker.AdvancedMarkerElement,
+  markers: google.maps.marker.AdvancedMarkerElement[],
+  thresholds: Thresholds
+) => {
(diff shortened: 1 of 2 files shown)


Sonje:
[fixed > view station > fit to scale > close station > reset scale: stations disappear from map, no loader is displayed, takes 25 seconds or longer to display stations again](https://trello.com/c/udDRtdYj/1991-fixed-view-station-fit-to-scale-close-station-reset-scale-stations-disappear-from-map-no-loader-is-displayed-takes-25-seconds-or)

evulins (approved):
I have no words 🦃  🥇 Works great

PR #801, opened 2024-09-09
Unify error logging and improve sentry logs

Sonje:
Changes:
- improved error interface for ApiErrors
- unified error logging across the slices
- added Sentry logs to errors from requests
- moved Sentry initialization into a separate utility function
- added discern between different environments(staging,  experimental, production), removed development environment from Sentry logging
- updated Sentry library
- I chose null for errors instead of undefined as it leads to more clear results - we have full control over errors, we are clearing them with each state change and there will be no unexpected "undefineds"
- reused default async thunk withing fixed sessions to fetchDormantFixedSessions (simplification while still using isAdditional arg)
- added types wherever they were missing


Sonje:
[Fix Sentry setup](https://trello.com/c/U2bauA4V/2011-fix-sentry-setup)

PR #802, opened 2024-09-10
[Performance] split the logic for fetching airbeam/government data to improve performance of /sessions2 endpoint

BartkiewiczM:


app/repositories/sessions_repository.rb
```diff
@@ -6,4 +6,29 @@ def find_with_streams(uuid:)
...
+      FROM sessions AS s
+      JOIN streams AS st ON s.id = st.session_id
+      LEFT JOIN (
+          SELECT stream_id, value,
+                 ROW_NUMBER() OVER (PARTITION BY stream_id ORDER BY date DESC) as rn
+          FROM stream_daily_averages
+      ) AS sdavalue ON st.id = sdavalue.stream_id AND sdavalue.rn = 1
+      WHERE LOWER(st.sensor_name) = #{sensor_name}
+        AND s.latitude BETWEEN #{south} AND #{north}
+        AND s.longitude BETWEEN #{west} AND #{east}
```
mpolec, line 28: How do we stand with the idea to use geometry data type for sessions location as well?

app/repositories/sessions_repository.rb
```diff
@@ -6,4 +6,29 @@ def find_with_streams(uuid:)
   def active_in_last_7_days
     Session.where('last_measurement_at > ?', Time.current - 7.days)
   end
+
+  def fixed_active_government_sessions(data)
```
mpolec, line 10: So far we were trying to expose a method interface, so it would be clear what params it takes.
I see that we have a long list here 🤔

app/controllers/api/fixed/active/sessions_controller.rb
```diff
@@ -17,7 +17,13 @@ def index
...
         def index2
           GoogleAnalyticsWorker::RegisterEvent.async_call('Fixed active sessions#index2')
-          result = Api::ToActiveSessionsJson.new(form: form).call
+          # splitting the logic for governemnt data, to improve performance
+          sensor_name = form.to_h.to_h[:sensor_name]
+          if sensor_name == 'government-pm2.5' || sensor_name == 'government-no2' || sensor_name == 'government-o3'
+            result = ::FixedSessions::IndexInteractor.new(form: form).call
+          else
+            result = Api::ToActiveSessionsJson.new(form: form).call
+          end
```
mpolec, line 26: I guess it's not very important, but I wonder if it should be split here or in the interactor.
Right now we have some duplication, in both services we check if the form is valid:
`return Failure.new(form.errors) if form.invalid?`

app/serializers/fixed_sessions_serializer.rb
```diff
@@ -0,0 +1,36 @@
...
+  def call(sessions)
+    sessions_array = []
+
+    sessions.map do |session|
+      sessions_array <<
+        {
+          'id' => session["session_id"],
+          'uuid' => session["uuid"],
+          'end_time_local' => session["end_time_local"]&.strftime('%Y-%m-%dT%H:%M:%S.%LZ'),
+          'start_time_local' => session["start_time_local"]&.strftime('%Y-%m-%dT%H:%M:%S.%LZ'),
```
mpolec, line 11: There could be a private method defined for time formatting.

app/repositories/sessions_repository.rb
```diff
@@ -6,4 +6,29 @@ def find_with_streams(uuid:)
...
+
+    sql = <<-SQL
+      SELECT s.*, st.*, sdavalue.value AS last_daily_average
+      FROM sessions AS s
+      JOIN streams AS st ON s.id = st.session_id
+      LEFT JOIN (
+          SELECT stream_id, value,
+                 ROW_NUMBER() OVER (PARTITION BY stream_id ORDER BY date DESC) as rn
+          FROM stream_daily_averages
+      ) AS sdavalue ON st.id = sdavalue.stream_id AND sdavalue.rn = 1
```
mpolec, line 25: Seems to be pretty complicated, maybe there is a way to simplify it?

PR #842, opened 2024-10-07
Change stream and  hover markers

evulins:
Use regular google markers for hover and stream markers.


Sonje (approved):
looks good codewise, works well in staging

PR #847, opened 2024-10-08
Fix stream markers

evulins:
Add custom marker overaly and use in hover markers and stream marker


Sonje (approved):
just a request to change class component into functional component since we use them everywhere else.

PR #851, opened 2024-10-08
Add fixed markers overlay with pulsation and selected options

Sonje:
Switch fixed markers to use overlay for performance improvements

- cluster pulsation is still behind markers, we'll fix it after updating the clusters


app/javascript/react/components/Map/Markers/FixedMarkers.tsx
```diff
@@ -87,6 +84,11 @@ export function FixedMarkers({
 
   const memoizedSessions = useMemo(() => sessions, [sessions]);
 
+  useEffect(() => {
+    console.log("url stream Id", selectedStreamId);
```
evulins, line 88: Do we need this useEffect with console.logs?

evulins (approved):
Amazing work! 🦃

app/javascript/react/components/Map/Markers/FixedMarkers.tsx
Sonje: that's a debugging leftorver, removed it -> thanks for catching that

PR #854, opened 2024-10-10
Fix filters on mobile devices

bazelajoanna:


Sonje (approved):
well done, and a good catch!

PR #858, opened 2024-10-10
Collapse sessions on sessions list

bazelajoanna:


Sonje (approved):
works well, keep up te great work!

PR #860, opened 2024-10-11
Use custom overlays to display clustered fixed markers

Sonje:
Done:
- converted clusters to use custom overlays
- adjusted cluster styles to resemble the original svgs
- made sure the cluster colors are changed when thresholds change
- adjusted clustering, so less singular markers appear on initial load (i think it got changed to 40, but it looks clamped with this setting, so I set it to 50 and it looks better - I will ask Michael if that is okay)

Test notes:
- make sure clusters pulsate over singular markers, but remain hidden behind them if they do not pulsate
- make sure clusters, markers are cleared when selected stream id is set
- make sure the performance remains the same or better (pulsating is definitely better)
- make sure the clustering works :D
- make sure clicking on clusters and markers work as expected
- make sure the selected stream id marker appears when you copy session in another tab and other markers are not displayed

### DIFF
--- a/app/javascript/react/components/Map/Markers/FixedMarkers.tsx
+++ b/app/javascript/react/components/Map/Markers/FixedMarkers.tsx
@@ -556,7 +556,6 @@ export function FixedMarkers({
   useEffect(() => {
     const handleSelectedStreamId = (streamId: number | null) => {
       if (fixedStreamStatus === StatusEnum.Pending) return;
-      console.log("streamId", streamId);
       if (streamId) {
         const { latitude, longitude } = fixedStreamData?.stream ?? {};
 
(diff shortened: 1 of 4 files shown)


Sonje:
[Update clusters](https://trello.com/c/iCweEyAz/2069-update-clusters)

app/javascript/react/components/Map/Markers/FixedMarkers.tsx
```diff
@@ -478,29 +547,163 @@ export function FixedMarkers({
...
     }
   }, [map, customRenderer, handleClusterClickInternal]);
 
   // Needed to handle fixed session opened in a new tab - otherwise markers won't appear
   useEffect(() => {
     const handleSelectedStreamId = (streamId: number | null) => {
-      if (!streamId || fixedStreamStatus === StatusEnum.Pending) return;
-      const { latitude, longitude } = fixedStreamData?.stream ?? {};
+      if (fixedStreamStatus === StatusEnum.Pending) return;
+      console.log("streamId", streamId);
```
bazelajoanna, line 559: Leftover, I guess :)

PR #866, opened 2024-10-15
Update map styles

Sonje:
Update map styles to resemble production as much as possible.


Sonje:
[make staging map styles match production map styles](https://trello.com/c/iXfTkTit/2070-make-staging-map-styles-match-production-map-styles)

PR #869, opened 2024-10-16
adjust map styles properties

Sonje:
Adjust map styles properties so there are no console errors. Concerns:
- administrative.reservation
- administrative.international_border


evulins (approved):
🚀

PR #872, opened 2024-10-18
Fix crowdmap issues

Sonje:


bazelajoanna:
I've also fixed going back from modal behaviour (https://trello.com/c/H3qdJ1DQ/2080-mobile-devices-click-mobile-marker-on-map-click-back-to-map-filters-are-open)

PR #875, opened 2024-10-18
Fix saving in local storage - saving in cookies

bazelajoanna:
Save info about the map (bounds and zoom) in cookies and reproduce old production behavior.


evulins (approved):
🥇

PR #876, opened 2024-10-22
Save bounds in url on map load

bazelajoanna:
To fix this issue:
timelapse > copy link > open new browser window and navigate to staging > pan & zoom map > paste copied link: link doesn't preserve map view and timelapse selection

We need to have map bounds in url on map load.


evulins (approved):
🏎️ 🚀

PR #880, opened 2024-11-05
View notes

bazelajoanna:


app/javascript/react/utils/groupNotes.ts
```diff
@@ -0,0 +1,19 @@
+import { Note } from "../types/note";
+
+const roundCoordinate = (coord: number): number => {
+  return Math.round(coord * 1000000) / 1000000;
```
evulins, line 4: move it to const this number that we know, what is this huge number :)

evulins (approved):
Great work! 💪  👍  one little comment :)

PR #881, opened 2024-11-07
Show loading indicator while fetching data

bazelajoanna:
- show loading indicator on Crowdmap Markers and Cluster Markers
- prevent closing Filters when click "Copy link" button


evulins (approved):
🥇

PR #906, opened 2024-11-29
Don't update bounds on each map move

bazelajoanna:
The issue: fixed > nyc > zoom out > timelapse: when timelapse is activated, the map refreshes so it includes all stations in the map view

Fix: map bounds are updated and set in URL on "Refresh map" button click, on mobile session type toggle and while "Update when map moves" is selected.


evulins (approved):
Great work! 🦃

PR #929, opened 2025-01-02
Prevent map zooming on FixedMarkers hover on graph

bazelajoanna:
fixed > graph > pan and zoom map > hover graph: map doesn’t maintain position, map pans and zooms back to default position


evulins (approved):
🥇

PR #933, opened 2025-01-03
Disabe 3D view in satelite mode, change export sessions error copy

bazelajoanna:
Set `tilt: 0` in the GoogleMap component to prevent 3D view in satellite mode.
Change export more than 100 sessions error copy.


evulins (approved):
👍

PR #962, opened 2025-01-21
Fix min and max values in the thresholds.

bazelajoanna:
Prevent `min` and `max` thresholds from being grabbed, don't allow to put `max` value higher than `high` value.


evulins (approved):
🚀

PR #1172, opened 2026-03-30
fix: Cap desktop sessions list export by sensor type

evulins:
Add tiered limits for desktop bulk session export

[Trello task](https://trello.com/c/A73CpQwM/2296-limit-sessions-list-exports-to-100-sessions-or-less)


app/services/api/validate_session_export_ids.rb
```diff
@@ -0,0 +1,81 @@
...
+  def initialize(session_ids:)
+    @session_ids = session_ids
+  end
+
+  def call
+    unique_ids = @session_ids.uniq
+
+    return Failure.new(session_ids: ['must include at least one session']) if unique_ids.empty?
+
+    sessions = Session.where(id: unique_ids).includes(:streams)
```
hanka: Minor thing, but converting sessions to array (`sessions = Session.where(id: unique_ids).includes(:streams)`) and then comparing `sessions.length` with the unique_ids.size would save 1 additional query to DB

hanka (approved):
Looks good, thanks!

